PR #10, opened 2020-10-15
Added AddOData overload with access to an IServiceCollection (#9)

### LINKED ISSUE #9: Please add an overload with an IServiceProvider accessor to AddOData()
Hi,

First and foremost, thanks for the great work, the new way of setting up OData makes much more sense!

I'd like to request adding another overload of the `AddOData` extension set, such that its setupAction delegate comes with an `IServiceProvider` argument, so that we can resolve services (e.g. `IEdmModel` from IoC).

weitzhandler:
Fixes #9.


xuzhg:
Merged at https://github.com/OData/AspNetCoreOData/commit/00342e054f1944f99f7bd528700e652f7b304169

PR #33, opened 2020-11-11
Changed CreatedODataResult and UpdatedODataResult to inherit from Action result

### LINKED ISSUE #17: 'UpdatedODataResult' and 'CreatedODataResult' are not convertible to 'ActionResult<T>'
Both OData-specific action result types `UpdatedODataResult` and `CreatedODataResult` do not currently support being converted into a `ActionResult<T>` value. This is because neither type inherit from the abstract `ActionResult` type, which is what allows the automatic conversion to take place.

From my understanding, this gap in functionality occurred when the core MVC team introduced `ActionResult<T>` and the `ActionResult` base class. At that time, these OData-specific types already existed and only implemented `IActionResult`.

Allowing the OData results to also be convertible to `ActionResult<T>` will result in a more consistent and friendly framework (imagine migrating normal controllers to OData for example, and the "principle of least astonishment" violation) as well as provide better readability by exposing the type on the return explicitly (be it for the developers, or tools like Swagger).

weitzhandler:
Fixes #17.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Results/CreatedODataResult.cs
+++ b/src/Microsoft.AspNetCore.OData/Results/CreatedODataResult.cs
@@ -18,7 +18,7 @@ namespace Microsoft.AspNetCore.OData.Results
     /// <remarks>This action result handles content negotiation and the HTTP prefer header. It generates a location
     /// header containing the edit link of the created entity and, if response has status code: NoContent, also
     /// generates an OData-EntityId header.</remarks>
-    public class CreatedODataResult<T> : IActionResult
+    public class CreatedODataResult<T> : ActionResult
     {
         private readonly T _innerResult;
 
@@ -48,7 +48,7 @@ namespace Microsoft.AspNetCore.OData.Results
         }
 
         /// <inheritdoc/>
-        public async virtual Task ExecuteResultAsync(ActionContext context)
+        public async override Task ExecuteResultAsync(ActionContext context)
         {
             if (context == null)
             {
--- a/src/Microsoft.AspNetCore.OData/Results/UpdatedODataResult.cs
+++ b/src/Microsoft.AspNetCore.OData/Results/UpdatedODataResult.cs
@@ -16,7 +16,7 @@ namespace Microsoft.AspNetCore.OData.Results
     /// </summary>
     /// <typeparam name="T">The entity type.</typeparam>
     /// <remarks>This action result handles content negotiation and the HTTP prefer header.</remarks>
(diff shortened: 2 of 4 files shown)


src/Microsoft.AspNetCore.OData/Results/CreatedODataResult.cs
```diff
@@ -18,7 +19,7 @@ namespace Microsoft.AspNetCore.OData.Results
     /// <remarks>This action result handles content negotiation and the HTTP prefer header. It generates a location
     /// header containing the edit link of the created entity and, if response has status code: NoContent, also
     /// generates an OData-EntityId header.</remarks>
-    public class CreatedODataResult<T> : IActionResult
+    public class CreatedODataResult<T> : ActionResult, IConvertToActionResult
```
julealgon, line 22: I believe implementing `IConvertToActionResult` is not needed here. This is implemented by `ActionResult<T>` and it already handles anything that inherits from the base `ActionResult` class.

If you check all other native result classes (like `OkObjectResult`, `OkCreatedResult`, etc) they all inherit from `ActionResult`, and do _not_  implement `IConvertToActionResult`.

Would recommend double checking and dropping this implementation.

src/Microsoft.AspNetCore.OData/Results/UpdatedODataResult.cs
```diff
@@ -16,7 +17,7 @@ namespace Microsoft.AspNetCore.OData.Results
     /// </summary>
     /// <typeparam name="T">The entity type.</typeparam>
     /// <remarks>This action result handles content negotiation and the HTTP prefer header.</remarks>
-    public class UpdatedODataResult<T> : IActionResult
+    public class UpdatedODataResult<T> : ActionResult, IConvertToActionResult
```
julealgon, line 20: Same as above. Implementing `IConvertToActionResult` shouldn't be needed (it might even be a semantic mistake since this already implements `IActionResult` so it is already convertible to it).

julealgon (changes requested):

julealgon:
Thanks again for taking this up, much appreciated.

src/Microsoft.AspNetCore.OData/Results/CreatedODataResult.cs
weitzhandler: You're right.

julealgon (approved):

xuzhg:
Merged https://github.com/OData/AspNetCoreOData/commit/faf5a6455c1b62560ec563d7a4d3fbec0a5e2c2c

Issue #34, opened 2020-11-18
Add AuthorizeAttribute in the endpoint metadata without namespace


xalhera:
For the action/function, the AuthorizeAttribute is not added in the endpoint metadata without namespace.

![image](https://user-images.githubusercontent.com/20375763/99476595-f82ce800-29a4-11eb-8b9d-a0236b96d632.png)

Framework .Net 5
Microsoft.AspNetCore.OData v8.0.0-preview2

Thanks a million for your support 👍

PR #35, opened 2020-11-18
fix issue #34, add the attributes to new selector model

xuzhg:
Related to issue #34.

It seems we can’t hit the Authorize because the new added "SelectorModel" doesn't have the attributes added.

 
In the debug endpoint, there’s no “AuthorizeAttribute” in the metadata of Endpoint

 
image

However,  for “Get” method, the metadata of Endpoint has the “AuthrozeAttribute” as below:

 
image

This PR is to fix this problem.


xalhera:
AuthorizeAttribute works fine for Actions and Functions with this PR 👍 

Thanks @xuzhg ! Great job 🥇

PR #38, opened 2020-11-27
async formatters

### LINKED ISSUE #4: Replace the synchronous writing with asynchronous writing call
So far, we have the https://github.com/OData/AspNetCoreOData/blob/master/src/Microsoft.AspNetCore.OData/Formatter/ODataInputFormatter.cs#L123-L127 setting to allow synchronous call. 

We should get rid of this setting and replace all writing call using asynchronous version.

cognvn:
Fixes #4 

Based on commit <https://github.com/OData/WebApi/commit/4ff7ca8e8a46bc00d639480149df3fb1d7e50ca8>


xuzhg:
Merged at https://github.com/OData/AspNetCoreOData/commit/f5021e31fed68481af47b956360ac53f7232eac7

@cognvn Thanks for your contribution. Looking forward more. 👍

Issue #119, opened 2021-03-29
Passing OData query options in request body


vistnesr:
Hi,

In version 7.5, it was possible to pass query options in the request body:
https://docs.microsoft.com/en-us/odata/webapi/query-options-in-request-body.

This does not seem to work in 8.0-beta.

Am I doing something wrong, or has it not yet been implemented?

xuzhg:
That's not ported from 7.x. 
@gathogojr Would you please help on it?

gathogojr:
Hi @xuzhg. It's long overdue. I have created a work item for it and will get started on it right away

xuzhg:
@gathogojr don't worry, I am almost done. Will share a link for you review.

xuzhg:
@gathogojr here it is: https://github.com/OData/AspNetCoreOData/commit/31fa96022390048c13146241408afd8287fb41a1

Based on your implementation in 7.x, i made some changes. please take a look and let me know your thoughts. 

By the way, https://github.com/OData/AspNetCoreOData/blob/master/test/Microsoft.AspNetCore.OData.E2E.Tests/QueryRequest/DollarQueryTests.cs#L231 is for you. :)

xuzhg:
@vistnesr You can try this feature in the latest nightly. Please let us know your questions.

gathogojr:
> @gathogojr here it is: [31fa960](https://github.com/OData/AspNetCoreOData/commit/31fa96022390048c13146241408afd8287fb41a1)
> 
> Based on your implementation in 7.x, i made some changes. please take a look and let me know your thoughts.
> 
> By the way, https://github.com/OData/AspNetCoreOData/blob/master/test/Microsoft.AspNetCore.OData.E2E.Tests/QueryRequest/DollarQueryTests.cs#L231 is for you. :)

Will take a look and revert.
_DollarQueryTests.cs_ contains the tests for the $query feature and I see you already migrated it over.

vistnesr:
> @vistnesr You can try this feature in the latest nightly. Please let us know your questions.

I just tried the latest nightly and it works. I looked over the code, too, and I like the way it was implemented.

Thanks for the quick response @xuzhg

gathogojr:
Feature ported to 8.x via https://github.com/OData/AspNetCoreOData/commit/31fa96022390048c13146241408afd8287fb41a1

PR #121, opened 2021-03-31
Fix test failing after porting feature to OData 8.0-beta

gathogojr:
Fixes test failing subsequent to porting feature to support passing of query options in request body
Related to #119

### DIFF
--- a/src/Microsoft.AspNetCore.OData/ODataServiceCollectionExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/ODataServiceCollectionExtensions.cs
@@ -109,7 +109,8 @@ namespace Microsoft.AspNetCore.OData
                 throw Error.ArgumentNull(nameof(services));
             }
 
-            services.TryAddSingleton<IODataQueryRequestParser, DefaultODataQueryRequestParser>();
+            services.TryAddEnumerable(
+                ServiceDescriptor.Singleton<IODataQueryRequestParser, DefaultODataQueryRequestParser>());
             services.TryAddSingleton<IAssemblyResolver, DefaultAssemblyResolver>();
             services.TryAddSingleton<IODataTypeMappingProvider, ODataTypeMappingProvider>();
 
--- a/test/Microsoft.AspNetCore.OData.E2E.Tests/QueryRequest/DollarQueryTests.cs
+++ b/test/Microsoft.AspNetCore.OData.E2E.Tests/QueryRequest/DollarQueryTests.cs
@@ -8,6 +8,7 @@ using System.Net.Http.Headers;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.OData.E2E.Tests.Commons;
 using Microsoft.AspNetCore.OData.TestCommon;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.OData.Edm;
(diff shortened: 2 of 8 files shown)


xuzhg (changes requested):

test/Microsoft.AspNetCore.OData.E2E.Tests/QueryRequest/DollarQueryTests.cs
```diff
@@ -16,14 +17,14 @@
 
 namespace Microsoft.AspNetCore.OData.E2E.Tests.Routing.QueryRequest
 {
-    public class DollarQueryTests : WebApiTestBase<DollarQueryTests>
```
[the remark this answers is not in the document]
gathogojr: @xuzhg When working on the feature, a reviewer had requested for a test that verifies two things:
- that a request fails when the query options part of the URL exceeds a particular length, and
- that a request succeeds if you pass the same query options in the request body

That is what gave birth to the test in question. The challenge with `WebApiTestBase<T>` is that it makes use of `TestServer` and I couldn't figure out a way to impose request limits. Kestrel server on the other hand imposes the limits that make it possible to implement the requested test

That leaves us with the following options:
- get rid of the test altogether
- or, figure out a way to fail a request when the header violates configured limits when using `WebApiTestBase<T>`/`TestServer` - maybe we could use a middleware to achieve that?
- or, move this particular test to its own test class that derives from `WebHostTestBase<T>` class

test/Microsoft.AspNetCore.OData.E2E.Tests/QueryRequest/DollarQueryTests.cs
```diff
@@ -227,52 +227,5 @@ public async Task ODataQueryOptionsInRequestBody_Empty()
             // Assert
             Assert.True(response.IsSuccessStatusCode);
         }
-
```
gathogojr, line 230: Apart from confirming a known fact - that web server impose limitations on the request header size - this test doesn't add more value so its justifiable to drop it

src/Microsoft.AspNetCore.OData/ODataServiceCollectionExtensions.cs
```diff
@@ -109,7 +109,8 @@ private static IODataBuilder AddODataCore(this IServiceCollection services)
                 throw Error.ArgumentNull(nameof(services));
             }
 
-            services.TryAddSingleton<IODataQueryRequestParser, DefaultODataQueryRequestParser>();
+            services.TryAddEnumerable(
+                ServiceDescriptor.Singleton<IODataQueryRequestParser, DefaultODataQueryRequestParser>());
```
gathogojr, line 113: @xuzhg Okay to register as Singleton or would Transient be preferable?

xuzhg: I think it should be singleton.

---
In reply to: [610428027](https://github.com/OData/AspNetCoreOData/pull/121#discussion_r610428027) [](ancestors = 610428027)

test/Microsoft.AspNetCore.OData.E2E.Tests/QueryRequest/DollarQueryTests.cs
xuzhg: would you please share a link for that?

---
In reply to: [610427377](https://github.com/OData/AspNetCoreOData/pull/121#discussion_r610427377) [](ancestors = 610427377)

xuzhg (approved):
:shipit:

test/Microsoft.AspNetCore.OData.Tests/PublicApi/Microsoft.AspNetCore.OData.PublicApi.NetCore31.bsl
```diff
@@ -1369,7 +1369,7 @@ public class Microsoft.AspNetCore.OData.Query.ODataQueryOptions`1 : Microsoft.As
 }
 
 public class Microsoft.AspNetCore.OData.Query.ODataQueryRequestMiddleware {
-	public ODataQueryRequestMiddleware (Microsoft.AspNetCore.OData.Query.IODataQueryRequestParser queryRequestParser, Microsoft.AspNetCore.Http.RequestDelegate next)
```
xuzhg, line 1372: >queryRequestParser [](start = 95, length = 18)

It's weird,  It works without updating .NET 5 public API baseline?

test/Microsoft.AspNetCore.OData.E2E.Tests/QueryRequest/DollarQueryTests.cs
gathogojr: > would you please share a link for that?
> 
> In reply to: [610427377](https://github.com/OData/AspNetCoreOData/pull/121#discussion_r610427377) [](ancestors = 610427377)

@xuzhg http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#_Toc31360953

test/Microsoft.AspNetCore.OData.Tests/PublicApi/Microsoft.AspNetCore.OData.PublicApi.NetCore31.bsl
gathogojr: I updated the baseline for both .NET 3.1 and 5

Issue #74, opened 2021-01-21
$count in $filter expression doesn't work


SEnglishSMP:
According to the [section 4.8 of the OData v4.01 URL Conventions spec](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_AddressingtheCountofaCollection), the following query should be supported:
```
 http://host/service/Categories?$filter=Products/$count gt 0
```
However, attempting this query with an AspNetCoreOData project returns the an error with the following message:
> The query specified in the URI is not valid. Validating OData QueryNode of kind Count is not supported by FilterQueryValidator.

This is a bug that was apparently fixed in the OData Lib (see https://github.com/OData/odata.net/issues/440) but is still an issue here.

Leandro-Albano:
I noticed that this still happening.
Some [unit tests](https://github.com/OData/AspNetCoreOData/blob/master/test/Microsoft.AspNetCore.OData.Tests/Query/FilterQueryOptionTest.cs#L46) exercise this behavior, but the issue seems to be because on [FilterQueryValidator](https://github.com/OData/AspNetCoreOData/blob/master/src/Microsoft.AspNetCore.OData/Query/Validator/FilterQueryValidator.cs#L792) the `QueryNodeKind.Count` falls into the default switch option and throws `NotSupported`.

@xuzhg  I'm opening a PR for this, could you tell me if you see any side effect of changing that?

PR #127, opened 2021-04-02
#74 count in filter throws not supported

Leandro-Albano:
- Added `QueryNodeKind.Count` as a valid option on  `ValidateSingleValueNode`
- Added e2e tests for $count in $filter
- Fixed missing Id's in customer list on ODataRoutingSample

### DIFF
--- a/test/Microsoft.AspNetCore.OData.E2E.Tests/Query/ActionResult/ActionResultTests.cs
+++ b/test/Microsoft.AspNetCore.OData.E2E.Tests/Query/ActionResult/ActionResultTests.cs
@@ -47,7 +47,7 @@ namespace Microsoft.AspNetCore.OData.E2E.Tests.Query.ActionResult
         public async Task ActionResultODataPathReturnsExpansion()
         {
             // Arrange
-            string queryUrl = string.Format("odata/Customers?$expand=books");
+            string queryUrl = "odata/Customers?$expand=books";
             HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Get, queryUrl);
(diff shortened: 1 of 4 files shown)


test/Microsoft.AspNetCore.OData.E2E.Tests/Query/ActionResult/ActionResultTests.cs
```diff
@@ -89,5 +89,30 @@ public async Task ActionResultODataPathReturnsBaseWithoutExpansion()
...
+        /// <summary>
+        /// $filter should work with $count.
+        /// </summary>
+        [Theory]
+        [InlineData(1, 1)]
+        [InlineData(0, 0)]
+        public async Task ActionResultShouldAllowCountInFilter(int filterParam, int expectedItemsCount)
+        {
+            // Arrange
+            string queryUrl = string.Format($"odata/Customers?$filter=Books/$count eq {filterParam}");
```
xuzhg, line 102: >string.Format [](start = 30, length = 13)

It seems we don't need string.Format here

xuzhg:
>             string queryUrl = string.Format("odata/Customers");

would you please help to clear others in this file?

---
Refers to: test/Microsoft.AspNetCore.OData.E2E.Tests/Query/ActionResult/ActionResultTests.cs:76 in 1a4f57b. [](commit_id = 1a4f57b732b0091167fcacc1750b24196758911c, deletion_comment = False)

xuzhg (approved):
:shipit:

Leandro-Albano:
Good point! I added another commit.

Issue #110, opened 2021-03-12
Newtonsoft support


max-2k20:
AspNetCoreOData support newtonsoft serialization provider?

xuzhg:
@max-2k20 Could you please share more detail?
AspNetCoreOData has OData formatter registered to serialize as OData payload. It doesn't register other formatter. 
ASP.NET Core has the extension method that you can register newtonsoft serializer.

max-2k20:
@xuzhg i use `services.AddMvc().AddNewtonsoftJson()` for register NewtonsoftJson deserelization for my asp.net core project, but 
AspNetCoreOData use System.Text.Json for data serelization

xuzhg:
@max-2k20 

> @xuzhg i use `services.AddMvc().AddNewtonsoftJson()` for register NewtonsoftJson deserelization for my asp.net core project, but
> AspNetCoreOData use System.Text.Json for data serelization

Would you please point me which part related to `AspNetCoreOData use System.Text.Json for data serelization`? 
Do you mean these converter? 'https://github.com/OData/AspNetCoreOData/blob/master/src/Microsoft.AspNetCore.OData/ODataServiceCollectionExtensions.cs#L140-L142' 
If yes, I have to say I don't provide the newtonsoft serialization. In my design, `System.Text.Json` is our solution.

snappyfoo:
@xuzhg thisnpossible relates to https://github.com/OData/AspNetCoreOData/issues/20 which has been ignored for a while now. Is there anything more you need?

xuzhg:
@snappyfoo Would you please share some links for me to "if you could just provide your own accompanying package that provided these Newtonsoft support with these extra converters. Same way ASP.NET Core does, Swagger, etc. etc."

snappyfoo:
@xuzhg Sure. To clarify what I mean, following the release of STJ, the core packages for those products switched to STJ and stopped supporting Newtonsoft by default, but supplied separate packages that people can add to their project if they are sticking with Newtonsoft.

**ASP.NET Core MVC**: https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc.NewtonsoftJson/

This provides an `IMvcBuilder.AddNewtonsoftJson` extension method for chaining on to e.g. `services.AddControllers()` in startup.

**Swashbuckle (Swagger)**: https://www.nuget.org/packages/Swashbuckle.AspNetCore.Newtonsoft/

[...]

PR #137, opened 2021-04-08
Add Newtonsoft.Json converter

xuzhg:
See issue #110.

This PR is to add a new project to enable Newtonsoft.JSON converter for:

1) SelectExpandWrapper
2) DynamicWrapper
3) PageResult
4) SingleResult

Test added.


omnilogix:
@xuzhg - Thanks for the very fast response!.  Does your pull request allow the JsonSerialization to be added during services setup? 

The Microsoft.AspNetCore.Mvc.Newtonsoft package allows configuration like this:

```
 services.AddControllers()
              .AddNewtonsoftJson(options => 
            {
                options.SerializerSettings.DefaultValueHandling = Newtonsoft.Json.DefaultValueHandling.Ignore;
                options.SerializerSettings.ReferenceLoopHandling = Newtonsoft.Json.ReferenceLoopHandling.Ignore;
                options.SerializerSettings.ContractResolver = WebApiJsonResolver.Instance;
            });
 
```

I would think an option like this would be perfect:
```
            services.AddOData(opt =>
            {
                opt.AddModel("odata", GetEdmModel());
                opt.Count().Filter().OrderBy().Expand().SetMaxTop(1).Select();
            }) 
            .AddNewtonsoftJson(options => 
            {
                options.SerializerSettings.DefaultValueHandling = Newtonsoft.Json.DefaultValueHandling.Ignore;
                options.SerializerSettings.ReferenceLoopHandling = Newtonsoft.Json.ReferenceLoopHandling.Ignore;
                options.SerializerSettings.ContractResolver = WebApiJsonResolver.Instance;
            });

```

xuzhg:
@omnilogix See the commit 2. It's not finished yet. So, let me know your thought.

omnilogix:
@xuzhg - Yes!  This looks very promising.   Using the MvcNewtonsoftJsonOptions is a great idea.  I think the default converters should be added internally so no startup.cs configuration needs to be done unless customizing. 

This to use Newtonsoft defaults:
`services.AddOData(opt => opt.Select().Filter()).AddNewtonsoftJson();`

This to customize the serialization behavior:
```
services.AddOData(opt => opt.Select().Filter()).AddNewtonsoftJson(
                options =>
                {
                    options.SerializerSettings.DefaultValueHandling = Newtonsoft.Json.DefaultValueHandling.Ignore;
                    options.SerializerSettings.ReferenceLoopHandling = Newtonsoft.Json.ReferenceLoopHandling.Ignore;
                    options.SerializerSettings.ContractResolver = MyCustomContractResolver;
                });
```

 Please ping me when a build is available to test.

omnilogix:
@xuzhg +1 for supporting DI of the controller level

omnilogix:
@xuzhg Any idea when a package will be available to test?

xuzhg:
@omnilogix Please forgive me, i have to focus on other parts. I will resume it some time this week.
Actually, It's still in development. When the development is done, i'd update the build definition to build the nightly package.

I will update you when i am ready.

Besides, I also shared it with .NET Core team, the feedback is that It could be better to have: 

```
services.AddControllers()
                .AddNewtonsoftJson(option => options.AddODataJsonConverters());
```
I am not sure now, will investigate it more.

xuzhg:
Merged at c11b4d825264dfd89083d9549c6d629d652fdca2

It's now available at nightly:
https://www.myget.org/feed/webapinetcore/package/nuget/Microsoft.AspNetCore.OData.NewtonsoftJson/8.0.0-Nightly202104151654

@omnilogix Would you please help test it?

omnilogix:
@xuzhg I have installed and tested.  It does not seem to use the Newtonsoft options.  Is this the correct usage?
 ```
services.AddOData(opt =>
            {
                opt.AddModel("odata", GetEdmModel());
                opt.Count().Filter().OrderBy().Expand().SetMaxTop(1).Select();
            }).AddNewtonsoftJson( options =>
            {
                options.SerializerSettings.DefaultValueHandling = Newtonsoft.Json.DefaultValueHandling.Ignore;
                options.SerializerSettings.ReferenceLoopHandling = Newtonsoft.Json.ReferenceLoopHandling.Ignore;
                options.SerializerSettings.ContractResolver = WebApiJsonResolver.Instance;
            });
```

Also, there is a dependency on Microsoft.AspNetCore.Mvc.NewtonsoftJson.  Is this correct?

**Comment by xuzhg** [comment]
@omnilogix do you mind to share your whole project to me to dig more?

by the way, i have the sample project that you can play: https://github.com/OData/AspNetCoreOData/tree/master/sample/ODataNewtonsoftJsonSample

Let me know which way can work.

**Comment by omnilogix** [comment]
@xuzhg I checked your sample and it is working.  The problem exists when using this:
`                opt.AddModel("odata", GetEdmModel());
`
I have modified your sample so you can see the problem

[AspNetCoreOData-master-Problem1.zip](https://github.com/OData/AspNetCoreOData/files/6320538/AspNetCoreOData-master-Problem1.zip)

**Comment by prashantaggarwal1990** [comment]
Can we add the same into .Net Core 3.1?

**Comment by xuzhg** [comment]
@prashantaggarwal1990 is there any blocker blocking you update to .NET 5?

**Comment by xuzhg** [comment]
@omnilogix 
If you register an OData route. You will have the following two routes: (use latest nightly, send "$odata" request):

image

in this case:
* ~/odata/WeatherForecast   ==> returns OData, will never use Newtonsoft.json
* ~/WeatherForecast  ==> returns normal JSON, will use Newtonsoft.JSON.

I updated the sample, please take a look and let me know your thoughts.

**Comment by omnilogix** [comment]
> @omnilogix
> If you register an OData route. You will have the following two routes: (use latest nightly, send "$odata" request):
> 
> image
> 
> in this case:
> 
> * ~/odata/WeatherForecast   ==> returns OData, will never use Newtonsoft.json
> * ~/WeatherForecast  ==> returns normal JSON, will use Newtonsoft.JSON.
> 
> I updated the sample, please take a look and let me know your thoughts.

@xuzhg - I don't think I understand.  The point of Newtonsoft extension is to get Newtonsoft.Json for odata request.  Can I get newtonsoft.json with Odata result?

**Comment by omnilogix** [comment]
@xuzhg - I tried the sample.  As you say, OData endpoint does not use Newtonsoft.json, but that is exactly what is needed.

OData query should create OData result set and serialize using Newtonsoft.json and return the Newtonsoft.json formatted result instead of MS.json result.   If I use NON-Odata endpoint, there is no purpose for OData.

**Comment by xuzhg** [comment]
@omnilogix if you use odata routing, ASP.NET Core OData uses the built-in JSON (not System.Text.Json, not Newtonsoft.JSON) writer and reader to write and read the payload. The payload should be OData payload, that's defined in [OData spec](http://docs.oasis-open.org/odata/odata-json-format/v4.01/odata-json-format-v4.01.html).

Besides, we support non-odata endpoint,  (actually, if you only want to basic routing, there's no need for OData newtonsoft converter.) BUT, BUT, If you want to use OData query, you should have OData newtonsoft converter registered. 

For example: we have the following two requests with "$select" (You can try using my sample)

* ~/odata/WeatherForecast?$select=TemperatureC  
* ~/WeatherForecast?$select=TemperatureC 

The responses are different payloads. 

If you don't enable newtonsoft converter, you will get an internal "selectexpandwrapper" payload, that's not a correct payload. you can remove "AddnewtonsoftJson()" and try.

**Comment by xuzhg** [comment]
By the way, if you want to return "normal JSON", you can use "$format=application/json;odata.metadata=none" to get the plain data without control metadata.

**Comment by omnilogix** [comment]
@xuzhg - The OData spec is quite large, so forgive me if I don't understand the payload.    If I query ~/odata/weatherforecast?$select=Summary,TemperatureC, I get:

```
{
  "@odata.context": "http://localhost:12197/odata/$metadata#WeatherForecast(Summary,TemperatureC)",
  "value": [
    {
      "TemperatureC": 0,
      "Summary": "Balmy"
    },
    {
      "TemperatureC": 0,
      "Summary": "Cool"
    },
    {
      "TemperatureC": 0,
      "Summary": "Warm"
    },
    {
      "TemperatureC": 0,
      "Summary": "Hot"
    },
    {
      "TemperatureC": 0,
      "Summary": "Cool"
    }
  ]
}
```

What I need is :

```
{
  "@odata.context": "http://localhost:12197/odata/$metadata#WeatherForecast(Summary,TemperatureC)",
  "value": [
    {
      "Summary": "Balmy"
    },
    {
      "Summary": "Cool"
    },
    {
      "Summary": "Warm"
    },
    {
      "Summary": "Hot"
"TemperatureC": 35,
    },
    {
      "Summary": "Cool"
    }
  ]
}
```

The Newtonsoft.json serializer lets me choose to exclude the TemperatureC when it is the default value of 0.   This significantly reduces the payload size and serialization time.  That is why it is needed.  Otherwise, I can see no value in the way it is implemented now.  I would just use regular MVC.

Since this is an extension, shouldn't I have a choice about how to apply it to the data?   The OData header can still be sent, but let me decide what data to serialize in the "value" part of the result set.

**Comment by omnilogix** [comment]
@xuzhg With the microsoft json, we can set a flag to ignore null values and this does not violate the OData spec.  The Newtonsoft.json lets me choose other conditions like omit empty strings, omit default values, omit empty arrays, etc.  I don't think that this is any different situation, just much more flexible.   What is needed is for the json serializer that writes the "value" section of the OData result be configurable.

**Comment by prashantaggarwal1990** [comment]
We have a project in core 3.1  which is recently migrated, response from the OData API is always returning in Pascal case always even request which contains custom models does not parsing Json objects which are in camel case. I guess Newtonsoft will resolve this problem. We can’t migrate to 5 for now, is there any work around for 3.1.
________________________________

**Comment by xuzhg** [comment]
@omnilogix

omit null value (OData way) is still in draft PR. (It's two years PR, I am sorry for that. 😞 )

In your requirement, you just need a customized OData resource serializer, 

Here's a prototype workaround

1) Create your own ResourceSerializer and Derived from ODataResourceSerializer, inject the setting
2) Derived from DefaultODataSerializerProvider, override GetEdmTypeSerializer() and return your own resource serializer 
3) Override CreateResource() in your own ODataResourceSerializer, make the decision for your property based on conditions?

You can try and let me know any question.

 next week in case i am ok, i can help to create a sample. Or maybe i can expose OData formatter setting to control the serialize.

**Comment by omnilogix** [comment]
@xuzhg 
Thanks for the possible workaround, this might be a solution that I can use, but I would like to use the Newtonsoft.Json library to perform the serialization.  I already include it for many other data operations in my apps.  It is well tested and reliable.

Can you speak with the team that builds Microsoft.AspNetCore.Mvc.NewtonsoftJson and ask how they allow custom serialization?  Their solution works very well for standard WebApi.  We need the same functionality for results of OData query.
Or possibly speak with @JamesNK (author of Newtonsoft.json) about how to implement with OData?

**Comment by omnilogix** [comment]
@xuzhg - How do I inject the serializer?  In startup.cs I tried

`services.AddSingleton(typeof(CustomODataSerializerProvider));`

and
           
```
services.AddOData(opt => opt.Select().Filter().AddModel("odata", GetEdmModel(), builder =>
            {
                builder.AddService(Microsoft.OData.ServiceLifetime.Singleton, typeof(DefaultODataSerializerProvider), x => new CustomODataSerializerProvider(x));
            }));
```



but it does not get called

**Comment by omnilogix** [comment]
@xuzhg - I figured out how to register the service.  Although the workaround is not the solution I was hoping for, I think the following will work for my use case:

**Custom Serializer:**

```
namespace ODataNewtonsoftJsonSample
{
    public class CustomODataSerializerProvider : DefaultODataSerializerProvider
    {
        public CustomODataSerializerProvider(IServiceProvider container) : base(container) {}

        public override ODataEdmTypeSerializer GetEdmTypeSerializer(IEdmTypeReference edmType)
        {
            if (edmType.Definition.TypeKind == EdmTypeKind.Entity)
                return new CustomEntitySerializer(this);

            return base.GetEdmTypeSerializer(edmType);
        }
    }


    public class CustomEntitySerializer : ODataResourceSerializer
    {
        public CustomEntitySerializer(ODataSerializerProvider serializerProvider) : base(serializerProvider) {}

        public override Microsoft.OData.ODataProperty CreateStructuralProperty(IEdmStructuralProperty structuralProperty, ResourceContext resourceContext)
        {
            Microsoft.OData.ODataProperty property = base.CreateStructuralProperty(structuralProperty, resourceContext);

            var t = Type.GetType(structuralProperty.DeclaringType.FullTypeName());

            var attr = t.GetProperty(property.Name)
                        .GetCustomAttributes(typeof(DefaultValueAttribute), false)
                        .Cast<DefaultValueAttribute>()
                        .FirstOrDefault();

            if (attr?.Value.Equals(property.Value) == true)
                return null;

            return property;
        }

    }
}
```

**Integrated in Startup.cs like this:**

```
            services.AddOData(opt => opt.Select().Filter().AddModel("odata", GetEdmModel(), builder =>
            {
                builder.AddService(Microsoft.OData.ServiceLifetime.Singleton, typeof(ODataSerializerProvider), x => new CustomODataSerializerProvider(x));
            }));


```

For performance, is the underlying data type available?  I would like to have access to the serialized type so that I can check for custom attributes.  The following code is not performant.

`var t = Type.GetType(structuralProperty.DeclaringType.FullTypeName());`

omnilogix:
@xuzhg - The problem with the ResourceSerializer approach is that I need to create 2 different serializers for OData and non-OData endpoints which is bad design and requires more unit tests, etc.

Why don't we have 1 place to register a custom json serializer - and it just works for all routes?

PR #148, opened 2021-04-16
Incorrect Assert in ConvertPrimitiveValue

### LINKED ISSUE #131: Incorrect Assert in ConvertPrimitiveValue
Hitting an incorrect assert in `ConvertPrimativeValue` in [EdmPrimitiveHelper.cs](https://github.com/OData/AspNetCoreOData/blob/e1ded5dc6120c7871b7db1c4938d9b06f6884788/src/Microsoft.AspNetCore.OData/Edm/EdmPrimitiveHelper.cs#L127
)
```
Contract.Assert(type == typeof(uint) || type == typeof(ushort) || type == typeof(ulong));
```

The controller has a signature of:

```
public async Task<ActionResult> Delete([FromODataUri] int key)
```

The URI sent is:

```
DELETE http://localhost:5000/odata/Certifications('PDMSDT')
```

The result is that the `type`  is `int32` and the `value` is `"PDMSDT"` .

This should **NOT** be an assert causing a 500 status code as it is not a library bug, it is a user sending an incorrect value that should cause a `ValidationException`, such as `PropertyMustBeInt32` (or some such).

In `ConvertPrimativeValue` the code should be expanded to always call `Convert.ChangeType` and catch the well defined exceptions: `InvalidCastException`, `FormatException`, and `OverflowException`. _[Not Compiled]_

```
    // Note that we are not casting the return value to nullable<T> as even if we do it
    // CLR would unbox it back to T.
    try
    {
        return Convert.ChangeType(value, type, CultureInfo.InvariantCulture);
    }
    catch (Exception ex)
    {
        switch (ex)
        {
            case InvalidCastException:
            case FormatException:
            case OverflowException:
                throw new ValidationException(Error.Format(SRResources.PropertyInvalidType));
            default:
                throw;
        }
    }
```

### LINKED ISSUE #140: LoggerError unnecessarily adds an additional frame to call stack
In LoggerError:

```
if (logger == null)
{
    throw ex;
}

logger.LogError(ex, String.Empty);
```

The use of `throw ex` will cause the exception stack from the `throw` to have an extra frame for `LoggerError` that is not present in the `logger`.

The correct approach is to use `ExceptionDispatchInfo` to capture and rethrow with the correct frames.

```
if (logger == null)
{
    ExceptionDispatchInfo.Capture(ex).Throw();
}

logger.LogError(ex, String.Empty);
```

Now the exception will have the stack from where it was thrown, not including `LoggerError`.

dxrdxr:
Incorrect Assert in ConvertPrimitiveValue
LoggerError unnecessarily adds an additional frame to call stack
Fixes #131 
Fixes #140 

Change `ConvertPrimitiveValue` to simply call `Convert.ChangeType` and catch the common exceptions that indicate a type conversion is not allowed, then throw a corresponding `ValidationException`.

Update test cases to cover almost all of the code paths.

**Note**: `ConvertPrimitiveValue_NonStandardPrimitives_Data` is shared by tests for other methods with slightly different semantics, so the two tests vectors cannot be merged.
 
Rewrite the Throw test cases to use a common simplified test method.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Edm/EdmPrimitiveHelper.cs
+++ b/src/Microsoft.AspNetCore.OData/Edm/EdmPrimitiveHelper.cs
@@ -124,11 +124,24 @@ namespace Microsoft.AspNetCore.OData.Edm
                 }
                 else
                 {
-                    Contract.Assert(type == typeof(uint) || type == typeof(ushort) || type == typeof(ulong));
-
-                    // Note that we are not casting the return value to nullable<T> as even if we do it
-                    // CLR would unbox it back to T.
-                    return Convert.ChangeType(value, type, CultureInfo.InvariantCulture);
+                    try
+                    {
+                        // Note that we are not casting the return value to nullable<T> as even if we do it
+                        // CLR would unbox it back to T.
+                        return Convert.ChangeType(value, type, CultureInfo.InvariantCulture);
+                    }
+                    catch (InvalidCastException)
+                    {
+                        throw new ValidationException(Error.Format(SRResources.PropertyCannotBeConverted, type));
+                    }
+                    catch (FormatException)
+                    {
+                        throw new ValidationException(Error.Format(SRResources.PropertyUnrecognizedFormat, type));
+                    }
+                    catch (OverflowException)
+                    {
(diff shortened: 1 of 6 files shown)


xuzhg (approved):
:shipit:

xuzhg:
Looks good to me.

PR #174, opened 2021-05-20
Migrate to net framework

marabooy:


src/Microsoft.AspNetCore.OData.Nightly.nuspec
```diff
@@ -4,8 +4,8 @@
     <id>Microsoft.AspNetCore.OData</id>
     <title>Microsoft ASP.NET Core 3.x and 5.x for OData v4.0</title>
     <version>$VersionFullSemantic$-Nightly$NightlyBuildVersion$</version>
-    <authors>Microsoft</authors>
-    <copyright>&#169; Microsoft Corporation. All rights reserved.</copyright>
```
xuzhg, line 8: Does this change block the release for aspnetcore.odata?

xuzhg: As we discussed, we only port newtonsoftjson project to .net foundation? right?

marabooy: Hi Sam, currently this should not be a blocker as the old pipeline will be running to sign the current binary.
Once we move the 7x series it should proceed normally. I will revert this to the original until we move the 7x.

azure_pipelines.yml
```diff
@@ -0,0 +1,84 @@
...
+  - task: NuGetToolInstaller@1
+    displayName: Use NuGet >=5.2.0
+    inputs:
+      versionSpec: '>=5.2.0'
+      checkLatest: true
+  - task: UseDotNet@2
+    displayName: Use .NET Core sdk 3.1
+    inputs:
+      version: 3.x
+  - task: UseDotNet@2
```
xuzhg, line 29: Can we move the yml files to a subfolder?

xuzhg (approved):

PR #187, opened 2021-06-02
Fix behaviour for Preflight requests on the ODataBatchMiddleware

marabooy:


src/Microsoft.AspNetCore.OData/Batch/ODataBatchMiddleware.cs
```diff
@@ -53,10 +53,15 @@ public async Task Invoke(HttpContext context)
             {
                 throw Error.ArgumentNull(nameof(context));
             }
-
             string prefixName;
             ODataBatchHandler batchHandler;
-            if (_batchMapping != null && _batchMapping.TryGetPrefixName(context, out prefixName, out batchHandler))
+
+            // The batch middleware should not handle the options requests for cors to properly function.
+            bool isPreFlight = HttpMethods.IsOptions(context.Request.Method);
```
xuzhg, line 60: From OData spec, 
`Batch requests are submitted as a single HTTP POST request to the batch endpoint of a service, located at the URL $batch relative to the service root.`

So, can we modify to allow "POST" HTTP method only? then we needn't compare "Options" HTTP method?

xuzhg (changes requested):

habbes (approved):

xuzhg (approved):

PR #200, opened 2021-06-15
Add ODataResourceSerializer tests and fixes for dynamic property support

habbes:
- Add more tests for dynamic property support in `ODataResourceSerializer`.
- Fixed a potential bug where a dynamic property with null value could be added to the resource even if it had the same name as an existing declared property
- Allowed null dynamic properties by default (see explanation below for more context)

There's a test called `CreateResource_Works_ToAppendNullDynamicProperties_ForOpenEntityType()`, based on the name, I assume that it should verify that dynamic properties with null values are added to the resource. However, the test actually verifies that dynamic properties with null values are actually *not* included in the resource. So I was confused about what the expected behaviour is.

Looking at the existing code, it seems that null dynamic properties are not allowed by default. They are only allowed when the resource is for an EdmDeltaComplexObject or an EdmDeltaResourceObject. So the name of the test is confusing

Furthermore, in WebApi 7.x, null dynamic properties can also be enabled by setting an option (`resourceContext.InternalRequest.Options.NullDynamicPropertyEnabled`), but this is not checked in 8.x

After an offline discussion with @xuzhg, we concluded to allow null dynamic properties by default. I updated the code and tests to reflect that.

### DIFF
--- a/test/Microsoft.AspNetCore.OData.Tests/Scenarios/OpenComplexTypeTests.cs
+++ b/test/Microsoft.AspNetCore.OData.Tests/Scenarios/OpenComplexTypeTests.cs
@@ -45,7 +45,7 @@ namespace Microsoft.AspNetCore.OData.Tests.Scenarios
             Assert.Equal("My Dynamic Place", result["Place"]);
             Assert.Equal("2c1f450a-a2a7-4fe1-a25d-4d9332fc0694", result["Token"]);
             Assert.Equal("2015-03-02", result["Birthday"]);
-            Assert.Null(result["Region"]);
+            Assert.Equal(JTokenType.Null, result["Region"].Type);
         }
 
         [Fact]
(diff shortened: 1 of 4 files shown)


test/Microsoft.AspNetCore.OData.Tests/Scenarios/OpenComplexTypeTests.cs
```diff
@@ -45,7 +45,7 @@ public async Task OpenComplexType_SimpleSerialization()
             Assert.Equal("My Dynamic Place", result["Place"]);
             Assert.Equal("2c1f450a-a2a7-4fe1-a25d-4d9332fc0694", result["Token"]);
             Assert.Equal("2015-03-02", result["Birthday"]);
-            Assert.Null(result["Region"]);
+            Assert.Equal(JTokenType.Null, result["Region"].Type);
```
habbes, line 48: I've updated these tests assuming that dynamic properties with null values should be included in the response. `result["Region"]` returns null only when the `Region` property does exist in the JSON object. If the `Region` property exists but has a null value, then `result["Region"]` returns a `JToken` object whose `Type` property is `JTokenType.Null`.

xuzhg (approved):
:shipit:

PR #201, opened 2021-06-15
OData route Debug controller formatting

chrisspre:
small changes to HTML template in the OData route controller

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Routing/ODataRouteDebugMiddleware.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/ODataRouteDebugMiddleware.cs
@@ -31,22 +31,13 @@ namespace Microsoft.AspNetCore.OData.Routing
         /// <param name="next">The next middleware.</param>
         public ODataRouteDebugMiddleware(string routePattern, RequestDelegate next)
         {
-            _next = next ?? throw Error.ArgumentNull(nameof(next)); ;
-
             if (routePattern == null)
             {
                 throw Error.ArgumentNull(nameof(routePattern));
             }
-
             // ensure _routePattern starts with /
-            if (routePattern.StartsWith('/'))
-            {
-                _routePattern = routePattern;
-            }
-            else
-            {
-                _routePattern = $"/{routePattern}";
-            }
+            _routePattern = routePattern.StartsWith('/') ? routePattern : $"/{routePattern}";
+            _next = next ?? throw Error.ArgumentNull(nameof(next)); ;
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


xuzhg:
@chrisspre Thanks for your contribution.

ODataEndpointController.cs in the sample project is for community reference. That's not in the product code.

Instead, the similar endpoint controller is replaced with middelware in the product code at:

The 
https://github.com/OData/AspNetCoreOData/blob/master/src/Microsoft.AspNetCore.OData/Routing/ODataRouteDebugMiddleware.cs

Any interesting to help this middleware?

chrisspre:
@xuzhg  send an update with changes to ODataRouteDebugMiddleware.cs. Please have a look. 

There was something in the HTML template about prefixing the templates with ~/ that I didn't fully grasp. but the HTML page works fine without.

src/Microsoft.AspNetCore.OData/Routing/ODataRouteDebugMiddleware.cs
```diff
@@ -30,26 +31,22 @@ internal class ODataRouteDebugMiddleware
         /// <param name="next">The next middleware.</param>
         public ODataRouteDebugMiddleware(string routePattern, RequestDelegate next)
         {
+            _next = next ?? throw Error.ArgumentNull(nameof(next)); ;
```
xuzhg, line 34: I'd prefer to check/verify the input parameters using the order of their definition.

src/Microsoft.AspNetCore.OData/Routing/ODataRouteDebugMiddleware.cs
```diff
@@ -30,26 +31,22 @@ internal class ODataRouteDebugMiddleware
...
             {
                 throw Error.ArgumentNull(nameof(routePattern));
             }
 
-            if (next == null)
-            {
-                throw Error.ArgumentNull(nameof(next));
-            }
-
-            if (routePattern.StartsWith("/", StringComparison.OrdinalIgnoreCase))
```
xuzhg, line 43: without using "StringComparision" might be building warning.

chrisspre: The new code is written as `routePattern.StartsWith('/')` using a character which has no StringComparison parameter and since '/' has no casing so I think it wouldn't change anything anyways.

src/Microsoft.AspNetCore.OData/Routing/ODataRouteDebugMiddleware.cs
chrisspre: fixed

xuzhg (approved):

PR #208, opened 2021-06-29
OData WebApi 8.0 code improvements for routing feature

gathogojr:
OData WebApi 8.0 code improvements for routing feature

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Routing/Conventions/RefRoutingConvention.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/Conventions/RefRoutingConvention.cs
@@ -86,10 +86,7 @@ namespace Microsoft.AspNetCore.OData.Routing.Conventions
 
             // Find the navigation property if have
             IEdmNavigationProperty navigationProperty = null;
-            if (property != null)
-            {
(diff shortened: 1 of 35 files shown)


src/Microsoft.AspNetCore.OData/Routing/Conventions/RefRoutingConvention.cs
```diff
@@ -86,10 +86,7 @@ public bool AppliesToAction(ODataControllerActionContext context)
 
             // Find the navigation property if have
             IEdmNavigationProperty navigationProperty = null;
-            if (property != null)
-            {
-                navigationProperty = declaringType.DeclaredNavigationProperties().FirstOrDefault(p => p.Name == property);
-            }
+            navigationProperty = declaringType.DeclaredNavigationProperties().FirstOrDefault(p => p.Name == property);
```
gathogojr, line 89: No need for null check since that is catered for in [Line 82](https://github.com/OData/AspNetCoreOData/pull/208/files#diff-f962cd7ff643bbbe205141cbc4be99bc69118af26381baf8731694e2b96409edR82) above

xuzhg (approved):

PR #209, opened 2021-06-29
Improvement to (de)serialization code

habbes:
- Create `IODataDeserializer` interface, (I kept the `ODataDeserializer` abstract class as well)
- Create `IODataEdmDeserializer` interface (also kept `ODataEdmDeserializer`)
- Create `IODataDeserializerProvider` interface and remove the `ODataDeserializerProvider` abstract class
- Rename `DefaultODataDeserializerProvider` to `ODataDeserializerProvider`
- Create `IODataSerializer` interface (I also kept the `ODataSerializer` abstract class)
- Create `IODataEdmTypeSerializer` interface (still kept the `ODataEdmTypeSerializer` abstract class)
- Move `ODataEdmTypeSerializer.CreateProperty` to an internal helper extension method `ODataSerializerPropertyHelper.CreateProperty`. Moving it to the interface would require it to be public and I didn't want to add it to the public API (it also complicated some of the tests that rely on mocking `CreateODataValue()`). I could also have kept it internal to the abstract class and not move it to the interface, but that would mean that `serializerProvider.GetEdmTypeSerializer()` would need to return the abstract class instead of the interface in order to work, which would make the interface useless. I figure moving it to a helper extension method would have less impact on the code that used the method and would still keep it internal. The downside is that the extensions contain the base implementation for `ODataEdmTypeSerializer` as well as the overridden implementation for `ODataCollectionSerializer` and uses an if statement with type matching to decide which implementation to call. Not sure if this a clean approach.
- Create `IODataSerializerProvider` interface and remove the `ODataSerializerProvider` abstract class
- Rename `DefaultODataSerializerProvider` to `ODataSerializerProvider`
- Updated public API files (.bsl)
- Updated PublicAPI.Unshipped.txt

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataCollectionSerializer.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataCollectionSerializer.cs
@@ -204,6 +204,7 @@ namespace Microsoft.AspNetCore.OData.Formatter.Serialization
             return value;
         }
 
+        /// <inheritdoc />
         public override ODataProperty CreateProperty(object graph, IEdmTypeReference expectedType, string elementName,
             ODataSerializerContext writeContext)
         {
--- a/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataCollectionSerializer.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataCollectionSerializer.cs
@@ -204,26 +204,6 @@ namespace Microsoft.AspNetCore.OData.Formatter.Serialization
             return value;
         }
 
-        /// <inheritdoc />
-        public override ODataProperty CreateProperty(object graph, IEdmTypeReference expectedType, string elementName,
-            ODataSerializerContext writeContext)
-        {
-            Contract.Assert(elementName != null);
-            var property = CreateODataValue(graph, expectedType, writeContext);
-            if (property != null)
-            {
-                return new ODataProperty
-                {
-                    Name = elementName,
-                    Value = property
-                };
-            }
-            else
-            {
-                return null;
-            }
-        }
-
         /// <summary>
         /// Adds the type name annotations required for proper json light serialization.
         /// </summary>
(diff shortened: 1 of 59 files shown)


src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataCollectionSerializer.cs
```diff
@@ -204,25 +204,6 @@ public ODataCollectionSerializer(ODataSerializerProvider serializerProvider)
             return value;
         }
 
-        internal override ODataProperty CreateProperty(object graph, IEdmTypeReference expectedType, string elementName,
```
xuzhg, line 207: why do we remove this override?

xuzhg: I see you move it to build as an extension. But, it's no way to override in other serializer?

habbes: I made `IODataEdmTypeSerializer` interface and made the `serializerProvider.GetEdmTypeSerializer` return the interface instead of abstract class. So for the method to be accessible from the interface, then I would need to define the method `CreateProperty` on the interface as well. And in this case I would need to make the method public instead of internal. I wasn't sure whether it would be a good idea to add it to the public API. But I could go ahead and make it public. I did actually try to make it public but got some issues with some of the tests that mock `GetODataValue()`. I guess the tests could be worked around. Let me know if adding the method to the public API is okay.

I moved it to an extension method in a helper class so that I could keep it internal without breaking the interface. I have doubt about this solution because I had to use pattern matching to distinguish between `ODataCollectionSerializer` and other `EdmTypeSerializer` implementations instead of relying on polymorphism.

xuzhg: Yes. I know. Let's continue on the refactor on this part.

xuzhg:
@habbes It seems there's still test case failing? Will you take a look?

habbes:
> @habbes It seems there's still test case failing? Will you take a look?

The tests are passing locally. I have re-run the build pipeline. Based on the error messages, I think it's flaky tests in the e2e suite

habbes:
@xuzhg I re-ran the tests and now they pass successfully.

xuzhg (approved):

xuzhg:
Merged at: c0b060d54a2bab0c975f63a09c097dcae42adf8a

PR #211, opened 2021-06-29
Move from nuget pack to dotnet pack.

marabooy:


pipelines/azure_pipelines_nightly.yml
```diff
@@ -9,6 +9,18 @@ resources:
...
 
+variables:
+  buildPlatform: 'Any CPU'
+  BuildPlatformDirName: 'AnyCpu'
+  buildConfiguration: 'Release'
+  rootDir: '$(Build.SourcesDirectory)'
+  signigConfigPath: '$(rootDir)\configs'
+  signingConfigFiles: '*.*'
+  productBinPath: '$(rootDir)\bin\$(BuildPlatformDirName)\$(buildConfiguration)'
+  productFiles: 'Microsoft.OData.Extensions.*?(*.dll|*.config|*.pdb)'
```
xuzhg, line 20: Microsoft.OData.Extensions. **  That's not a correct pattern for 8.0

xuzhg (changes requested):

pipelines/azure_pipelines_nightly.yml
xuzhg: Make sure only bring the variables used now into the yaml file. Otherwise, it will be a painpoint to maintain them.

xuzhg (approved):

PR #245, opened 2021-07-21
Routing conventions code improvements

gathogojr:
Routing conventions code improvements

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Routing/Conventions/ActionRoutingConvention.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/Conventions/ActionRoutingConvention.cs
@@ -47,7 +47,7 @@ namespace Microsoft.AspNetCore.OData.Routing.Conventions
         }
 
         /// <inheritdoc />
-        protected override bool IsOperationParameterMeet(IEdmOperation operation, ActionModel action)
+        protected override bool IsOperationParameterMatched(IEdmOperation operation, ActionModel action)
         {
             Contract.Assert(operation != null);
             Contract.Assert(action != null);
--- a/src/Microsoft.AspNetCore.OData/Routing/Conventions/NavigationRoutingConvention.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/Conventions/NavigationRoutingConvention.cs
@@ -57,15 +57,6 @@ namespace Microsoft.AspNetCore.OData.Routing.Conventions
 
             ActionModel action = context.Action;
 
-            // Filter by the action name.
-            // The action for navigation property request should follow up {httpMethod}{PropertyName}[From{Declaring}]
-            string actionName = action.ActionName;
-            string method = SplitActionName(actionName, out string property, out string declared);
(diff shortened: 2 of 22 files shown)


gathogojr:
@xuzhg Please note that the change from `IsOperationParameterMeet` to `IsOperationParameterMatched` meets the definition of a breaking change. If you'd like us to defer the name correction please advise.

src/Microsoft.AspNetCore.OData/Routing/Conventions/ActionRoutingConvention.cs
```diff
@@ -47,7 +47,7 @@ public override bool AppliesToAction(ODataControllerActionContext context)
         }
 
         /// <inheritdoc />
-        protected override bool IsOperationParameterMeet(IEdmOperation operation, ActionModel action)
+        protected override bool IsOperationParameterMatched(IEdmOperation operation, ActionModel action)
```
xuzhg, line 50: >IsOperationParameterMatched [](http://example.com/codeflow?start=32&length=27)

I agree to change it.

src/Microsoft.AspNetCore.OData/Routing/Conventions/NavigationRoutingConvention.cs
```diff
@@ -57,15 +57,6 @@ public virtual bool AppliesToAction(ODataControllerActionContext context)
 
             ActionModel action = context.Action;
 
-            // Filter by the action name.
```
gathogojr, line 60: Moved this block of code a little further below. No use doing the splitting before other conditions are fulfilled, i.e.,
```csharp
            // filter by action parameter
            IEdmEntityType entityType = navigationSource.EntityType();
            bool hasKeyParameter = action.HasODataKeyParameter(entityType);
            if (!(context.Singleton != null ^ hasKeyParameter))
            {
                // Singleton, doesn't allow to query property with key
                // entityset, doesn't allow for non-key to query property
                return false;
            }
```

test/Microsoft.AspNetCore.OData.Tests/Routing/Conventions/RefRoutingConventionTests.cs
```diff
@@ -95,7 +95,7 @@ public void SplitRefActionName_WorksAsExpected()
 
         [Theory]
         [MemberData(nameof(RefConventionTestData))]
-        public void SingletonRoutingConventionTestDataRunsAsExpected(string actionName, string[] templates)
+        public void RefRoutingConventionTestDataRunsAsExpected(string actionName, string[] templates)
```
xuzhg, line 98: >RefRoutingConventionTestDataRunsAsExpected [](http://example.com/codeflow?start=20&length=42)

Thanks for fixing it.

xuzhg:
I think this change is ok.

---
In reply to: [884403334](https://github.com/OData/AspNetCoreOData/pull/245#issuecomment-884403334) [](http://example.com/codeflow?ancestors=884403334,884403334)

xuzhg:
Looks good to me.

PR #280, opened 2021-08-18
Fix the $expand command, It was not working with the dynamicmodel sam…

danleydmello:
…ple. Its needs to be set on the ODataFeature.SetSelectExpandClause.


xuzhg:
Looks good to me. :shipit:

PR #295, opened 2021-09-03
Added Nested Property tests

dxrdxr:
The DeltaTests do not fully cover Nested Properties.

Added tests and increased code coverage 10+%


xuzhg (approved):

src/Microsoft.AspNetCore.OData/Deltas/DeltaOfT.cs
```diff
@@ -646,11 +641,6 @@ private bool TrySetPropertyValueInternal(string name, object value)
 
         private bool TrySetNestedResourceInternal(string name, object deltaNestedResource)
         {
-            if (name == null)
```
KenitoInc, line 649: Why are we deleting this null check?

src/Microsoft.AspNetCore.OData/Deltas/DeltaOfT.cs
```diff
@@ -616,11 +616,6 @@ private void CopyUnchangedDynamicValues(T targetEntity)
 
         private bool TrySetPropertyValueInternal(string name, object value)
         {
-            if (name == null)
```
KenitoInc, line 619: Why are we deleting this null check?

KenitoInc (commented):
A few comments

src/Microsoft.AspNetCore.OData/Deltas/DeltaOfT.cs
dxrdxr: This is dead code that cannot be reached and cannot be tested.
The method is private and can only be called from one other function that already has this check.

src/Microsoft.AspNetCore.OData/Deltas/DeltaOfT.cs
dxrdxr: This is dead code that cannot be reached and cannot be tested.
The method is private and can only be called from one other function that already has this check.

KenitoInc (approved):

src/Microsoft.AspNetCore.OData/Deltas/DeltaOfT.cs
gathogojr: We could have replaced it with `Debug.Assert`...

gathogojr (approved):
LGTM

PR #291, opened 2021-08-31
Redesign FilterBinder

KenitoInc:
The main changes:
- Wrap the required `FilterBinder` properties in a `FilterBinderContext`
- Make the `FilterBinder` implement `IFilterBinder`
- Inject `FilterBinder` as an implementation of `IFilterBinder`
- Make some `Bind` methods `protected` and `virtual`

Note
- The servicelifetime remains `Transient` since the `FilterBinder` has state. `FilterBinder` inherits from `ExpressionBinderBase` and relies on some of it's properties and methods. Therefore injecting `FilterBinder` as a `Singleton` will make it thread unsafe
- We need a separate work item to refractor how we use the `ExpressionBinderBase` 
- The `ExpressionBinderBase`  is also used by `ComputeBinder` and `AggregationBinder`

PR #296, opened 2021-09-03
Redesign SelectExpandBinder to allow customizations

### LINKED ISSUE #304: Allow to override behavior of SelectExpandBinder.
## Motivation and use cases

Our application has database (postgresql) tables like here
```SQL
CREATE TABLE my_table
(
    id SERIAL,
    some_value text,
    ...
    metadata jsonb
);
```
Entity framework allows to map System.Text.Json.JsonDocument to that json column. So our database object looks like
```c#
public class MyObject
{
    int Id { get; init; }
    string SomeValue { get; init; }
    ...
    JsonDocument Metadata { get; init; }
}
```

Model for this type defined like this
```c#
            EdmEntityType type = new("My.Project.Namespace",
                                     nameof(MyObject));

            type.AddStructuralProperty(nameof(MyObject.SomeValue), EdmPrimitiveTypeKind.String);

            keyProp = type.AddStructuralProperty(nameof(MyObject.Id), EdmPrimitiveTypeKind.Int32);
            type.AddKeys(keyProp);

            EdmComplexType metadataType =
                new("System.Text.Json", nameof(JsonDocument), null, false, true);

            EdmComplexTypeReference typeRef =
                new(metadataType,
                        isNullable: true);

            type.AddStructuralProperty(nameof(MyObject.Metadata), typeRef);
            EdmEntitySet objects = container.AddEntitySet("objects", type, false);

            model.AddElement(type);
            model.AddElement(metadataType);
```

I defined such FilterBinder implementation that allows to create proper expressions for that (.../odata/objects?$filter=metadata/prop eq 'MyValue' to obj.Metadata.RootObject.GetProperty(prop.Name).GetString()) and implementation for ODataResourceSerializer to write values from JsonDocument to ODataWriter. And now everything works except $select. SelectExpandBinder cannot create proper expressions for JsonDocument and as a result when I do .../odata/objects?$select=metadata I receive [... "value": {"metadata": {}} ...]. To be able to process JsonDocument there I would like to provide my own implementation of SelectExpandBinder that will build proper expressions. 

## Change proposal
1. Make SelectExpandBinder class public and maybe introduce interface for it.
2. Register it in DI system instead and use object from there instead of calling static method (examples in SelectExpandQueryOptions).

KenitoInc:
Fixes #304 

The main changes:
- Make `SelectExpandBinder` public.
- Add a new interface `ISelectExpandBinder` and make `SelectExpandBinder` inherit the interface.
- Add a new class `SelectExpandBinderContext` to wrap all properties needed by the `ISelectExpandBinder`.
- Register `ISelectExpandBinder` to the dependency injection container.

If a customer has their own custom implementation e.g `MySelectExpandBinder`, 
The `MySelectExpandBinder`  class should implement  `ISelectExpandBinder` directly or inherit from `SelectExpandBinder`.

*First option:*
```
public class MySelectExpandBinder : ISelectExpandBinder
{ 
} 
``` 
*Second option:*
```
public class MySelectExpandBinder : SelectExpandBinder
{ 
} 
```
In the first option, the customer will write their own implementation of all methods.
In the second option, the customer can choose the specific method to override and use the default implementation of the SelectExpandBinder for the other methods.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Abstracts/ContainerBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Abstracts/ContainerBuilderExtensions.cs
@@ -100,7 +100,7 @@ namespace Microsoft.AspNetCore.OData.Abstracts
             // Binders.
             builder.AddService<ODataQuerySettings>(ServiceLifetime.Scoped);
             builder.AddService<FilterBinder>(ServiceLifetime.Transient);
-            builder.AddService<ISelectExpandBinder, SelectExpandBinder>(ServiceLifetime.Scoped);
+            builder.AddService<ISelectExpandBinder, SelectExpandBinder>(ServiceLifetime.Singleton);
 
             // HttpRequestScope.
             builder.AddService<HttpRequestScope>(ServiceLifetime.Scoped);
--- a/src/Microsoft.AspNetCore.OData/Abstracts/ContainerBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Abstracts/ContainerBuilderExtensions.cs
@@ -100,6 +100,7 @@ namespace Microsoft.AspNetCore.OData.Abstracts
             // Binders.
             builder.AddService<ODataQuerySettings>(ServiceLifetime.Scoped);
             builder.AddService<FilterBinder>(ServiceLifetime.Transient);
+            builder.AddService<ISelectExpandBinder, SelectExpandBinder>(ServiceLifetime.Scoped);
 
             // HttpRequestScope.
             builder.AddService<HttpRequestScope>(ServiceLifetime.Scoped);
--- a/src/Microsoft.AspNetCore.OData/Query/Expressions/ISelectExpandBinder.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Expressions/ISelectExpandBinder.cs
@@ -1,0 +1,46 @@
+﻿//-----------------------------------------------------------------------------
+// <copyright file="ISelectExpandBinder.cs" company=".NET Foundation">
+//      Copyright (c) .NET Foundation and Contributors. All rights reserved.
+//      See License.txt in the project root for license information.
+// </copyright>
+//------------------------------------------------------------------------------
+
+using System.Linq;
+using System.Linq.Expressions;
+using Microsoft.OData.UriParser;
+
+namespace Microsoft.AspNetCore.OData.Query.Expressions
+{
+    /// <summary>
+    /// Exposes the ability to translate an OData $select or $expand parse tree represented by <see cref="SelectExpandClause"/> to
+    /// an <see cref="Expression"/> and applies it to an <see cref="IQueryable"/> or an <see cref="object"/>.
+    /// </summary>
+    public interface ISelectExpandBinder
+    {
+        /// <summary>
+        /// Translate an OData $select or $expand parse tree represented by <see cref="SelectExpandClause"/> to
+        /// an <see cref="Expression"/> and applies it to an <see cref="IQueryable"/>.
+        /// </summary>
+        /// <param name="source">The original <see cref="IQueryable"/>.</param>
+        /// <param name="context">An instance of the <see cref="SelectExpandBinderContext"/>.</param>
+        /// <returns>The new <see cref="IQueryable"/> after the select/expand query has been applied to.</returns>
+        IQueryable Bind(IQueryable source, SelectExpandBinderContext context);
+
+        /// <summary>
+        /// Translate an OData $select or $expand parse tree represented by <see cref="SelectExpandClause"/> to
+        /// an <see cref="Expression"/> and applies it to an <see cref="object"/>.
+        /// </summary>
+        /// <param name="source">The original <see cref="object"/>.</param>
+        /// <param name="context">An instance of the <see cref="SelectExpandBinderContext"/>.</param>
+        /// <returns>The new <see cref="object"/> after the select/expand query has been applied to.</returns>
+        object Bind(object source, SelectExpandBinderContext context);
+
+        /// <summary>
+        /// Translate an OData $select or $expand parse tree represented by <see cref="SelectExpandClause"/> to
+        /// an <see cref="Expression"/>
+        /// </summary>
+        /// <param name="selectExpandQuery">The <see cref="SelectExpandQueryOption"/> which contains the $select and $expand query options.</param>
+        /// <returns>An <see cref="Expression"/> which can be later applied to an <see cref="IQueryable"/> or an <see cref="object"/>.</returns>
+        Expression GetProjectionLambda(SelectExpandQueryOption selectExpandQuery);
+    }
+}
--- a/src/Microsoft.AspNetCore.OData/Query/Expressions/ISelectExpandBinder.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Expressions/ISelectExpandBinder.cs
@@ -34,13 +34,5 @@ namespace Microsoft.AspNetCore.OData.Query.Expressions
         /// <param name="context">An instance of the <see cref="SelectExpandBinderContext"/>.</param>
         /// <returns>The new <see cref="object"/> after the select/expand query has been applied.</returns>
         object Bind(object source, SelectExpandBinderContext context);
-
-        /// <summary>
-        /// Translate an OData $select or $expand parse tree represented by <see cref="SelectExpandClause"/> to
-        /// an <see cref="Expression"/>
-        /// </summary>
-        /// <param name="selectExpandQuery">The <see cref="SelectExpandQueryOption"/> which contains the $select and $expand query options.</param>
-        /// <returns>A <see cref="LambdaExpression"/> which can be later applied to an <see cref="IQueryable"/> or an <see cref="object"/>.</returns>
-        LambdaExpression GetProjectionLambda(SelectExpandQueryOption selectExpandQuery);
     }
 }
--- a/src/Microsoft.AspNetCore.OData/Query/Expressions/ISelectExpandBinder.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Expressions/ISelectExpandBinder.cs
@@ -23,7 +23,7 @@ namespace Microsoft.AspNetCore.OData.Query.Expressions
         /// </summary>
         /// <param name="source">The original <see cref="IQueryable"/>.</param>
         /// <param name="context">An instance of the <see cref="SelectExpandBinderContext"/>.</param>
-        /// <returns>The new <see cref="IQueryable"/> after the select/expand query has been applied to.</returns>
+        /// <returns>The new <see cref="IQueryable"/> after the select/expand query has been applied.</returns>
         IQueryable Bind(IQueryable source, SelectExpandBinderContext context);
 
         /// <summary>
@@ -32,7 +32,7 @@ namespace Microsoft.AspNetCore.OData.Query.Expressions
         /// </summary>
         /// <param name="source">The original <see cref="object"/>.</param>
         /// <param name="context">An instance of the <see cref="SelectExpandBinderContext"/>.</param>
-        /// <returns>The new <see cref="object"/> after the select/expand query has been applied to.</returns>
+        /// <returns>The new <see cref="object"/> after the select/expand query has been applied.</returns>
         object Bind(object source, SelectExpandBinderContext context);
 
         /// <summary>
--- a/src/Microsoft.AspNetCore.OData/Query/Expressions/ISelectExpandBinder.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Expressions/ISelectExpandBinder.cs
@@ -40,7 +40,7 @@ namespace Microsoft.AspNetCore.OData.Query.Expressions
         /// an <see cref="Expression"/>
         /// </summary>
         /// <param name="selectExpandQuery">The <see cref="SelectExpandQueryOption"/> which contains the $select and $expand query options.</param>
-        /// <returns>An <see cref="Expression"/> which can be later applied to an <see cref="IQueryable"/> or an <see cref="object"/>.</returns>
-        Expression GetProjectionLambda(SelectExpandQueryOption selectExpandQuery);
+        /// <returns>A <see cref="LambdaExpression"/> which can be later applied to an <see cref="IQueryable"/> or an <see cref="object"/>.</returns>
+        LambdaExpression GetProjectionLambda(SelectExpandQueryOption selectExpandQuery);
     }
 }
--- a/src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
@@ -91,7 +91,7 @@ namespace Microsoft.AspNetCore.OData.Query.Expressions
         }
 
         /// <inheritdoc/>
-        public virtual Expression GetProjectionLambda(SelectExpandQueryOption selectExpandQuery)
+        public virtual LambdaExpression GetProjectionLambda(SelectExpandQueryOption selectExpandQuery)
         {
             Contract.Assert(selectExpandQuery != null);
             Type elementType = selectExpandQuery.Context.ElementClrType;
@@ -110,7 +110,7 @@ namespace Microsoft.AspNetCore.OData.Query.Expressions
         private object Bind(object entity, SelectExpandQueryOption selectExpandQuery)
         {
             // Needn't to verify the input, that's done at upper level.
-            LambdaExpression projectionLambda = GetProjectionLambda(selectExpandQuery) as LambdaExpression;
+            LambdaExpression projectionLambda = GetProjectionLambda(selectExpandQuery);
 
             // TODO: cache this ?
             return projectionLambda.Compile().DynamicInvoke(entity);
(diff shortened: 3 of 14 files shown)


src/Microsoft.AspNetCore.OData/Query/Expressions/ISelectExpandBinder.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <returns>The new <see cref="object"/> after the select/expand query has been applied to.</returns>
+        object Bind(object source, SelectExpandBinderContext context);
+
+        /// <summary>
+        /// Translate an OData $select or $expand parse tree represented by <see cref="SelectExpandClause"/> to
+        /// an <see cref="Expression"/>
+        /// </summary>
+        /// <param name="selectExpandQuery">The <see cref="SelectExpandQueryOption"/> which contains the $select and $expand query options.</param>
+        /// <returns>An <see cref="Expression"/> which can be later applied to an <see cref="IQueryable"/> or an <see cref="object"/>.</returns>
+        Expression GetProjectionLambda(SelectExpandQueryOption selectExpandQuery);
```
odero, line 44: It seems existing implementations are returning `LambdaExpression` and the name of the method also implies returning a Lambda. Why not have the return type be a `LambdaExpression`?

src/Microsoft.AspNetCore.OData/Query/Expressions/ISelectExpandBinder.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <returns>The new <see cref="IQueryable"/> after the select/expand query has been applied to.</returns>
+        IQueryable Bind(IQueryable source, SelectExpandBinderContext context);
+
+        /// <summary>
+        /// Translate an OData $select or $expand parse tree represented by <see cref="SelectExpandClause"/> to
+        /// an <see cref="Expression"/> and applies it to an <see cref="object"/>.
+        /// </summary>
+        /// <param name="source">The original <see cref="object"/>.</param>
+        /// <param name="context">An instance of the <see cref="SelectExpandBinderContext"/>.</param>
+        /// <returns>The new <see cref="object"/> after the select/expand query has been applied to.</returns>
```
odero, line 35:
```suggestion
        /// <returns>The new <see cref="object"/> after the select/expand query has been applied.</returns>
```

src/Microsoft.AspNetCore.OData/Query/Expressions/ISelectExpandBinder.cs
KenitoInc: There is not specific reason I am returning `Expression`. I just decided to return the `Expression` since it's easy to cast it to a `LambdaExpression`

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -45,28 +54,63 @@ public SelectExpandBinder(ODataQuerySettings settings, ODataQueryContext context
             _settings = settings;
         }
 
-        public static IQueryable Bind(IQueryable queryable, ODataQuerySettings settings, SelectExpandQueryOption selectExpandQuery)
```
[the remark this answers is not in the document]
KenitoInc: Exactly

src/Microsoft.AspNetCore.OData/Query/Expressions/ISelectExpandBinder.cs
KenitoInc: #Resolved

src/Microsoft.AspNetCore.OData/Abstracts/ContainerBuilderExtensions.cs
```diff
@@ -100,6 +100,7 @@ public static IContainerBuilder AddDefaultWebApiServices(this IContainerBuilder
             // Binders.
             builder.AddService<ODataQuerySettings>(ServiceLifetime.Scoped);
             builder.AddService<FilterBinder>(ServiceLifetime.Transient);
```
FlaviusHouk, line 102: While interface is added to SelectExpandBinder, wouldn't it be consistent to add interface for FilterBinder also?

src/Microsoft.AspNetCore.OData/Query/Expressions/ISelectExpandBinder.cs
FlaviusHouk: But in general it will allow to implementers to return not ```LambdaExpression```. It is against Liskov Substitution Principle, but should be fortified by interface definition? Because now name of the method and return type is slightly confusing.

src/Microsoft.AspNetCore.OData/Abstracts/ContainerBuilderExtensions.cs
KenitoInc: @FlaviusHouk We have a PR related to the FilterBinder #291

FlaviusHouk: @KenitoInc thank you for pointing.

src/Microsoft.AspNetCore.OData/Query/Expressions/ISelectExpandBinder.cs
KenitoInc: #Resolved

odero (approved):
lgtm

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/SelectExpandBinderTest.cs
```diff
@@ -90,7 +90,15 @@ public void Bind_ReturnsIEdmObject_WithRightEdmType(string select)
...
             // Act
-            IQueryable queryable = SelectExpandBinder.Bind(_queryable, _settings, selectExpand);
+            SelectExpandBinderContext selectExpandBinderContext = new SelectExpandBinderContext()
+            {
+                SelectExpandQuery = selectExpand,
+                QuerySettings = _settings,
+                QueryContext = selectExpand.Context
+            };
+
+            SelectExpandBinder binder = new SelectExpandBinder(_settings, selectExpand.Context);
```
habbes, line 100: If these arguments are passed to the `Bind()` method via the `selectExpandBinderContext`, why do you need to also pass them to the constructor?

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/SelectExpandBinderTest.cs
```diff
@@ -90,7 +90,15 @@ public void Bind_ReturnsIEdmObject_WithRightEdmType(string select)
             SelectExpandQueryOption selectExpand = new SelectExpandQueryOption(select: select, expand: null, context: _context);
 
             // Act
-            IQueryable queryable = SelectExpandBinder.Bind(_queryable, _settings, selectExpand);
+            SelectExpandBinderContext selectExpandBinderContext = new SelectExpandBinderContext()
+            {
+                SelectExpandQuery = selectExpand,
+                QuerySettings = _settings,
+                QueryContext = selectExpand.Context
```
habbes, line 97: Is it necessary to set `QueryContext` as separate property if it can be retrieved from `selectExpand.Context`? Are there cases where these two are not the same? Or cases where `selectExpand.Context` is not defined?

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -45,22 +54,57 @@ public SelectExpandBinder(ODataQuerySettings settings, ODataQueryContext context
...
+            Contract.Assert(context != null);
+            Contract.Assert(context.SelectExpandQuery != null);
+            Contract.Assert(context.QueryContext != null);
+            Contract.Assert(context.QueryContext.Model != null);
+            Contract.Assert(context.QuerySettings != null);
+            Contract.Assert(context.QuerySettings.HandleNullPropagation != HandleNullPropagationOption.Default);
+
+            _context = context.QueryContext;
+            _model = _context.Model;
+            _settings = context.QuerySettings;
```
habbes, line 88: I think it's potentially a problem that we modify the private variables, i.e. the state of the object, whenever `Bind()` is called. In the pre-existing implementation, the static method `SelectExpandBinder.Bind()` creates a new instance of `SelectExpandBinder` each time.

However, with this change, we're registering a single instance of `SelectExpandBinder` as a singleton, this means modifying the state is inherently not thread safe.

If we have two concurrent requests and each one calls `Bind()` one can modify the context that was set by the other request before it finished executing, leading to sporadic, hard-to-debug and hard-to-reproduce errors (we've already faced a few of those due to using writeable/stateful objects as singletons that are modified by different requests concurrently, [see this for example](https://github.com/OData/odata.net/issues/2057)).

I think the best way to avoid such issues is to get rid of this private fields altogether, unless they're read-only and only set in the constructor. It's safer to just pass the context object down to every method that needs it.

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -45,22 +54,57 @@ public SelectExpandBinder(ODataQuerySettings settings, ODataQueryContext context
...
         {
-            Contract.Assert(queryable != null);
-            Contract.Assert(selectExpandQuery != null);
+            Contract.Assert(source != null);
+            Contract.Assert(context != null);
+            Contract.Assert(context.SelectExpandQuery != null);
+            Contract.Assert(context.QueryContext != null);
+            Contract.Assert(context.QueryContext.Model != null);
+            Contract.Assert(context.QuerySettings != null);
+            Contract.Assert(context.QuerySettings.HandleNullPropagation != HandleNullPropagationOption.Default);
```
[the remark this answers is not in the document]
habbes: I see it also exists in the old code. Still not sure what it's meant to achieve.

src/Microsoft.AspNetCore.OData/Query/Expressions/ISelectExpandBinder.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <returns>The new <see cref="object"/> after the select/expand query has been applied.</returns>
+        object Bind(object source, SelectExpandBinderContext context);
+
+        /// <summary>
+        /// Translate an OData $select or $expand parse tree represented by <see cref="SelectExpandClause"/> to
+        /// an <see cref="Expression"/>
+        /// </summary>
+        /// <param name="selectExpandQuery">The <see cref="SelectExpandQueryOption"/> which contains the $select and $expand query options.</param>
+        /// <returns>A <see cref="LambdaExpression"/> which can be later applied to an <see cref="IQueryable"/> or an <see cref="object"/>.</returns>
+        LambdaExpression GetProjectionLambda(SelectExpandQueryOption selectExpandQuery);
```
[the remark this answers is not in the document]
habbes: It seems that in the current code `GetProjectionLambda` is private, that means that it's not being called by any other classes. So we can safely assume that not making it public will not cause any issues.

I propose we remove it from the interface. Since the customer is already going to implement the `Bind()` methods, which is what we care about, then we shouldn't care whether they implement `GetProjectLambda` or use a different approach.

But if we want to make it easier for them to re-use the same logic for both `Bind()` methods, and take advantage of lambda expressions in both-cases, then I think that's already taken care of by the fact that you made `GetProjectionLambda` public and virtual in the default implementation. I think it make sense to keep it there, but not on the interface.

habbes (changes requested):

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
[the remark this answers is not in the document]
KenitoInc: The assert ensures those properties are not null.

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/SelectExpandBinderTest.cs
KenitoInc: That constructor is `internal`. It's used for testing purposes only. We can clean that though

src/Microsoft.AspNetCore.OData/Query/Expressions/ISelectExpandBinder.cs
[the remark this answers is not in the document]
KenitoInc: #Resolved

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/SelectExpandBinderTest.cs
KenitoInc: #Resolved

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/SelectExpandBinderTest.cs
KenitoInc: #Resolved

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
KenitoInc: #Resolved

habbes (approved):

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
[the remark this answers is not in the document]
gathogojr: I think we should be using debug asserts here rather than contract asserts

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -27,78 +27,69 @@ namespace Microsoft.AspNetCore.OData.Query.Expressions
...
-        }
-
-        private LambdaExpression GetProjectionLambda(SelectExpandQueryOption selectExpandQuery)
+        /// <summary>
+        /// Translate an OData $select or $expand parse tree represented by <see cref="SelectExpandClause"/> to
+        /// an <see cref="Expression"/>
+        /// </summary>
+        /// <param name="context">The <see cref="SelectExpandBinderContext"/> which is a wrapper for properties used by the <see cref="SelectExpandBinder"/>.</param>
+        /// <returns>A <see cref="LambdaExpression"/> which can be later applied to an <see cref="IQueryable"/> or an <see cref="object"/>.</returns>
+        public virtual LambdaExpression GetProjectionLambda(SelectExpandBinderContext context)
```
gathogojr, line 74: Would there be any harm in making this method protected?

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -627,7 +625,8 @@ private Expression CreateTotalCountExpression(Expression source, bool? countOpti
             }
         }
 
-        private Expression BuildPropertyContainer(Expression source, IEdmStructuredType structuredType,
+        private Expression BuildPropertyContainer(Expression source, SelectExpandBinderContext context,
```
gathogojr, line 628: nit: In other methods where you have made similar changes, you added `SelectExpandBinderContext context` as the first parameter

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -798,18 +806,18 @@ private Expression CreateTotalCountExpression(Expression source, bool? countOpti
...
-            Expression countExpression = CreateTotalCountExpression(propertyValue, pathSelectItem.CountOption);
+            Expression countExpression = CreateTotalCountExpression(context, propertyValue, pathSelectItem.CountOption);
 
             // be noted: the property structured type could be null, because the property maybe not a complex property.
             IEdmStructuredType propertyStructuredType = structuralProperty.Type.ToStructuredType();
             ModelBoundQuerySettings querySettings = null;
             if (propertyStructuredType != null)
             {
-                querySettings = EdmHelpers.GetModelBoundQuerySettings(structuralProperty, propertyStructuredType, _context.Model);
+                querySettings = EdmHelpers.GetModelBoundQuerySettings(structuralProperty, propertyStructuredType, model);
```
gathogojr, line 816: Take note of [this PR](https://github.com/OData/AspNetCoreOData/pull/306) where `GetModelBoundQuerySettings` has been converted into an extension method. If you merge your PR after, you made need to rebase and refactor your code

src/Microsoft.AspNetCore.OData/Query/ODataQueryContextExtensions.cs
```diff
@@ -67,5 +67,26 @@ public static FilterBinder GetFilterBinder(this ODataQueryContext context, OData
...
+            if (context == null)
+            {
+                throw Error.ArgumentNull(nameof(context));
+            }
+
+            ISelectExpandBinder binder = null;
+            if (context.RequestContainer != null)
+            {
+                binder = context.RequestContainer.GetService<ISelectExpandBinder>();
+            }
```
gathogojr, line 87: You could replace this with:
```csharp
ISelectExpandBinder binder = context.RequestContainer?.GetService<ISelectExpandBinder>();
```

gathogojr (approved):
Left some comments otherwise LGTM

src/Microsoft.AspNetCore.OData/Query/ODataQueryContextExtensions.cs
KenitoInc: Done

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
KenitoInc: Done

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
KenitoInc: I have changed it to protected

Issue #297, opened 2021-09-03
ODATA 8.0.1 [AutoExpand] Attribute Not Working


martinicr:
**Note**: I'm still not sure if this attribute is fully supported in the new version of OData (8.0.1).

I have the following classes:

```csharp
public class ClassA
{
  public string Id { get; set; };

  [AutoExpand]
  public ClassB ClazzB { get; set; };
}

public class ClassB
{
  public string SomeProp { get; set; };

  [AutoExpand]
  public ClassC ClazzC { get; set; }
}

public class ClassC
{
  public string Id { get; set; }

  public string AnotherProp { get; set; }
}
```

When querying `ClassA` model, mny expectation is to get something like:

```json
{
    "@odata.context": "https://...$metadata#ClassA/$entity",
    "id": "a6fb7183-e87a-428e-8b62-c159920b2a5e",
    "clazzB": {
        "someProp": "some-prop",
        "clazzC" : {
           "Id": "random-id",
           "anotherProp": "another-prop"
         }
    }
}
```

However, actual response body is:

```json
{
    "@odata.context": "https://...$metadata#ClassA/$entity",
    "id": "a6fb7183-e87a-428e-8b62-c159920b2a5e",
    "clazzB": {
        "someProp": "some-prop"
         }
    }
}
```

I see the `[AutoExpand]` attribute is documented [here](https://docs.microsoft.com/en-us/odata/webapi/autoexpand-attribute), however, I'm not sure if this is supported in version 8. If not, what's the preferred way to enable auto expansion in OData models?

Thanks in advance.

xuzhg:
@martinicr It's a known issue same as https://github.com/OData/WebApi/issues/2552.

Because `ClassB` is defined as a complex type.  Autoexpand navigation property on complex type is not fully supported.

I'd like to enable it in the next version if it's ok.

PR #306, opened 2021-09-16
Enable auto expand navigation property on complex type

xuzhg:
See related issue at: https://github.com/OData/AspNetCoreOData/issues/297

## Main changes

1) Move AutoExpand and AutoSelect into AutoSelectExpandHelper
2) Add Extension method to test AutoSelect and AutoExpand
3) Add the Extension method to retrieve AutoSelect and AutoExpand
4) Add ExpandModelPath and SelectModelPath
5) Add Related E2E test cases.


src/Microsoft.AspNetCore.OData/Edm/AutoSelectExpandHelper.cs
```diff
@@ -0,0 +1,354 @@
...
+    internal static class AutoSelectExpandHelper
+    {
+        #region Auto Select and Expand Test
+        /// <summary>
+        /// Tests whether there are auto select properties.
+        /// So far, we only test one depth for auto select, shall we go through the deeper depth?
+        /// </summary>
+        /// <param name="edmModel">The Edm model.</param>
+        /// <param name="structuredType">The type from value or from path.</param>
+        /// <param name="property">The property form path, it can be null.</param>
```
gathogojr, line 28:
```suggestion
        /// <param name="property">The property from path, it can be null.</param>
```

src/Microsoft.AspNetCore.OData/Edm/AutoSelectExpandHelper.cs
```diff
@@ -0,0 +1,354 @@
...
+            return autoSelectProperties;
+        }
+
+        /// <summary>
+        /// Gets the auto expand paths.
+        /// </summary>
+        /// <param name="edmModel">The Edm model.</param>
+        /// <param name="structuredType">The Edm structured type.</param>
+        /// <param name="property">The property starting from, it can be null.</param>
+        /// <param name="isSelectPresent">Is dollar select presented.</param>
```
gathogojr, line 202:
```suggestion
        /// <param name="isSelectPresent">Is dollar select present.</param>
```

src/Microsoft.AspNetCore.OData/Edm/AutoSelectExpandHelper.cs
```diff
@@ -0,0 +1,354 @@
...
+    {
+        #region Auto Select and Expand Test
+        /// <summary>
+        /// Tests whether there are auto select properties.
+        /// So far, we only test one depth for auto select, shall we go through the deeper depth?
+        /// </summary>
+        /// <param name="edmModel">The Edm model.</param>
+        /// <param name="structuredType">The type from value or from path.</param>
+        /// <param name="property">The property form path, it can be null.</param>
+        /// <returns>true/false.</returns>
```
gathogojr, line 29:
```suggestion
        /// <returns>true if the structured type has auto select properties; otherwise false.</returns>
```

src/Microsoft.AspNetCore.OData/Edm/AutoSelectExpandHelper.cs
```diff
@@ -0,0 +1,354 @@
...
+            return false;
+        }
+
+        /// <summary>
+        /// Tests whether there are auto expand properties.
+        /// </summary>
+        /// <param name="edmModel">The Edm model.</param>
+        /// <param name="structuredType">The Edm structured type.</param>
+        /// <param name="property">The property form path, it can be null.</param>
+        /// <returns>true/false.</returns>
```
gathogojr, line 70:
```suggestion
        /// <returns>true if the structured type has auto expand properties; otherwise false.</returns>
```

src/Microsoft.AspNetCore.OData/Edm/AutoSelectExpandHelper.cs
```diff
@@ -0,0 +1,354 @@
...
+
+            return false;
+        }
+
+        /// <summary>
+        /// Tests whether there are auto expand properties.
+        /// </summary>
+        /// <param name="edmModel">The Edm model.</param>
+        /// <param name="structuredType">The Edm structured type.</param>
+        /// <param name="property">The property form path, it can be null.</param>
```
gathogojr, line 69:
```suggestion
        /// <param name="property">The property from path, it can be null.</param>
```

src/Microsoft.AspNetCore.OData/Edm/AutoSelectExpandHelper.cs
```diff
@@ -0,0 +1,354 @@
...
+                foreach (IEdmProperty property in properties)
+                {
+                    switch (property.PropertyKind)
+                    {
+                        case EdmPropertyKind.Structural:
+                            IEdmStructuralProperty structuralProperty = (IEdmStructuralProperty)property;
+                            IEdmTypeReference typeRef = property.Type.GetElementTypeOrSelf();
+                            if (typeRef.IsComplex() && edmModel.CanExpand(typeRef.AsComplex().ComplexDefinition(), structuralProperty))
+                            {
+                                IEdmStructuredType subStrucutredType = typeRef.AsStructured().StructuredDefinition();
```
gathogojr, line 113:
```suggestion
                                IEdmStructuredType subStructuredType = typeRef.AsStructured().StructuredDefinition();
```

src/Microsoft.AspNetCore.OData/Edm/AutoSelectExpandHelper.cs
```diff
@@ -0,0 +1,354 @@
...
+                                    return true;
+                                }
+                            }
+                            break;
+
+                        case EdmPropertyKind.Navigation:
+                            IEdmNavigationProperty navigationProperty = (IEdmNavigationProperty)property;
+                            if (IsAutoExpand(navigationProperty, pathProperty, edmStructuredType, edmModel))
+                            {
+                                return true; // find an auto-expand navigation property path
```
gathogojr, line 125:
```suggestion
                                return true; // found an auto-expand navigation property path
```

src/Microsoft.AspNetCore.OData/Edm/AutoSelectExpandHelper.cs
```diff
@@ -0,0 +1,354 @@
...
+        #endregion
+
+        /// <summary>
+        /// Gets the auto select paths.
+        /// </summary>
+        /// <param name="edmModel">The Edm model.</param>
+        /// <param name="structuredType">The Edm structured type.</param>
+        /// <param name="pathProperty">The property form path, it can be null.</param>
+        /// <param name="querySettings">The query settings.</param>
+        /// <returns>The auto expand paths.</returns>
```
gathogojr, line 143:
```suggestion
        /// <returns>The auto select paths.</returns>
```

src/Microsoft.AspNetCore.OData/Edm/AutoSelectExpandHelper.cs
```diff
@@ -0,0 +1,354 @@
...
+            return false;
+        }
+        #endregion
+
+        /// <summary>
+        /// Gets the auto select paths.
+        /// </summary>
+        /// <param name="edmModel">The Edm model.</param>
+        /// <param name="structuredType">The Edm structured type.</param>
+        /// <param name="pathProperty">The property form path, it can be null.</param>
```
gathogojr, line 141:
```suggestion
        /// <param name="pathProperty">The property from path, it can be null.</param>
```

src/Microsoft.AspNetCore.OData/Edm/AutoSelectExpandHelper.cs
```diff
@@ -0,0 +1,354 @@
...
+            List<SelectModelPath> autoSelectProperties = new List<SelectModelPath>();
+
+            List<IEdmStructuredType> structuredTypes = new List<IEdmStructuredType>();
+            structuredTypes.Add(structuredType);
+            structuredTypes.AddRange(edmModel.FindAllDerivedTypes(structuredType));
+
+            foreach (IEdmStructuredType edmStructuredType in structuredTypes)
+            {
+                IEnumerable<IEdmStructuralProperty> properties;
+                if (edmStructuredType == structuredType)
```
gathogojr, line 166: In `HasAutoSelectProperty` and `HasAutoExpandProperty` methods, you used the ternary expression instead. When considered together with the helpful comments you added here, I like the if...else better. Or maybe you could refactor this into a method that the various method can all call.

src/Microsoft.AspNetCore.OData/Edm/ExpandModelPath.cs
```diff
@@ -0,0 +1,105 @@
...
+                {
+                    if (index < count - 1 || foundNavProp)
+                    {
+                        throw new ODataException(Error.Format(SRResources.InvalidLastSegmentInSelectExpandPath, element.GetType().Name));
+                    }
+
+                    foundNavProp = true;
+                    Navigation = navigationProperty;
+
+                    // identifiers.Add(navigationProperty.Name); // don't add the navigation property into
```
gathogojr, line 87: Explain what is happening here...

src/Microsoft.AspNetCore.OData/Edm/EdmHelpers.cs
```diff
@@ -528,6 +301,145 @@ public static bool IsNotExpandable(IEdmProperty edmProperty, IEdmModel edmModel)
...
+                    mergedQuerySettings.CopyExpandConfigurations(propertyTypeQuerySettings.ExpandConfigurations);
+                    mergedQuerySettings.DefaultExpandType = propertyTypeQuerySettings.DefaultExpandType;
+                    mergedQuerySettings.DefaultMaxDepth = propertyTypeQuerySettings.DefaultMaxDepth;
+                }
+            }
+
+            return mergedQuerySettings;
+        }
+
+        internal static QueryableRestrictionsAnnotation GetPropertyRestrictions(IEdmProperty edmProperty, IEdmModel edmModel)
```
gathogojr, line 435: Would it be better to make this an extension method for `IEdmModel` type?

src/Microsoft.AspNetCore.OData/Edm/EdmHelpers.cs
```diff
@@ -528,6 +301,145 @@ public static bool IsNotExpandable(IEdmProperty edmProperty, IEdmModel edmModel)
...
+        {
+            ModelBoundQuerySettings mergedQuerySettings = new ModelBoundQuerySettings(propertyQuerySettings);
+            if (propertyTypeQuerySettings != null)
+            {
+                if (!mergedQuerySettings.PageSize.HasValue)
+                {
+                    mergedQuerySettings.PageSize = propertyTypeQuerySettings.PageSize;
+                }
+
+                if (mergedQuerySettings.MaxTop == 0 && propertyTypeQuerySettings.MaxTop != 0)
```
gathogojr, line 396: If I recall, `MaxTop` is nullable. If the value is `null`, then `propertyTypeQuerySettings.MaxTop != 0` evaluates to `true`. Thought I should confirm you're aware of that

src/Microsoft.AspNetCore.OData/Edm/AutoSelectExpandHelper.cs
xuzhg: It looks no difference, What's your suggestion?

src/Microsoft.AspNetCore.OData/Edm/AutoSelectExpandHelper.cs
xuzhg: Ok

src/Microsoft.AspNetCore.OData/Edm/AutoSelectExpandHelper.cs
xuzhg: Oh, I see.

src/Microsoft.AspNetCore.OData/Edm/ExpandModelPath.cs
xuzhg: identifiers is a list to templory save the path identifiers, It is used to construct the navigation path.

The navigation path (without the last navigation property name identifier) is used to retrieve the navigation binding.

I added more comments into the source.

src/Microsoft.AspNetCore.OData/Edm/EdmHelpers.cs
xuzhg: I'd like to leave it to another PR avoiding noise in one PR.

src/Microsoft.AspNetCore.OData/Edm/AutoSelectExpandHelper.cs
xuzhg: haha~ I wrote the method "GetAutoExpandPaths" first and at that method, i have to push the "derived type" into Stack< >.

When I wrote the method "GetAutoSelectPaths", i copied and pasted most of codes from "GetAutoExpandPaths". 
And I realized the Stack<  > is not needed, so I removed the push() method call but keep the If... else clause.

I refactor it to ternary expression. (I like that pattern. :). And I added the comments.

src/Microsoft.AspNetCore.OData/Edm/EdmHelpers.cs
xuzhg: the codes are existed codes, I just moved the place.
I'd like to keep the codes unchanged. Do you think it's a bug? If yes, we should fix it in another PR.

xuzhg:
@gathogojr @KenitoInc @mikepizzo any comments?

src/Microsoft.AspNetCore.OData/Edm/AutoSelectExpandHelper.cs
gathogojr: Fix spelling `subStrucutredType` => `subStructuredType`

src/Microsoft.AspNetCore.OData/Edm/SelectModelPath.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// <summary>
+        /// Gets the select path.
+        /// </summary>
+        public string SelectPath => _selectPath;
+
+        private void ValidateAndCalculateElementPath()
+        {
+            IList<string> identifiers = new List<string>();
+            foreach (IEdmElement element in this)
+            {
```
gathogojr, line 41: In select path, doesn't the order of property and type cast matter? Asking since we don't appear to be validating that here

gathogojr (approved):
Left a few comments; otherwise LGTM

src/Microsoft.AspNetCore.OData/Edm/SelectModelPath.cs
KenitoInc: @gathogojr Order in `select` doesn't matter... and I don't think we have type cast in select

KenitoInc (approved):
Approved with a few comments

src/Microsoft.AspNetCore.OData/Edm/SelectModelPath.cs
gathogojr: @KenitoInc My question was based on the class documentation
```csharp
    /// The select property path is a model path with the following restriction:
    /// 1) only include type case and structural property
```

PR #312, opened 2021-09-21
Enable nested resource case insensitive when readuntypedasvalue

xuzhg:
Since ODL now supports to read the property with different cases as "Edm.Untyped" in 7.9.2, this PR is to enable "Edm.Untyped" nested resource deserialization.

1) Enable the "Edm.Untyped" in resource deserialization
2) Add the test cases

### DIFF
--- a/test/Microsoft.AspNetCore.OData.E2E.Tests/EdmUntyped/EdmUntypedControllers.cs
+++ b/test/Microsoft.AspNetCore.OData.E2E.Tests/EdmUntyped/EdmUntypedControllers.cs
@@ -1,0 +1,78 @@
+//-----------------------------------------------------------------------------
+// <copyright file="EdmUntypedControllers.cs" company=".NET Foundation">
+//      Copyright (c) .NET Foundation and Contributors. All rights reserved.
+//      See License.txt in the project root for license information.
+// </copyright>
+//------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.OData.Deltas;
+using Microsoft.AspNetCore.OData.Query;
+using Microsoft.AspNetCore.OData.Routing.Controllers;
+using Xunit;
+
+namespace Microsoft.AspNetCore.OData.E2E.Tests.EdmUntyped
+{
+    public class BillsController : ODataController
+    {
+        [EnableQuery]
+        public IActionResult Post([FromBody]Bill bill)
+        {
+            Assert.NotNull(bill);
+
+            // Let's verify the 'bill' from the "Post" test case
+            Assert.Equal(921, bill.ID);
+            Assert.Equal("Fan", bill.Name);
+            Assert.Equal(Frequency.BiWeekly, bill.Frequency);
+            Assert.Equal(3.14, bill.Weight);
+
+            Assert.NotNull(bill.HomeAddress);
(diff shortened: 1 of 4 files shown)


test/Microsoft.AspNetCore.OData.E2E.Tests/EdmUntyped/EdmUntypedControllers.cs
```diff
@@ -0,0 +1,78 @@
...
+            return Ok(true);
+        }
+
+        [EnableQuery]
+        public IActionResult Patch(int key, Delta<Bill> delta)
+        {
+            Assert.Equal(2, key);
+
+            // Let's verify the 'delta' from the "Patch" test case
+            delta.TryGetPropertyValue("Frequency", out object frequency);
```
gathogojr, line 52: As an enhancement, you could change the `TryGetPropertyValue` calls to Assert.True(delta.TryGetPropertyValue("Frequency", out object frequency));

gathogojr:
@xuzhg In 8.x, is case-insensitive request property binding support by default and is that option configurable? I remember in 7.x there was a compatibility option [`DisableCaseInsensitiveRequestPropertyBinding`](https://github.com/OData/WebApi/blob/0338075c9940fe29c6fd18cd3f9e0433fd82d745/src/Microsoft.AspNet.OData.Shared/CompatibilityOptions.cs#L31) @KenitoInc introduced. If my understanding is correct, in 8.x we no longer offer the option to enforce case-sensitive request property binding, no?

xuzhg:
> @xuzhg In 8.x, is case-insensitive request property binding support by default and is that option configurable? I remember in 7.x there was a compatibility option [`DisableCaseInsensitiveRequestPropertyBinding`](https://github.com/OData/WebApi/blob/0338075c9940fe29c6fd18cd3f9e0433fd82d745/src/Microsoft.AspNet.OData.Shared/CompatibilityOptions.cs#L31) @KenitoInc introduced. If my understanding is correct, in 8.x we no longer offer the option to enforce case-sensitive request property binding, no?

@gathogojr Thanks for pointing it out.  In 8.x, we don't need such a configuration. We need to support the case-insensitive by default.

KenitoInc:
> > @xuzhg In 8.x, is case-insensitive request property binding support by default and is that option configurable? I remember in 7.x there was a compatibility option [`DisableCaseInsensitiveRequestPropertyBinding`](https://github.com/OData/WebApi/blob/0338075c9940fe29c6fd18cd3f9e0433fd82d745/src/Microsoft.AspNet.OData.Shared/CompatibilityOptions.cs#L31) @KenitoInc introduced. If my understanding is correct, in 8.x we no longer offer the option to enforce case-sensitive request property binding, no?
> 
> @gathogojr Thanks for pointing it out. In 8.x, we don't need such a configuration. We need to support the case-insensitive by default.

I had created a workitem for tracking this https://github.com/OData/WebApi/issues/2364
The compatibility flag was added so as not to breaking existing customers

test/Microsoft.AspNetCore.OData.E2E.Tests/EdmUntyped/EdmUntypedControllers.cs
xuzhg: #Resolved.

xuzhg:
@gathogojr @KenitoInc any other comments?

gathogojr (approved):
:shipit:

KenitoInc (approved):

Issue #85, opened 2021-02-06
DateTime Created with too Much Decimal Precision Causing SQL Query to Fail


wjvii:
I posted this on the ODataConnectedService repo as https://github.com/OData/ODataConnectedService/issues/189 but I am adding it here because maybe it can and should be handled on the server side. The problem is that the ODataConnectedService client creates a DateTime with too much precision and this causes a Microsoft.Data.SqlClient.SqlException (0x80131904): Conversion failed when converting date and/or time from character string.

```
var dt = DateTime.Now.AddDays(-30);
var query = (DataServiceQuery)context.Customer.Where(c => c.CreationDate >= dt);

The query uri is: https://localhost:5001/odata/Customer?$filter=CreationDate ge 2020-11-25T19:19:18.0172084-06:00

It generates the following SQL
WHERE [p].[CreationDate] >= '2020-11-25T19:19:18.0172084'
```

If I test that query in SSMS it fails unless I reduce the precision to 3 decimal points.

I even tried making sure the decimal time would be zero

WHERE [p].[CreationDate] >= '2021-02-01T00:00:00.0000000'

This also fails.

Seems like at least with SQL Server it needs to truncate the precision to 3 decimal places.

xuzhg:
`var dt = DateTime.Now.AddDays(-30);` uses the local time zone info to create the datetime?
Can you use the 'DataTimeOffset' and set the precision?

henning-krause:
@xuzhg I've created a PR for this issue. Could you take a look?

KenitoInc:
This was fixed by #323

PR #323, opened 2021-10-01
Fixed issue #85: DateTime Created with too Much Decimal Precision Causing SQL Query to Fail

henning-krause:
This will fix the issue #85 by using property parameterization if possible for DateTimeOffset conversions.


xuzhg (approved):
:shipit:

PR #324, opened 2021-10-03
Async query enumeration and cancellation support for long running queries

henning-krause:
A long time ago I created an [issue](https://github.com/OData/WebApi/issues/1788) to properly cancel long running queries (when the HttpContext.RequestAborted cancellation token is triggered) for the 7.9 version of this code. This was never solved, I think mainly because most of the code is still synchronous.

Now that I moved on to AspNetCoreOData I'm facing the same problem and gave it another shot.

@xuzhg I know that this PR is far from complete. But I wanted to get an opinion from you if I'm moving in the right direction with this and if you would approve this PR once it's complete before I put more effort in this.

For now, I have mainly redesigned the TruncatedCollection to properly support async enumeration (and remove the buffering of the whole response along with this change) as well as the ODataResourceSetSerializer.

So, for paged queries, this approach works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,9 @@ This is the official ASP.NET Core OData repository.
 
 * [ASP.NET Core OData 8.0 Preview for .NET 5](https://devblogs.microsoft.com/odata/asp-net-odata-8-0-preview-for-net-5/)
 
+* [API versioning extension with ASP.NET Core OData 8](https://devblogs.microsoft.com/odata/api-versioning-extension-with-asp-net-core-odata-8/)
+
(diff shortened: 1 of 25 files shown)


henning-krause:
@corranrogue9 I've resolved all the comments you had.

corranrogue9:
> @corranrogue9 I've resolved all the comments you had.

Thanks, I've approved

henning-krause:
@xuzhg All unit tests are green and @corranrogue9 approved the changes.

henning-krause:
@xuzhg Is there anything I need to do to get this PR into a state where it can be merged?

Issue #345, opened 2021-10-26
DateTimeOffset deserialization doesn't seem to be supported.


ayo10:
If you try and create an entity that has a DateTimeOffset property, it causes a bad request to occur. The entity is null, and the model state will have an error in it with that property as the offender.

Specifically, the call from ODataResourceDeserializer.ApplyStructuralProperty makes it all the way to EdmPrimitiveHelper.ConvertPrimitiveValue, which doesn't handle DateTimeOffset types, which in turn causes it to call Convert.ChangeType which itself doesn't natively support converting strings to DateTimeOffset. Breakage ensues and a ValidationException is thrown.

The callstack:
>  at Microsoft.AspNetCore.OData.Edm.EdmPrimitiveHelper.ConvertPrimitiveValue(Object value, Type type, TimeZoneInfo timeZoneInfo) in D:\a\1\s\src\Microsoft.AspNetCore.OData\Edm\EdmPrimitiveHelper.cs:line 150
   at Microsoft.AspNetCore.OData.Formatter.Deserialization.DeserializationHelpers.SetDeclaredProperty(Object resource, EdmTypeKind propertyKind, String propertyName, Object propertyValue, IEdmProperty edmProperty, ODataDeserializerContext readContext) in D:\a\1\s\src\Microsoft.AspNetCore.OData\Formatter\Deserialization\DeserializationHelper.cs:line 97
   at Microsoft.AspNetCore.OData.Formatter.Deserialization.DeserializationHelpers.ApplyProperty(ODataProperty property, IEdmStructuredTypeReference resourceType, Object resource, IODataDeserializerProvider deserializerProvider, ODataDeserializerContext readContext) in D:\a\1\s\src\Microsoft.AspNetCore.OData\Formatter\Deserialization\DeserializationHelper.cs:line 65
   at Microsoft.AspNetCore.OData.Formatter.Deserialization.ODataResourceDeserializer.ApplyStructuralProperty(Object resource, ODataProperty structuralProperty, IEdmStructuredTypeReference structuredType, ODataDeserializerContext readContext) in D:\a\1\s\src\Microsoft.AspNetCore.OData\Formatter\Deserialization\ODataResourceDeserializer.cs:line 519
[...]

PR #346, opened 2021-10-28
Fix convert type issue when property name case-insensitive

xuzhg:
#345 

 Since we support property name case-insensitive and enable "ReadUntypedAsString=true" by default.
 ODL reads "unknown" property value as "value", not as "ODataUntypedValue".
 For most of the primitive types, for example, DateTimeOffset, the "value" from ODL is a "string".
 We should convert the "string" value to its correct format.
Ideally, ODL should read "case-insensitive" property value as its correct format. So far, ODL hasn't been enabled.

This PR is to add a workaround for fixing most primitive-type property case-insensitive.

The ideal fix is to fix it on the ODL side and allow to read property name case-insensitive.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Formatter/Deserialization/DeserializationHelper.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Deserialization/DeserializationHelper.cs
@@ -338,7 +338,7 @@ namespace Microsoft.AspNetCore.OData.Formatter.Deserialization
                     return date;
                 }
 
-                throw new ValidationException(Error.Format(SRResources.PropertyCannotBeConverted, clrType.FullName));
+                throw new ValidationException(Error.Format(SRResources.PropertyCannotBeConverted, stringValue, clrType.FullName));
             }
 
             // TimeOfDay
@@ -349,7 +349,7 @@ namespace Microsoft.AspNetCore.OData.Formatter.Deserialization
                     return date;
                 }
 
-                throw new ValidationException(Error.Format(SRResources.PropertyCannotBeConverted, clrType.FullName));
+                throw new ValidationException(Error.Format(SRResources.PropertyCannotBeConverted, stringValue, clrType.FullName));
             }
 
             // DateTimeOffset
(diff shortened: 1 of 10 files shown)


src/Microsoft.AspNetCore.OData/Formatter/Deserialization/DeserializationHelper.cs
```diff
@@ -279,6 +279,111 @@ internal static void SetProperty(object resource, string propertyName, object va
...
+
+            // Date
+            if (clrType == typeof(Date))
+            {
+                if (Date.TryParse(stringValue, CultureInfo.InvariantCulture, out Date date))
+                {
+                    return date;
+                }
+
+                throw new ValidationException(Error.Format(SRResources.PropertyCannotBeConverted, clrType.FullName));
```
corranrogue9, line 341: How far up does this exception bubble? Would it be helpful to any caller to know the value of `stringValue` in the exception message?

test/Microsoft.AspNetCore.OData.E2E.Tests/DateTimeOffsetSupport/DateTimeOffsetController.cs
```diff
@@ -20,8 +20,6 @@ public class FilesController : ODataController
     {
         private readonly FilesContext _db;
 
-
```
corranrogue9, line 23: It seems like the test code didn't make it into your commit for some reason...

xuzhg: No, here's just the cleanup for unnecessary white lines.

xuzhg:
@corranrogue9 Do you have any other comments?

corranrogue9:
I would like to see a test of some sort

corranrogue9:
Sorry, accidentally closed the PR...

xuzhg:
When updated to ODL 7.9.4, it's enabled and fixed in ODL side.  See at: 1b26be5f9c05ac0f17ef229f2a9578f9c9b0dc2a
Let me close it

PR #354, opened 2021-11-01
Permit Count, to prevent exception when using ..../$count. 

mperdeck:
Also set max top, so $top can be used.


mperdeck:
This is to fix the ODataNewtonsoftJsonSample sample project, so when $count is used (as shown in the $odata page), it no longer throws an exception.

xuzhg (approved):

Issue #351, opened 2021-10-30
OData, Net5: How to implement $count when controllers return entities of multiple types


mperdeck:
I am adding OData to an existing large API in .Net5, where the action methods in each controller return different types (such as OrganizationFull, OrganizationList, etc.)

This is working, except that the responses do not contain OData meta data, such as $count. So when a consumer calls /api/OrganizationList?$count=true, it doesn't get the count.

I understand that to get this meta data, I have to add an EDM model, like so:

````
// Startup.cs

        public void ConfigureServices(IServiceCollection services)
        {
            services.AddControllers()
                .AddOData(opt => opt.Select().Filter().SetMaxTop(10).AddRouteComponents("odata", GetEdmModel()))

        IEdmModel GetEdmModel()
        {
            var odataBuilder = new ODataConventionModelBuilder();

            odataBuilder.EntitySet<OrganizationFull>(nameof(Organization));
            odataBuilder.EntitySet<OrganizationList>(nameof(Organization)); // throws
            return odataBuilder.GetEdmModel();
        }
````

Here I specified 2 EntitySets, one for each entity type returned from the OrganizationController. 

However, this throws an exception "The entity set 'Organization' was already configured with a different EntityType ('OrganizationFull').". And if I remove the EntitySet for OrganizationList, when I hit the List end point, I get a 404.

How can I get my end points to return the $count meta data, while keeping my API (which returns multiple entity types from one controller)?

**Comment by gathogojr** [comment]
@mperdeck On the `AddOData` method call, enable support for `Count` as follows:
```csharp
                .AddOData(opt => opt.Select().Filter().Count().SetMaxTop(10).AddRouteComponents("odata", GetEdmModel()))
```

mperdeck:
You're right, I forgot to include that. However, this fix doesn't solve my problem - I need to retrieve the count somehow, but I understand the only way to do that is to provide an EDM model. And to provide an EDM model, I understand that each controller can return only one model. To me (and I guess many others), this would mean breaking changes in my API. 

If I'm getting this wrong, please let me know.

I created a pull request with a possible solution at https://github.com/OData/AspNetCoreOData/pull/355

mperdeck:
Found a solution, which however if very dependent on the current version of this library. Added this to the same issue that I raised on Stackoverflow:
https://stackoverflow.com/questions/69776758/odata-net5-how-to-implement-count-when-controllers-return-entities-of-multipl/69879343#69879343

PR #355, opened 2021-11-02
If $count=true, add count to context item

mperdeck:
I added OData to an existing large API, It works great. However, when a client adds $count=true to the url, they do not get the count. To make this happen, I would have to provide an EDM and go via the odata controller. However:

- I would also have to change many of my end points, meaning breaking changes for my clients.
- The responses from the end points would also change (they would now include the odata meta data, and the data would sit in a separate array).

Created this pull request to allow me (and your other users) to retrieve the count without making these changes. If the url contains $count=true, and if the count is available, the count will be added to the HttpContext, as item "ODataCount".

This way, the count can be retrieved in middleware (from the context item). I used this in my API to create a response header with the count. Other users of this library could process the count in another way, or not at all. Specifically, this change does not in itself modify the HTTP response from APIs using this library, so no surprises for your users.  

Thanks in advance for considering this pull request.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/EnableQueryAttribute.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/EnableQueryAttribute.cs
@@ -484,6 +484,19 @@ namespace Microsoft.AspNetCore.OData.Query
                     }
                 }
 
+                // if $count=true is used, add the count to the context. This allows the consuming web site
+                // to retrieve the count in middleware without having to use the odata controller.
+
+                bool countParameterIsTrue = queryOptions.Count?.Value ?? false;
+                if (countParameterIsTrue)
+                {
+                    long? count = request.ODataFeature().TotalCount;
+                    if (count.HasValue)
+                    {
+                        request.HttpContext.Items["ODataCount"] = count;
+                    }
+                }
+
                 return queryable;
             }
         }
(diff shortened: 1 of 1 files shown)


corranrogue9:
Hello @mperdeck, have you discussed this change with anyone from the OData team up to this point?

mperdeck:
No, I haven't. I was under time pressure from my client to somehow serve up the count, without restructuring my API. 

Previously I created the issue that this PR is a response to, at https://github.com/OData/AspNetCoreOData/issues/351

xuzhg:
@mperdeck I think you can do the similar thing at your side by deriving from EnableQueryAttribute and override the virtual method?

mperdeck:
Thanks for your reply. Following your suggestion, I found a solution. Added this to the Stackoverflow counterpart of this issue:
https://stackoverflow.com/questions/69776758/odata-net5-how-to-implement-count-when-controllers-return-entities-of-multipl/69879343#69879343

I'm not sure my solution will still work with future versions of this package. Also, I suspect that I am not the only one who wants to get the count without dealing with an EDM model or changing the contract of their API. Would like to suggest making something like this solution part of the official code base.

corranrogue9:
@mperdeck I'm glad that you found a different avenue to resolve this on your end. Have you previously files an issue for this? Doing so will help us track and prioritize for future versions

mperdeck:
Yes I did, at https://github.com/OData/AspNetCoreOData/issues/351

PR #358, opened 2021-11-04
Add more virtual methods to extend.

FlaviusHouk:
Further extend of SelectExpandBinder class to make it more flexible and allow user to override parts of it's behavior. Related to issue #304.


corranrogue9:
I would like to see how these new instance methods are getting used before adding them to the public API. Maybe it would make sense to have a feature branch for this since resolving the issue seems like it will take more than one PR?

FlaviusHouk:
@corranrogue9 I can add an example of usage for these methods in some of the samples that present in the repository.

KenitoInc:
@FlaviusHouk 
Related changes merged in this commit https://github.com/OData/AspNetCoreOData/commit/65186071cdffb93b77207192eb6e1ed073022c77

PR #362, opened 2021-11-05
Unmapped properties are in Delta<T>

### LINKED ISSUE #347: Unmapped properties are in Delta<T>
_Same issue as WebApi [2532](https://github.com/OData/WebApi/issues/2532)_

In my OData convention based EDM model I have properties that I exclude from the model by using the `[NotMapped]` attribute.

In an entity that has a property with `[NotMapped]`, when I call the Patch method I see that the property is in the `_allProperties` and is accessible (although it is always empty).

Should be ported from WebAPI.

dxrdxr:
Same issue as WebApi [2532](https://github.com/OData/WebApi/issues/2532)

Fixes #347

Ported from WebAPI.

### Notes

- The two new test classes are split into separate files due because the monolithic TestDataModels doesn't exist in this repro

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Deltas/DeltaOfT.cs
+++ b/src/Microsoft.AspNetCore.OData/Deltas/DeltaOfT.cs
@@ -8,11 +8,13 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations.Schema;
 using System.Diagnostics;
 using System.Diagnostics.Contracts;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
+using System.Runtime.Serialization;
 using Microsoft.AspNetCore.OData.Abstracts;
 using Microsoft.AspNetCore.OData.Common;
 
@@ -495,7 +497,7 @@ namespace Microsoft.AspNetCore.OData.Deltas
                 _structuredType,
                 (backingType) => backingType
                     .GetProperties(BindingFlags.Instance | BindingFlags.Public)
-                    .Where(p => (p.GetSetMethod() != null || TypeHelper.IsCollection(p.PropertyType)) && p.GetGetMethod() != null)
+                    .Where(p => !IsIgnoredProperty(backingType.GetCustomAttributes(typeof(DataContractAttribute), inherit: true).Any(), p) && (p.GetSetMethod() != null || TypeHelper.IsCollection(p.PropertyType)) && p.GetGetMethod() != null)
                     .Select<PropertyInfo, PropertyAccessor<T>>(p => new FastPropertyAccessor<T>(p))
                     .ToDictionary(p => p.Property.Name));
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.OData/Deltas/DeltaOfT.cs
```diff
@@ -495,7 +497,7 @@ private void InitializeProperties(IEnumerable<string> updatableProperties)
                 _structuredType,
                 (backingType) => backingType
                     .GetProperties(BindingFlags.Instance | BindingFlags.Public)
-                    .Where(p => (p.GetSetMethod() != null || TypeHelper.IsCollection(p.PropertyType)) && p.GetGetMethod() != null)
+                    .Where(p => !IsIgnoredProperty(backingType.GetCustomAttributes(typeof(DataContractAttribute), inherit: true).Any(), p) && (p.GetSetMethod() != null || TypeHelper.IsCollection(p.PropertyType)) && p.GetGetMethod() != null)
```
corranrogue9, line 500: Won't this boolean get re-evaluated for every property in the backing type? Maybe the JITer does something smart here, but it seems like we should just avoid this in the first place

corranrogue9: Sorry, I'm referring to the `Any` call, not sure how to make github have my comment refer to a subsection of a line...

corranrogue9, line 500: I'm not sure about this, but I think `IsIgnoredProperty` is probably the most expensive part of the predicate, and is the least likely to get hit. Maybe it should be the last part of the predicate so that short circuit evaluation prevents computing it unnecessarily?

src/Microsoft.AspNetCore.OData/Deltas/DeltaOfT.cs
```diff
@@ -514,6 +516,28 @@ private void InitializeProperties(IEnumerable<string> updatableProperties)
...
+            //3. Its marked with IgnoreDataMember (but not where types datacontract and property marked with datamember)
+
+            bool hasNotMappedAttr = propertyInfo.GetCustomAttributes(typeof(NotMappedAttribute), inherit: true).Any();
+
+            if (hasNotMappedAttr)
+            {
+                return true;
+            }
+
+            if (isTypeDataContract)
```
corranrogue9, line 533: Oh, sometimes needing this value isn't even required? The `Where` clause in the caller should be reconsidered given this information. We really shouldn't require evaluating the data contract attribute of the backing type every time if it won't be required when the not mapped attribute is used.

src/Microsoft.AspNetCore.OData/Deltas/DeltaOfT.cs
```diff
@@ -514,6 +516,28 @@ private void InitializeProperties(IEnumerable<string> updatableProperties)
...
+            {
+                return !propertyInfo.GetCustomAttributes(typeof(DataMemberAttribute), inherit: true).Any();
+            }
+
+            return propertyInfo.GetCustomAttributes(typeof(IgnoreDataMemberAttribute), inherit: true).Any();
```
corranrogue9, line 538: Minor performance improvement: is there a way to combine these `Any` calls into one? There's going to be a lot of reflection if we fall through to this line

src/Microsoft.AspNetCore.OData/Deltas/DeltaOfT.cs
dxrdxr: This code is in the existing WebAPI version that I directly ported.

@Sreejithpin 
Can you comment?

Sreejithpin: This is existing code in 7.x looks like just ported to 8.0

dxrdxr: This is just a port of WebAPI code to AspNetCoreOData.
If a change is needed, an issue for both should created to keep code in alignment.

dxrdxr: This probably could be optimized.

But this is just a port of WebAPI code to AspNetCoreOData.
If a change is needed, an issue for both should created to keep code in alignment.

src/Microsoft.AspNetCore.OData/Deltas/DeltaOfT.cs
dxrdxr: This is just a port of WebAPI code to AspNetCoreOData.
If a change is needed, an issue for both should created to keep code in alignment.

src/Microsoft.AspNetCore.OData/Deltas/DeltaOfT.cs
dxrdxr: This probably could be optimized.

But this is just a port of WebAPI code to AspNetCoreOData.
If a change is needed, an issue for both should created to keep code in alignment.

xuzhg:
@dxrdxr Please re-create your PR since the default branch is switched from "master" to "main".

PR #368, opened 2021-11-11
Migrate https://github.com/OData/WebApi/pull/2512 to AspNetCoreOData.

marabooy:
It seems the issue exists and was not patched in AspnetCoreOData.

https://stackoverflow.com/questions/69910220/create-custom-response-in-net-core-5-with-odata-8-controller.
Using `IStatusCodeActionResult` means multiple requests will be able to be checked. Users can now send JsonActionResult and other instances that implement `IStatusCodeActionResult` for their errors and have the errors propagated.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/EnableQueryAttribute.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/EnableQueryAttribute.cs
@@ -20,6 +20,7 @@ using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Abstractions;
 using Microsoft.AspNetCore.Mvc.Controllers;
 using Microsoft.AspNetCore.Mvc.Filters;
+using Microsoft.AspNetCore.Mvc.Infrastructure;
 using Microsoft.AspNetCore.OData.Common;
 using Microsoft.AspNetCore.OData.Edm;
 using Microsoft.AspNetCore.OData.Extensions;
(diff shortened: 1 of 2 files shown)


xuzhg (approved):

src/Microsoft.AspNetCore.OData/Query/EnableQueryAttribute.cs
```diff
@@ -224,8 +225,8 @@ public override void OnActionExecuted(ActionExecutedContext actionExecutedContex
             {
                 // actionExecutedContext.Result might also indicate a status code that has not yet
                 // been applied to the result; make sure it's also successful.
-                StatusCodeResult statusCodeResult = actionExecutedContext.Result as StatusCodeResult;
-                if (statusCodeResult == null || IsSuccessStatusCode(statusCodeResult.StatusCode))
+                IStatusCodeActionResult statusCodeResult = actionExecutedContext.Result as IStatusCodeActionResult;
+                if (statusCodeResult == null || statusCodeResult.StatusCode == null || IsSuccessStatusCode(statusCodeResult.StatusCode.Value))
```
gathogojr, line 229:
```suggestion
                if (statusCodeResult?.StatusCode == null || IsSuccessStatusCode(statusCodeResult.StatusCode.Value))
```

gathogojr (approved):
Left a comment, otherwise looks good

PR #393, opened 2021-11-24
Refactor SelectExpandBinder

KenitoInc:
- [x] Replace the `SelectExpandBinderContext` with `QueryBinderContext`.
- [x] Make `SelectExpandBinder.Bind***(...)` method return `Expression`.
- [x] Bind `IQueryable` in extension methods.
- [x] Make more methods of the SelectExpandBinder instance and virtual:
- CreatePropertyValueExpression
- CreatePropertyNameExpression
- CreateTotalCountExpression
- CreateTypeNameExpression
- BuildDynamicProperty


src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinderContext.cs
```diff
@@ -132,6 +132,11 @@ public QueryBinderContext(QueryBinderContext context, Type clrType)
         /// </summary>
         public IDictionary<string, Expression> ComputedProperties { get; set; }
 
+        /// <summary>
+        ///  Gets or Sets the given <see cref="ODataQueryContext"/>.
+        /// </summary>
+        public ODataQueryContext QueryContext { get; set; }
```
xuzhg, line 138: >QueryContext [](http://example.com/codeflow?start=33&length=12)

I considered it a lot. 
Finally, i think we should not include ODataQueryContext.

Because, our binder should not rely on the HttpRequest, or others. If we do need something, let's add them into the QueryBinderContext directly.

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -47,70 +47,9 @@ public virtual Expression BindSelectExpand(SelectExpandClause selectExpandClause
...
-        /// <returns>A <see cref="LambdaExpression"/> which can be later applied to an <see cref="IQueryable"/> or an <see cref="object"/>.</returns>
-        protected virtual LambdaExpression GetProjectionLambda(SelectExpandBinderContext context)
-        {
-            Contract.Assert(context != null);
-            Type elementType = context.SelectExpand.Context.ElementClrType;
-            IEdmStructuredType structuredType = context.SelectExpand.Context.ElementType as IEdmStructuredType;
-            IEdmNavigationSource navigationSource = context.SelectExpand.Context.NavigationSource;
-            SelectExpandClause selectExpandClause = context.SelectExpand.SelectExpandClause;
+            Type elementType = context.ElementClrType;
+            IEdmStructuredType structuredType = context.QueryContext.ElementType as IEdmStructuredType;
```
xuzhg, line 51: >ElementType [](http://example.com/codeflow?start=69&length=11)

Add a IEdmStructuredType property in the QueryBinderContext and use it here. 

Or Get the edm structured type from model using ElementClrType.

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -47,70 +47,9 @@ public virtual Expression BindSelectExpand(SelectExpandClause selectExpandClause
...
+            IEdmNavigationSource navigationSource = context.QueryContext.NavigationSource;
```
xuzhg, line 52: >NavigationSource [](http://example.com/codeflow?start=73&length=16)

Add a navigation source into QueryBinderContext

xuzhg:
>                 IFilterBinder binder = queryContext.GetFilterBinder();

Get the nested filterbinder from QueryBinderContext.

---
Refers to: src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs:223 in a05ca77. [](commit_id = a05ca77339548a872b46f8fc5216d2fa25b5de4b, deletion_comment = False)

xuzhg:
>                     GetNestedFilterBinder = () => binder

Since here we are creating the nested querybinder context.
Let's copy the GetNestedFilterBinder in the constructor.

So, we don't need to get the FilterBinder again.

---
Refers to: src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs:226 in a05ca77. [](commit_id = a05ca77339548a872b46f8fc5216d2fa25b5de4b, deletion_comment = False)

xuzhg:
Updated something and merged at https://github.com/OData/AspNetCoreOData/commit/65186071cdffb93b77207192eb6e1ed073022c77.

Thanks for your PR.

PR #395, opened 2021-11-29
Trim forward slashes from route prefix

### LINKED ISSUE #382: "@odata.context" is inconsistently returned
Sometimes my JSON payload includes an `@odata.context` key and a `value` key, sometimes it doesn't.

What controls this? I don't really care if it's there or not, but I need it to be consistent across projects.

habbes:
Fix #382 

This PR strips leading and trailing forward slashes from the OData path route prefix before registering route components service provider. This avoids the issues mentioned in #382 which result in the route prefix with leading slashes not being routed to OData endpoints. Route prefix with trailing slashes were causing an exception to be thrown because it would be combined with the endpoint path leading to 2 consecutive forward slashes.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.xml
+++ b/src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.xml
@@ -6190,6 +6190,13 @@
             <param name="setupAction">The setup config.</param>
             <returns>The built service provider.</returns>
         </member>
+        <member name="M:Microsoft.AspNetCore.OData.ODataOptions.SanitizeRoutePrefix(System.String)">
+            <summary>
+            Sanitizes the route prefix by stripping leading and trailing forward slashes.
+            </summary>
+            <param name="routePrefix">Route prefix to sanitize.</param>
+            <returns>Sanitized route prefix.</returns>
+        </member>
         <member name="T:Microsoft.AspNetCore.OData.ODataOptionsSetup">
             <summary>
             Sets up default options for <see cref="T:Microsoft.AspNetCore.OData.ODataOptions"/>.
--- a/src/Microsoft.AspNetCore.OData/ODataOptions.cs
+++ b/src/Microsoft.AspNetCore.OData/ODataOptions.cs
@@ -7,6 +7,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Diagnostics.Contracts;
 using Microsoft.AspNetCore.OData.Abstracts;
 using Microsoft.AspNetCore.OData.Batch;
@@ -323,7 +324,9 @@ namespace Microsoft.AspNetCore.OData
         /// <returns>Sanitized route prefix.</returns>
         private string SanitizeRoutePrefix(string routePrefix)
         {
-            if (!routePrefix.StartsWith('/') && !routePrefix.EndsWith('/'))
+            Debug.Assert(routePrefix != null);
+
+            if (routePrefix.Length > 0 && routePrefix[0] != '/' && routePrefix[^1] != '/')
             {
                 return routePrefix;
             }
--- a/src/Microsoft.AspNetCore.OData/ODataOptions.cs
+++ b/src/Microsoft.AspNetCore.OData/ODataOptions.cs
@@ -134,7 +134,7 @@ namespace Microsoft.AspNetCore.OData
                 throw Error.ArgumentNull(nameof(routePrefix));
             }
 
-            string sanitizedRoutePrefix = routePrefix.Trim('/');
+            string sanitizedRoutePrefix = SanitizeRoutePrefix(routePrefix);
 
             if (RouteComponents.ContainsKey(sanitizedRoutePrefix))
             {
@@ -155,9 +155,16 @@ namespace Microsoft.AspNetCore.OData
         /// <returns>The root service provider for the route (prefix) name.</returns>
         public IServiceProvider GetRouteServices(string routePrefix)
         {
-            if (routePrefix != null && RouteComponents.ContainsKey(routePrefix))
+            if (routePrefix == null)
             {
-                return RouteComponents[routePrefix].ServiceProvider;
+                return null;
+            }
+
+            string sanitizedRoutePrefix = SanitizeRoutePrefix(routePrefix);
+
+            if (RouteComponents.TryGetValue(sanitizedRoutePrefix, out var components))
+            {
+                return components.ServiceProvider;
             }
 
             return null;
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNetCore.OData/ODataOptions.cs
```diff
@@ -129,14 +129,22 @@ public ODataOptions AddRouteComponents(string routePrefix, IEdmModel model, Acti
...
             }
 
-            if (RouteComponents.ContainsKey(routePrefix))
+            if (routePrefix == null)
             {
-                throw Error.InvalidOperation(SRResources.ModelPrefixAlreadyUsed, routePrefix);
+                throw Error.ArgumentNull(nameof(routePrefix));
             }
 
+            string sanitizedRoutePrefix = routePrefix.Trim('/');
```
gathogojr, line 137: I believe `Trim` does both a `TrimStart` and a `TrimeEnd`. Are we okay with that?

src/Microsoft.AspNetCore.OData/ODataOptions.cs
```diff
@@ -129,14 +129,22 @@ public ODataOptions AddRouteComponents(string routePrefix, IEdmModel model, Acti
...
+            if (RouteComponents.ContainsKey(sanitizedRoutePrefix))
+            {
+                throw Error.InvalidOperation(SRResources.ModelPrefixAlreadyUsed, sanitizedRoutePrefix);
+            }
+
+
             // Consider to use Lazy<IServiceProvider> ?
             IServiceProvider serviceProvider = BuildRouteContainer(model, configureServices);
-            RouteComponents[routePrefix] = (model, serviceProvider);
+            RouteComponents[sanitizedRoutePrefix] = (model, serviceProvider);
```
gathogojr, line 147: I'm debating in my mind whether it's better to throw an exception if the user includes the slashes in the route prefix or cure the problem and move on... What was the behaviour in 7.x?

habbes: Based on a previous triage discussion. It was agreed to just sanitize the slashes. If we don't sanitize the slashes then it should be clearly documented that the prefix should not contain slashes.

src/Microsoft.AspNetCore.OData/ODataOptions.cs
```diff
@@ -300,5 +315,20 @@ private IServiceProvider BuildRouteContainer(IEdmModel model, Action<IServiceCol
...
+        /// Sanitizes the route prefix by stripping leading and trailing forward slashes.
+        /// </summary>
+        /// <param name="routePrefix">Route prefix to sanitize.</param>
+        /// <returns>Sanitized route prefix.</returns>
+        private string SanitizeRoutePrefix(string routePrefix)
+        {
+            if (!routePrefix.StartsWith('/') && !routePrefix.EndsWith('/'))
+            {
+                return routePrefix;
+            }
```
habbes, line 329: I add this check to avoid allocating a new string in case the string contains no slashes. But I confirmed that the implementation of `string.Trim` returns the original string if no actual trimming was performed. But I kept this code here for clarity and also cause it seems to be doing less work compared to `string.Trim` when there's nothing to trim.

src/Microsoft.AspNetCore.OData/ODataOptions.cs
habbes: Yeah. If you don't trim the end, an exception is thrown by AspNetCore as mentioned in the description.

src/Microsoft.AspNetCore.OData/ODataOptions.cs
gathogojr: If `string.Trim` actually does that then I'm inclined to think that this method is an overkill 'coz I'd expect that `string.Trim` does it almost same way that you're doing. Another thing... The above logic would throw an exception if `routePrefix` was null, though I assume that the null check is done elsewhere - you could just add a `Debug.Assert`. If you choose to retain this, maybe this would give us better performance:
```csharp
if (routePrefix.Length > 0 && routePrefix[0] != '/' && routePrefix[routePrefix.Length - 1] != '/'
```

src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.xml
```diff
@@ -6190,6 +6190,13 @@
             <param name="setupAction">The setup config.</param>
             <returns>The built service provider.</returns>
         </member>
+        <member name="M:Microsoft.AspNetCore.OData.ODataOptions.SanitizeRoutePrefix(System.String)">
```
gathogojr, line 6193: Do private methods cause changes to this file?

src/Microsoft.AspNetCore.OData/ODataOptions.cs
```diff
@@ -129,14 +129,22 @@ public ODataOptions AddRouteComponents(string routePrefix, IEdmModel model, Acti
                 throw Error.ArgumentNull(nameof(model));
             }
 
-            if (RouteComponents.ContainsKey(routePrefix))
+            if (routePrefix == null)
```
marabooy, line 132: Don't we have cases where the prefix is null?

src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.xml
marabooy: They do cause the file to change.

marabooy (commented):
🕚

src/Microsoft.AspNetCore.OData/ODataOptions.cs
habbes: The `MapODataRoute` method in 7.x removes the last trailing slash. I doesn't appear like it removes leading slashes.

src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.xml
habbes: I didn't change it manually. It happened automatically. So I just checked it in.

src/Microsoft.AspNetCore.OData/ODataOptions.cs
habbes: We have a case where the prefix is empty (the default case). Not sure about null though.

In any case, you can't add a `null` key to a dictionary. So I think throwing on `null` here is the right thing to do.

src/Microsoft.AspNetCore.OData/ODataOptions.cs
habbes: I do check that the prefix is not null on the public methods that call `SanitizeRoutePrefix`.

I've added the Debug.Assert and used your suggested if statement. I assume the string.StartsWith and string.EndsWith would have been inlined eitherway?

src/Microsoft.AspNetCore.OData/ODataOptions.cs
KenitoInc: I agree with @habbes. We should just sanitize the slashes instead of throwing an exception.

KenitoInc (commented):
Consider adding tests for these edge cases:
- Single slash as the route prefix
- Double slash as the route prefix
What should be the expected behaviour?

habbes:
> Consider adding tests for these edge cases:
> 
> * Single slash as the route prefix
> * Double slash as the route prefix
>   What should be the expected behaviour?

Good point. A single slash route prefix would be handled as an empty string. I think that is the correct behaviour.

A double slash will also be handled as a an empty string, but I think in this case that is invalid and ought to be an error. But I don't know if we should go the extra-length of validating such route prefixes. I personally am not sure it's worth the effort of checking the string contents.

gathogojr (approved):
:shipit:

xuzhg (approved):

marabooy (approved):

PR #400, opened 2021-12-10
Restructure AggregationBinder

KenitoInc:


src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinderContext.cs
```diff
@@ -194,6 +198,30 @@ public void RemoveParameter(string name)
...
+        //public TransformationNode Transformation { get; private set; }
+
+        public IEnumerable<AggregateExpressionBase> AggregateExpressions { get; set; }
+        public IEnumerable<GroupByPropertyNode> GroupingProperties { get; set; }
+
+        public Type GroupByClrType { get; set; }
+        public Type ResultClrType { get; set; }
+        public ParameterExpression LambdaParameter { get; set; }
+        public Type TransformationElementType { get { return this.LambdaParameter.Type; } }
+        public bool ClassicEF { get; set; }
```
xuzhg, line 211: >public bool ClassicEF { get; set; } [](http://example.com/codeflow?start=8&length=35)

will focus on EFCore only

src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinderContext.cs
```diff
@@ -194,6 +198,30 @@ public void RemoveParameter(string name)
...
+        public bool HasInstancePropertyContainer;
+
+        /// <summary>
+        /// Base query used for the binder.
+        /// </summary>
+        public IQueryable BaseQuery;
```
xuzhg, line 217: >IQueryable [](http://example.com/codeflow?start=15&length=10)

don't add IQuerable

xuzhg:
@KenitoInc Please re-create your PR since the default branch is switched from "master" to "main".

PR #410, opened 2021-12-18
a workable version of $search

MolallaComm:
I implemented this several years ago, but while upgrading to .net 6 and the newer odata, thought I would make a pull request out of it in case others wanted to use it or extend - it would be nice if $search could be pluggable or something so that we wouldn't have to periodically bring the whole odata library into our tree, but until then - this has worked well for us.

It basically takes $search=param and converts it into a bunch of ORed $filters with some intelligence as far as looking at the model and the search parameter and doing the right thing.  It does not follow how the odata spec says $search should work - but IMHO what the spec states would be fairly useless even if implemented.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
@@ -12,6 +12,7 @@ using System.Diagnostics.Contracts;
 using System.Globalization;
 using System.Linq;
 using System.Reflection;
+using System.Text.RegularExpressions;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.AspNetCore.OData.Abstracts;
@@ -189,7 +190,8 @@ namespace Microsoft.AspNetCore.OData.Query
                  fixedQueryOptionName.Equals("$format", StringComparison.Ordinal) ||
                  fixedQueryOptionName.Equals("$skiptoken", StringComparison.Ordinal) ||
                  fixedQueryOptionName.Equals("$deltatoken", StringComparison.Ordinal) ||
-                 fixedQueryOptionName.Equals("$apply", StringComparison.Ordinal);
+                 fixedQueryOptionName.Equals("$apply", StringComparison.Ordinal) ||
+                 fixedQueryOptionName.Equals("$search", StringComparison.Ordinal);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


xuzhg:
@MolallaComm We are working on the refactor query binder. For $search, would you please create a PR based on current design?

xuzhg:
@MolallaComm I enabled $search binder interface at https://github.com/OData/AspNetCoreOData/commit/ccffcf4f301955ba8def9111a0dec86199b26a53.

Would you please take a look and let us know any concerns?

MolallaComm:
I will attempt to refactor my use case for $search as a custom searchBinder that i inject and provide it as maybe a sample or something.  It will be great to have it user customizable in my code rather than in the library.

xuzhg:
@MolallaComm Let me know any problems or concerns. Thanks.

MolallaComm:
This is hopefully a dumb question, but how exactly am i supposed to inject it in Startup.cs - I tried both:

services.AddTransient<ISearchBinder, MySearchBinder>();

and

services.AddSingleton<ISearchBinder, MySearchBinder>();

but my "BindSearch" is never getting called when i do a sample $search query:

http://localhost:64771/Products?$search=Goods

MolallaComm:
Never mind - i should have kept reading further down i see it in the e2e test that was in that same commit.

MolallaComm:
This worked really well - with a little bit of trial and error, I was able to easily encapsulate the logic i had hacked in to the library previously as a custom search binding.  My users just type in a box to search and there is no way they would get all the boolean logic, parens, and other geeky stuff the spec supports - so i just changed my client to do $search="whatever they type" which forces it to use the simple SearchTermNode case always no matter what they type which is what I want.

```
using Microsoft.AspNetCore.OData.Query.Expressions;
using Microsoft.OData.Edm;
using Microsoft.OData.UriParser;
using System.Linq;
using System.Linq.Expressions;
using System.Reflection;
using System.Text.RegularExpressions;

namespace MolallaFM.WebApi
{
    public class CustomSearchBinder : ISearchBinder
    {
        //internal static readonly MethodInfo StringContainsMethodInfo = typeof(string).GetMethod("Contains", new[] { typeof(string), typeof(StringComparison) });
        internal static readonly MethodInfo StringContainsMethodInfo = typeof(string).GetMethod("Contains", new[] { typeof(string) });
        internal static readonly Regex NumExpr = new(@"^\d+$");

        public Expression BindSearch(SearchClause searchClause, QueryBinderContext context)
        {
            Expression exp = BindSearchTerm((SearchTermNode)searchClause.Expression, context);

            LambdaExpression lambdaExp = Expression.Lambda(exp, context.CurrentParameter);

            return lambdaExp;
        }

        public Expression BindSearchTerm(SearchTermNode node, QueryBinderContext context)
        {
            List<Expression> exps = new(); string text = node.Text; Expression combined = null;

            var t = (Microsoft.OData.Edm.EdmEntityType)context.Model.FindDeclaredType(context.ElementType.FullTypeName());

            if (NumExpr.Match(text).Success && t.DeclaredKey.First().Type.FullName() == "Edm.Int32")   // search text is just a number - lookup by PK if PK is an int or AccountID if that property exists and is an int
            {
                int val = int.Parse(text);
                exps.Add(Expression.Equal(Expression.Property(context.CurrentParameter, t.DeclaredKey.First().Name), Expression.Constant(val, typeof(int))));

                var accountid = t.DeclaredProperties.FirstOrDefault(p => p.Name == "AccountID" && p.Name != t.DeclaredKey.First().Name && p.Type.FullName() == "Edm.Int32");

                if (accountid != null)
                {
                    if (accountid.Type.IsNullable)
                        exps.Add(Expression.Equal(Expression.Property(context.CurrentParameter, "AccountID"), Expression.Constant(val, typeof(int?))));
                    else
                        exps.Add(Expression.Equal(Expression.Property(context.CurrentParameter, "AccountID"), Expression.Constant(val, typeof(int))));
                }
            }

            foreach (var prop in t.DeclaredProperties)
            {
                if (prop.Type.FullName() == "Edm.String")
                {
                    var notnull = Expression.NotEqual(Expression.Property(context.CurrentParameter, prop.Name), Expression.Constant(null, typeof(string)));
                    var contains = Expression.Call(Expression.Property(context.CurrentParameter, prop.Name), StringContainsMethodInfo, Expression.Constant(text, typeof(string)));

                    exps.Add(Expression.AndAlso(notnull, contains));
                }
            }

            if (exps.Count > 0)
                combined = exps.Skip(1).Aggregate(exps.FirstOrDefault(), (exp1, exp2) => Expression.Or(exp1, exp2));
            else
                combined = Expression.Constant(false, typeof(bool));

            return combined;
        }
    }
}
```

xuzhg:
@MolallaComm Glad to hear that. 

By the way, I just finished a post related to $compute and $search. 
If you are interested in, please review at https://devblogs.microsoft.com/odata/compute-and-search-in-asp-net-core-odata-8/ and leave any comments or concerns.

PR #423, opened 2022-01-09
Async query enumeration and cancellation support for long running queries

henning-krause:
Note: I've reopened this PR since the target branch name changes to main. The original PR can be found here: https://github.com/OData/AspNetCoreOData/pull/324 and was already approved by @corranrogue9.

A long time ago I created an issue to properly cancel long running queries (when the HttpContext.RequestAborted cancellation token is triggered) for the 7.9 version of this code. This was never solved, I think mainly because most of the code is still synchronous.

Now that I moved on to AspNetCoreOData I'm facing the same problem and gave it another shot.

@xuzhg I know that this PR is far from complete. But I wanted to get an opinion from you if I'm moving in the right direction with this and if you would approve this PR once it's complete before I put more effort in this.

For now, I have mainly redesigned the TruncatedCollection to properly support async enumeration (and remove the buffering of the whole response along with this change) as well as the ODataResourceSetSerializer.

So, for paged queries, this approach works.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
@@ -10,6 +10,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Diagnostics.Contracts;
+using System.Linq;
 using System.Runtime.Serialization;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http.Extensions;
@@ -145,24 +146,24 @@ namespace Microsoft.AspNetCore.OData.Formatter.Serialization
             resourceSet.NextPageLink = null;
             await writer.WriteStartAsync(resourceSet).ConfigureAwait(false);
             object lastResource = null;
-            foreach (object item in enumerable)
-            {
-                lastResource = item;
-                if (item == null || item is NullEdmComplexObject)
-                {
-                    if (elementType.IsEntity())
-                    {
-                        throw new SerializationException(SRResources.NullElementInCollection);
-                    }
 
-                    // for null complex element, it can be serialized as "null" in the collection.
-                    await writer.WriteStartAsync(resource: null).ConfigureAwait(false);
-                    await writer.WriteEndAsync().ConfigureAwait(false);
-                }
-                else
-                {
-                    await resourceSerializer.WriteObjectInlineAsync(item, elementType, writer, writeContext).ConfigureAwait(false);
-                }
+            if (enumerable is ITruncatedCollection truncatedCollection && truncatedCollection.IsAsyncEnumerationPossible)
+            {
+	            await foreach (var item in truncatedCollection.GetAsyncEnumerable().WithCancellation(writeContext.Request.HttpContext.RequestAborted))
+	            {
+		            lastResource = 
+			            await WriteSingleResultElementAsync(writer, writeContext, item, elementType, resourceSerializer)
+				            .ConfigureAwait(false);
+	            }
+            }
+            else
+            {
+	            foreach (object item in enumerable)
+	            {
+		            lastResource =
+			            await WriteSingleResultElementAsync(writer, writeContext, item, elementType, resourceSerializer)
+				            .ConfigureAwait(false);
+	            }
             }
 
             // Subtle and surprising behavior: If the NextPageLink property is set before calling WriteStart(resourceSet),
@@ -174,6 +175,30 @@ namespace Microsoft.AspNetCore.OData.Formatter.Serialization
             resourceSet.NextPageLink = nextLinkGenerator(lastResource);
 
             await writer.WriteEndAsync().ConfigureAwait(false);
+        }
+
+        private static async Task<object> WriteSingleResultElementAsync(ODataWriter writer, ODataSerializerContext writeContext,
+	        object item, IEdmStructuredTypeReference elementType, IODataEdmTypeSerializer resourceSerializer)
+        {
+	        object lastResource;
+	        lastResource = item;
+	        if (item == null || item is NullEdmComplexObject)
+	        {
+		        if (elementType.IsEntity())
+		        {
+			        throw new SerializationException(SRResources.NullElementInCollection);
+		        }
+
+		        // for null complex element, it can be serialized as "null" in the collection.
+		        await writer.WriteStartAsync(resource: null).ConfigureAwait(false);
+		        await writer.WriteEndAsync().ConfigureAwait(false);
+	        }
+	        else
+	        {
+		        await resourceSerializer.WriteObjectInlineAsync(item, elementType, writer, writeContext).ConfigureAwait(false);
+	        }
+
+	        return lastResource;
         }
 
         /// <summary>
@@ -272,7 +297,7 @@ namespace Microsoft.AspNetCore.OData.Formatter.Serialization
                 {
                     SkipTokenHandler handler = writeContext.QueryContext.GetSkipTokenHandler();
                     return (obj) => { return handler.GenerateNextPageLink(new System.Uri(writeContext.Request.GetEncodedUrl()),
-                        (writeContext.Request.ODataFeature() as ODataFeature).PageSize, obj, writeContext); };
+                        (writeContext.Request.ODataFeature() as ODataFeature).PageSize(), obj, writeContext); };
                 }
             }
             else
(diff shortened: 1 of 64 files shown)


src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -115,7 +116,7 @@ await WriteObjectInlineAsync(graph, resourceSetType, writer, writeContext)
             IEdmStructuredTypeReference elementType = GetResourceType(resourceSetType);
             ODataResourceSet resourceSet = CreateResourceSet(enumerable, resourceSetType.AsCollection(), writeContext);
 
-            Func<object, Uri> nextLinkGenerator = GetNextLinkGenerator(resourceSet, enumerable, writeContext);
+            Func<Func<object, Uri>> nextLinkGenerator = () => GetNextLinkGenerator(resourceSet, enumerable, writeContext);
```
corranrogue9, line 119: I do not follow the need for the extra `Func`. Can you help me understand?

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -171,11 +175,40 @@ await WriteObjectInlineAsync(graph, resourceSetType, writer, writeContext)
             // object before calling WriteEnd(), the next page link will be written at the end, as required for
             // odata.streaming=true support.
 
-            resourceSet.NextPageLink = nextLinkGenerator(lastResource);
+            if (originalNexPageLink != null)
+            {
+                resourceSet.NextPageLink = originalNexPageLink;
+            }
+
+            resourceSet.NextPageLink = nextLinkGenerator()(lastResource);
```
corranrogue9, line 183: This might be related to my other question about `nextLinkGenerator`, but doesn't calling this after setting the original mean that a new link is not generated?

src/Microsoft.AspNetCore.OData/Query/Container/ITruncatedCollection.cs
```diff
@@ -25,5 +26,8 @@ public interface ITruncatedCollection : IEnumerable
         /// Gets a value representing if the collection is truncated or not.
         /// </summary>
         bool IsTruncated { get; }
+
+        bool IsAsyncEnumerationPossible { get; }
```
corranrogue9, line 30: I believe that these will be breaking changes for any current implementers of this interface

src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
```diff
@@ -6,42 +6,35 @@
...
+using System.Threading;
+using System.Threading.Tasks;
 
 namespace Microsoft.AspNetCore.OData.Query.Container
 {
     /// <summary>
     /// Represents a class that truncates a collection to a given page size.
     /// </summary>
     /// <typeparam name="T">The collection element type.</typeparam>
-    public class TruncatedCollection<T> : List<T>, ITruncatedCollection, IEnumerable<T>, ICountOptionCollection
```
corranrogue9, line 18: Despite that I agree we should remove `List<T>`, this is certainly a breaking change.

src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
```diff
@@ -50,106 +43,182 @@ public TruncatedCollection(IQueryable<T> source, int pageSize) : this(source, pa
...
         /// </summary>
-        /// <param name="source">The queryable collection to be truncated.</param>
-        /// <param name="pageSize">The page size.</param>
-        /// <param name="totalCount">The total count.</param>
-        /// <param name="parameterize">Flag indicating whether constants should be parameterized</param>
-        // NOTE: The queryable version calls Queryable.Take which actually gets translated to the backend query where as 
-        // the enumerable version just enumerates and is inefficient.
-        public TruncatedCollection(IQueryable<T> source, int pageSize, long? totalCount, bool parameterize)
-            : base(pageSize > 0 ? Take(source, pageSize, parameterize) : source)
+        public bool IsAsyncEnumerationPossible => _items is IAsyncEnumerable<T>;
```
corranrogue9, line 86: I think this is trying to follow the same pattern as truncation? I don't think I agree with either approach, and I'm not sure we should be adding more precedence for this type of pattern

src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
```diff
@@ -751,12 +749,12 @@ private OrderByQueryOption EnsureStableSortOrderBy(OrderByQueryOption orderBy, O
...
         }
 
-        internal static IQueryable LimitResults(IQueryable queryable, int limit, bool parameterize, out bool resultsLimited)
+        internal static IQueryable LimitResults(IQueryable queryable, int limit, bool parameterize, out Func<bool> resultsLimited)
         {
             MethodInfo genericMethod = _limitResultsGenericMethod.MakeGenericMethod(queryable.ElementType);
             object[] args = new object[] { queryable, limit, parameterize, null };
             IQueryable results = genericMethod.Invoke(null, args) as IQueryable;
-            resultsLimited = (bool)args[3];
+            resultsLimited = (Func<bool>)args[3];
```
corranrogue9, line 757: I don't get it. `parameterize` is not a reference or anything, it always has whatever value was passed into `LimitResults`, doesn't it?

corranrogue9: This logic bubbles up I think all the page to the change of `pageSize` to a delegate

test/Microsoft.AspNetCore.OData.Tests/Query/Container/PropertyContainerTest.cs
```diff
@@ -134,6 +134,8 @@ public void CreatePropertyContainer_PageSize()
             PropertyContainer container = ToContainer(containerExpression);
             var result = container.ToDictionary(new IdentityPropertyMapper())["PropertyName"];
             var truncatedCollection = Assert.IsType<TruncatedCollection<int>>(result);
+            // Enumerate the list so that the PageSize gets initialized.
+            truncatedCollection.ToList();
```
corranrogue9, line 138: What happens if we don't enumerate the sequence? Requiring enumeration to get `PageSize` is a breaking change, I think

test/Microsoft.AspNetCore.OData.Tests/Query/Container/TruncatedCollectionOfTTest.cs
```diff
@@ -8,80 +8,35 @@
...
 using Microsoft.AspNetCore.OData.Tests.Commons;
 using System.Collections.Generic;
+using System.Linq;
 using Xunit;
 
 namespace Microsoft.AspNetCore.OData.Tests.Query.Container
 {
     public class TruncatedCollectionTest
     {
-        [Fact]
```
corranrogue9, line 17: If I understand correctly, these tests are being removed because the constructors were removed. `TruncatedCollection` is `public`, we can't just remove public constructors

test/Microsoft.AspNetCore.OData.Tests/Query/Query/ODataQueryOptionTests.cs
```diff
@@ -794,12 +794,14 @@ public void LimitResults_LimitsResults(int limit, bool resultsLimitedExpected)
             ODataQueryContext context = new ODataQueryContext(model, typeof(Customer));
```
corranrogue9, line 794: I don't actually see any tests for the new functionality, can you please add them?

src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
corranrogue9: Also, there seems to be a lack of test coverage in general, or there is something critical that I've missed about why these delegates exist, because `(Func<bool>)((object)false)` I believe will throw `InvalidCastException`

henning-krause:
Thanks for the review @corranrogue9. I'm a little swamped right now but will get back to this ASAP.

akhyarkamili:
@henning-krause hoping you'd be able to complete this sometime! Would be cool if cancellation token is going to be supported.

SymbioticKilla:
@henning-krause  any chance?

henning-krause:
@SymbioticKilla Unfortunately, I won't have time for this in the foreseeable future.

rpallares:
Hi, could you consider to push this PR info the next major version that is coming?  
It contains many critical improvements regarding performance, scalability and safery.  
It have minor brealing changes regarding truncated collection and the major release IS a good opportunity to include it.

Thanks

henning-krause:
I don't think we can merge this into main. A lot has changes since I opened this PR. Apart from this, it should be much easier now to come up with a new PR, sind IAsyncEnumerable is now supported.

PR #439, opened 2022-01-12
Adds dependency injection support in non-model scenario

### LINKED ISSUE #422: Replace ODataUriResolver for non-edm model
In v7, for non-edm model, we could replace the default `ODataUriResolver` by adding our service in the `configureAction` of the [EnableDependencyInjection](https://github.com/OData/WebApi/blob/ca0c0850c74742f6847bd357704b043438c1e81a/src/Microsoft.AspNetCore.OData/Extensions/Endpoint/ODataEndpointBuilderExtensions.cs#L625) method.

In v8, I didn't find a way to do this. For edm model, there is the `configureServices` parameter in the [AddRouteComponents](https://github.com/OData/AspNetCoreOData/blob/ed424ceee33944b7c5674536642a55d75b1eff26/src/Microsoft.AspNetCore.OData/ODataOptions.cs#L126) method, but I didn't find an equivalent for non-edm model.

kakone:
Adds a `EnableDependencyInjection(Action<IServiceCollection> configureServices)` method in `ODataOptions` to be able to add services (`StringAsEnumResolver` for example) in non-model scenario.

Fixes #422

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
@@ -271,12 +271,6 @@ namespace Microsoft.AspNetCore.OData.Extensions
                 return requestContainer;
             }
 
-            // if the prefixName == null, it's a non-model scenario
-            if (request.ODataFeature().RoutePrefix == null)
-            {
-                return null;
-            }
-
             // HTTP routes will not have chance to call CreateRequestContainer. We have to call it.
             return request.CreateRouteServices(request.ODataFeature().RoutePrefix);
         }
@@ -300,6 +294,11 @@ namespace Microsoft.AspNetCore.OData.Extensions
             }
 
             IServiceScope requestScope = request.CreateRequestScope(routePrefix);
+            if (requestScope == null)
+            {
+                // non-model scenario with dependency injection non enabled
+                return null;
+            }
             IServiceProvider requestContainer = requestScope.ServiceProvider;
 
             request.ODataFeature().RequestScope = requestScope;
@@ -342,13 +341,14 @@ namespace Microsoft.AspNetCore.OData.Extensions
             ODataOptions options = request.ODataOptions();
 
             IServiceProvider rootContainer = options.GetRouteServices(routePrefix);
+            if (rootContainer == null)
+            {
+                return null;
+            }
             IServiceScope scope = rootContainer.GetRequiredService<IServiceScopeFactory>().CreateScope();
 
-            // Bind scoping request into the OData container.
-            if (!string.IsNullOrEmpty(routePrefix))
-            {
-                scope.ServiceProvider.GetRequiredService<HttpRequestScope>().HttpRequest = request;
-            }
+            // Bind scoping request into the OData container.            
+            scope.ServiceProvider.GetRequiredService<HttpRequestScope>().HttpRequest = request;
 
             return scope;
         }
--- a/src/Microsoft.AspNetCore.OData/ODataOptions.cs
+++ b/src/Microsoft.AspNetCore.OData/ODataOptions.cs
@@ -62,6 +62,19 @@ namespace Microsoft.AspNetCore.OData
         /// </summary>
         public ODataRouteOptions RouteOptions { get; } = new ODataRouteOptions();
 
+        private IServiceProvider ServiceProvider { get; set; }
+
+        /// <summary>
+        /// Enables dependency injection support for non-model scenario
+        /// </summary>
+        /// <param name="configureServices">The configuring action to add the services to the container.</param>
+        /// <returns>The current <see cref="ODataOptions"/> instance to enable fluent configuration.</returns>
+        public ODataOptions EnableDependencyInjection(Action<IServiceCollection> configureServices)
+        {
+            ServiceProvider = BuildContainer(configureServices);
+            return this;
+        }
+
         #endregion
 
         #region RouteComponents
@@ -144,7 +157,7 @@ namespace Microsoft.AspNetCore.OData
 
 
             // Consider to use Lazy<IServiceProvider> ?
-            IServiceProvider serviceProvider = BuildRouteContainer(model, configureServices);
+            IServiceProvider serviceProvider = BuildContainer(configureServices, model);
             RouteComponents[sanitizedRoutePrefix] = (model, serviceProvider);
             return this;
         }
@@ -158,7 +171,7 @@ namespace Microsoft.AspNetCore.OData
         {
             if (routePrefix == null)
             {
-                return null;
+                return ServiceProvider;
             }
 
             string sanitizedRoutePrefix = SanitizeRoutePrefix(routePrefix);
@@ -284,13 +297,11 @@ namespace Microsoft.AspNetCore.OData
         /// <summary>
         /// Build the container.
         /// </summary>
-        /// <param name="model">The Edm model.</param>
         /// <param name="setupAction">The setup config.</param>
+        /// <param name="model">The Edm model.</param>
         /// <returns>The built service provider.</returns>
-        private IServiceProvider BuildRouteContainer(IEdmModel model, Action<IServiceCollection> setupAction)
+        private IServiceProvider BuildContainer(Action<IServiceCollection> setupAction, IEdmModel model = null)
         {
-            Contract.Assert(model != null);
-
             ServiceCollection services = new ServiceCollection();
             DefaultContainerBuilder builder = new DefaultContainerBuilder();
 
@@ -306,10 +317,13 @@ namespace Microsoft.AspNetCore.OData
             // Set Uri resolver to by default enabling unqualified functions/actions and case insensitive match.
             builder.Services.AddSingleton<ODataUriResolver>(sp => new UnqualifiedODataUriResolver { EnableCaseInsensitive = true });
 
-            // Inject the Edm model.
-            // From Current ODL implement, such injection only be used in reader and writer if the input
-            // model is null.
-            builder.Services.AddSingleton(sp => model);
+            if (model != null)
+            {
+                // Inject the Edm model.
+                // From Current ODL implement, such injection only be used in reader and writer if the input
+                // model is null.
+                builder.Services.AddSingleton(sp => model);
+            }
 
             // Inject the customized services.
             setupAction?.Invoke(builder.Services);
--- a/src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.xml
+++ b/src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.xml
@@ -6102,6 +6102,13 @@
             Gets the <see cref="P:Microsoft.AspNetCore.OData.ODataOptions.RouteOptions"/> instance responsible for configuring the route template.
             </summary>
         </member>
+        <member name="M:Microsoft.AspNetCore.OData.ODataOptions.EnableDependencyInjection(System.Action{Microsoft.Extensions.DependencyInjection.IServiceCollection})">
+            <summary>
+            Enables dependency injection support for non-model scenario
+            </summary>
+            <param name="configureServices">The configuring action to add the services to the container.</param>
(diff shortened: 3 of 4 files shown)


xuzhg:
@kakone 

1) I don't like the method name
2) Where will you use 'StringAsEnumResolver'?

kakone:
> 1. I don't like the method name

I agree, I don't like it too. I chose it because it was the [method name I used with v7](https://github.com/OData/WebApi/blob/46a0d5a8ef34938066e65c0afc4fa5ca956f971e/src/Microsoft.AspNetCore.OData/Extensions/ODataRouteBuilderExtensions.cs#L621). A suggestion for another name ?

> 2. Where will you use 'StringAsEnumResolver'?

I have a simple webservice with `[EnableQuery]` attribute and a non-edm model. The registered `StringAsEnumResolver` will be retrieved automatically in the [`ODataQueryOptions` object](https://github.com/OData/AspNetCoreOData/blob/a4ada0f0038434caeda97b9524b4f770d461626e/src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs#L1103) and transmitted to the `ODataQueryOptionParser`.
Then, I can use an OData filter like `State eq '2'` instead of `State eq 'Completed'`.

dquilong:
Nice work! it's exactly what I needed. Please if you can merge it and create the realesase it as soon as possible. Thanks a lot!

src/Microsoft.AspNetCore.OData/ODataOptions.cs
```diff
@@ -62,6 +62,19 @@ public class ODataOptions
...
         public ODataRouteOptions RouteOptions { get; } = new ODataRouteOptions();
 
+        private IServiceProvider ServiceProvider { get; set; }
+
+        /// <summary>
+        /// Enables dependency injection support for non-model scenario
+        /// </summary>
+        /// <param name="configureServices">The configuring action to add the services to the container.</param>
+        /// <returns>The current <see cref="ODataOptions"/> instance to enable fluent configuration.</returns>
+        public ODataOptions EnableDependencyInjection(Action<IServiceCollection> configureServices)
```
corranrogue9, line 72: Maybe `ConfigureServiceCollection`?

corranrogue9 (changes requested):
I really like this change, but I think it is very important to add a test case

src/Microsoft.AspNetCore.OData/ODataOptions.cs
kakone: OK, I renamed it

src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
```diff
@@ -341,14 +340,15 @@ private static IServiceScope CreateRequestScope(this HttpRequest request, string
...
+            IServiceProvider rootContainer = options?.GetRouteServices(routePrefix);
+            if (rootContainer == null)
             {
-                scope.ServiceProvider.GetRequiredService<HttpRequestScope>().HttpRequest = request;
+                return null;
             }
+            IServiceScope scope = rootContainer.GetRequiredService<IServiceScopeFactory>().CreateScope();
+
+            // Bind scoping request into the OData container.            
+            scope.ServiceProvider.GetRequiredService<HttpRequestScope>().HttpRequest = request;
```
corranrogue9, line 351: Do we still need the check for `if (!string.IsNullOrEmpty(routePrefix))`?

corranrogue9: I suppose the `null` part is already covered. Do we care about the "empty" part?

test/Microsoft.AspNetCore.OData.E2E.Tests/NonEdm/ConfigureServiceCollectionTest.cs
```diff
@@ -0,0 +1,42 @@
...
+                {
+                    services.AddSingleton<ODataUriResolver>(sp => new StringAsEnumResolver() { EnableCaseInsensitive = true });
+                });
+            });
+        }
+
+        [Fact]
+        public async Task EnableConfigureServiceCollectionTest()
+        {
+            HttpResponseMessage response = await CreateClient().SendAsync(new HttpRequestMessage(HttpMethod.Get, $"api/Customers?$filter=Gender eq 'MaLe'"));
```
corranrogue9, line 37: `HttpResponseMessage` implements `IDisposable`. Please wrap it in a `using` block

corranrogue9:
I left a couple of small comments, but I really like this change, thanks for adding the test case.

@xuzhg, since you are more familiar with this repo can you take a look to make sure it fits?

src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
kakone: This check was unnecessary because when `routePrefix` is null or empty, the `rootContainer` is null. So we can't reach the `if (!string.IsNullOrEmpty(routePrefix))` check.

test/Microsoft.AspNetCore.OData.E2E.Tests/NonEdm/NonEdmController.cs
```diff
@@ -0,0 +1,43 @@
+﻿using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.OData.Query;
+using System.Collections.Generic;
+using System.Linq;
```
xuzhg, line 4: copy right head

test/Microsoft.AspNetCore.OData.E2E.Tests/NonEdm/NonEdmController.cs
```diff
@@ -0,0 +1,43 @@
...
+
```
xuzhg, line 5: move using System.* ahead
and sort

test/Microsoft.AspNetCore.OData.E2E.Tests/NonEdm/NonEdmController.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            _customers = Enumerable.Range(1, 5).Select(e =>
+                new Customer
+                {
+                    Id = e,
+                    Name = "Customer #" + e,
+                    Gender = e%2 == 0 ? Gender.Female : Gender.Male,
+                }).ToList();
+        }
+    }
```
xuzhg, line 42: put it into a new file

test/Microsoft.AspNetCore.OData.E2E.Tests/NonEdm/NonEdmController.cs
```diff
@@ -0,0 +1,43 @@
...
+{
+    [ApiController]
+    [Route("api/[controller]")]
+    public class CustomersController : ControllerBase
+    {
+        [HttpGet]
+        public IActionResult Get(ODataQueryOptions<Customer> options)
+        {
+            return Ok(options.ApplyTo(NonEdmDbContext.GetCustomers().AsQueryable()));
+        }
```
xuzhg, line 16: add a test using [EnableQuery] attribute

xuzhg:
> namespace Microsoft.AspNetCore.OData.E2E.Tests.NonEdm

add copyright

---
Refers to: test/Microsoft.AspNetCore.OData.E2E.Tests/NonEdm/NonEdmDataModel.cs:1 in e305e35. [](commit_id = e305e352207b61b572e68e43fc568ea3b2e1a197, deletion_comment = False)

test/Microsoft.AspNetCore.OData.E2E.Tests/NonEdm/ConfigureServiceCollectionTest.cs
```diff
@@ -0,0 +1,42 @@
...
+                opt.EnableQueryFeatures();
+                opt.ConfigureServiceCollection(services =>
+                {
+                    services.AddSingleton<ODataUriResolver>(sp => new StringAsEnumResolver() { EnableCaseInsensitive = true });
+                });
+            });
+        }
+
+        [Fact]
+        public async Task EnableConfigureServiceCollectionTest()
```
xuzhg, line 33: >EnableConfigureServiceCollectionTest [](http://example.com/codeflow?start=26&length=36)

add new tests, for example, $select, $expand, $top to make sure it's working.

src/Microsoft.AspNetCore.OData/ODataOptions.cs
```diff
@@ -62,6 +62,19 @@ public class ODataOptions
...
         public ODataRouteOptions RouteOptions { get; } = new ODataRouteOptions();
 
+        private IServiceProvider ServiceProvider { get; set; }
+
+        /// <summary>
+        /// Configures service collection for non-EDM scenario
+        /// </summary>
+        /// <param name="configureServices">The configuring action to add the services to the container.</param>
+        /// <returns>The current <see cref="ODataOptions"/> instance to enable fluent configuration.</returns>
+        public ODataOptions ConfigureServiceCollection(Action<IServiceCollection> configureServices)
```
xuzhg, line 72: >ConfigureServiceCollection [](http://example.com/codeflow?start=28&length=26)

maybe just use AddRouteComponents as the method name

src/Microsoft.AspNetCore.OData/ODataOptions.cs
```diff
@@ -62,6 +62,19 @@ public class ODataOptions
         /// </summary>
         public ODataRouteOptions RouteOptions { get; } = new ODataRouteOptions();
 
+        private IServiceProvider ServiceProvider { get; set; }
```
xuzhg, line 65: >private IServiceProvider ServiceProvider { get; set; } [](http://example.com/codeflow?start=8&length=54)

move the private field ahead

src/Microsoft.AspNetCore.OData/ODataOptions.cs
```diff
@@ -284,13 +297,11 @@ public ODataOptions SetMaxTop(int? maxTopValue)
         /// <summary>
         /// Build the container.
         /// </summary>
-        /// <param name="model">The Edm model.</param>
         /// <param name="setupAction">The setup config.</param>
+        /// <param name="model">The Edm model.</param>
         /// <returns>The built service provider.</returns>
-        private IServiceProvider BuildRouteContainer(IEdmModel model, Action<IServiceCollection> setupAction)
+        private IServiceProvider BuildContainer(Action<IServiceCollection> setupAction, IEdmModel model = null)
```
xuzhg, line 303: >BuildContainer [](http://example.com/codeflow?start=33&length=14)

it seems we don't need to change the method name

src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
```diff
@@ -341,14 +340,15 @@ private static IServiceScope CreateRequestScope(this HttpRequest request, string
...
-            IServiceScope scope = rootContainer.GetRequiredService<IServiceScopeFactory>().CreateScope();
-
-            // Bind scoping request into the OData container.
-            if (!string.IsNullOrEmpty(routePrefix))
+            IServiceProvider rootContainer = options?.GetRouteServices(routePrefix);
+            if (rootContainer == null)
             {
-                scope.ServiceProvider.GetRequiredService<HttpRequestScope>().HttpRequest = request;
+                return null;
             }
```
xuzhg, line 347: insert whiteline after {} block

xuzhg:
@kakone It looks ok to me but would you please take a look for my comments.
Besides, do you think we should make this available directly calling AddOData();

For example:

services.AddOData(); // this can inject the odata serivces into the global level service provider??
 

---
In reply to: [1169031180](https://github.com/OData/AspNetCoreOData/pull/439#issuecomment-1169031180) [](http://example.com/codeflow?ancestors=1169031180,1169031180)

KenitoInc:
@kakone Did you get a chance to look at the comments?

soargon:
please merge this pr quickly

PhilippeAntom:
Great! It would be nice to have this PR merged and released! Thanks for your hard work!

zang227:
Hey is there any update on this case? We're having the same issue that Kakone had and this would resolve it for us.

wertzui:
Any news on this?

PR #453, opened 2022-01-21
♻️ Nullable and non nullable type mapping registration simplification

julealgon:
This PR simplifies type mapping registrations by extracting a common method that enforces the ordering rules of nullable vs non-nullable type registrations.

This should prevent mistakes when managing these types or introducing new types in the future as well as reduce the overall amount of code duplication in the class.


julealgon:
@xuzhg

xuzhg (approved):
:shipit:

Issue #360, opened 2021-11-05
Batch reference new entity error


andygjp:
Hi,

I'm trying to create two entities and add a relationship between them in a single batch, like so:

```
POST http://localhost:5000/api/$batch
Content-Type: application/json

{
  "requests": [
    {
      "id": "1",
      "method": "POST",
      "url": "Suppliers",
      "headers": {
        "content-type": "application/json"
      },
      "body": {
        "CompanyName": "Paint Co"
      }
    },
    {
      "id": "2",
      "dependsOn": [ "1" ],
      "method": "POST",
      "url": "SalesProducts",
      "headers": {
        "content-type": "application/json"
      },
      "body": {
        "ShortDescription": "Paint"
      }
    },
    {
      "id": "3",
      "dependsOn": [ "1", "2" ],
      "method": "POST",
      "url": "$2/Suppliers/$ref",
      "headers": {
        "content-type": "application/json"
      },
      "body": {
        "@odata.id": "$1"
      }
    }
  ]
}
```

However, I get the following error:

```
System.InvalidOperationException: This operation is not supported for a relative URI.
   at System.Uri.get_Scheme()
   at Microsoft.AspNetCore.OData.Batch.HttpRequestExtensions.CopyAbsoluteUrl(HttpRequest request, Uri uri)
   at Microsoft.AspNetCore.OData.Batch.ODataBatchReaderExtensions.ReadOperationInternalAsync(ODataBatchReader reader, HttpContext originalContext, Guid batchId, Nullable`1 changeSetId, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.OData.Batch.ODataBatchReaderExtensions.ReadChangeSetRequestAsync(ODataBatchReader reader, HttpContext context, Guid batchId, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.OData.Batch.DefaultODataBatchHandler.ParseBatchRequestsAsync(HttpContext context)
   at Microsoft.AspNetCore.OData.Batch.DefaultODataBatchHandler.ProcessBatchAsync(HttpContext context, RequestDelegate nextHandler)
   at Microsoft.AspNetCore.OData.Batch.ODataBatchMiddleware.Invoke(HttpContext context)
   at Microsoft.AspNetCore.ResponseCompression.ResponseCompressionMiddleware.InvokeCore(HttpContext context)
[...]

PR #481, opened 2022-02-10
Fix to defer resolution of ContentID in Uri

### LINKED ISSUE #365: Invalid Uri for $ref request in a batch request with single changeset
Invalid Uri for $ref request in a batch request with single changeset

### Assemblies affected

- Microsoft.OData Client 7.9.2
- Microsoft.OData.Core 7.9.2
- Microsoft.AspNetCore.OData 7.x
- Microsoft.AspNetCore.OData 8.x

### Reproduce steps

Consider an OData client application containing two entities related as follows:
```csharp
public class TestA
{
    public int Id { get; set; }
    public List<TestB> TestBs { get; set; }
}

public class TestB
{
    public int Id { get; set; }
    public int Value { get; set; }
}
```
If one wishes to send a batch request with a single changeset containing `TestA` and `TestB` entities together with the link between the two, the payload generated contains an invalid `Uri` for the `$ref` request.

**Code in client application:**
```csharp
var testA = new TestA { Id = 1 };
var testB = new TestB { Id = 1, Value = 13 };
context.AddToTestAs(testA);
context.AddToTestBs(testB);
context.AddLink(testA, "TestBs", testB);

context.SaveChanges(SaveChangesOptions.BatchWithSingleChangeset);
```
**Batch payload generated internally:**
```
--batch_c6718d3f-c42b-4650-bbf2-acf33e38c65d
Content-Type: multipart/mixed; boundary=changeset_cd219c83-deef-4b21-a6cf-59f0b8d960e3

--changeset_cd219c83-deef-4b21-a6cf-59f0b8d960e3
Content-Type: application/http
Content-Transfer-Encoding: binary
Content-ID: 1

POST http://localhost/odata/TestAs HTTP/1.1
OData-Version: 4.0
OData-MaxVersion: 4.0
Content-Type: application/json;odata.metadata=minimal
Accept: application/json;odata.metadata=minimal
Accept-Charset: UTF-8
User-Agent: Microsoft.OData.Client/7.9.2

{"@odata.type":"#Ex113.Models.TestA","Id":1}
--changeset_cd219c83-deef-4b21-a6cf-59f0b8d960e3
Content-Type: application/http
Content-Transfer-Encoding: binary
Content-ID: 2

POST http://localhost/odata/TestBs HTTP/1.1
OData-Version: 4.0
OData-MaxVersion: 4.0
Content-Type: application/json;odata.metadata=minimal
Accept: application/json;odata.metadata=minimal
Accept-Charset: UTF-8
User-Agent: Microsoft.OData.Client/7.9.2

{"@odata.type":"#Ex113.Models.TestB","Id":1,"Value":13}
--changeset_cd219c83-deef-4b21-a6cf-59f0b8d960e3
Content-Type: application/http
Content-Transfer-Encoding: binary
Content-ID: 3

POST $1/TestBs/$ref HTTP/1.1
OData-Version: 4.0
OData-MaxVersion: 4.0
Accept: application/json;odata.metadata=minimal
Accept-Charset: UTF-8
Content-Type: application/json;odata.metadata=minimal
User-Agent: Microsoft.OData.Client/7.9.2

{"@odata.id":"$2"}
--changeset_cd219c83-deef-4b21-a6cf-59f0b8d960e3--
--batch_c6718d3f-c42b-4650-bbf2-acf33e38c65d--
```

### Expected result

Valid Uri: POST http://localhost/TestBs/$ref HTTP/1.1

### Actual result

Invalid Uri: POST $1/TestBs/$ref HTTP/1.1

### Additional Details
This issue is not observed if the parent item already exists, e.g., if `TestA` entity was retrieved from the service.

gathogojr:
Fixes #365

This PR fixes the issue by deferring resolution of ContentID in Uri (e.g. `$1/TestBs/$ref`) to much later when the dependent operations have already been executed and the ContentID mapped to a location for a resource. That will be done here:
https://github.com/OData/AspNetCoreOData/blob/69eec03c7003fe12d92cdc619efdc16781683694/src/Microsoft.AspNetCore.OData/Batch/ODataBatchRequestItem.cs#L44-L56
`CopyAbsoluteUrl(Uri)` is called and the request Url has been resolved

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Batch/ODataBatchReaderExtensions.cs
@@ -134,7 +134,14 @@ namespace Microsoft.AspNetCore.OData.Batch
             HttpRequest request = context.Request;
 
             request.Method = batchRequest.Method;
-            request.CopyAbsoluteUrl(batchRequest.Url);
+            if (batchRequest.Url.IsAbsoluteUri)
+            {
+                request.CopyAbsoluteUrl(batchRequest.Url);
+            }
+            else
+            {
+                request.Path = new PathString(string.Concat("/", batchRequest.Url.OriginalString));
+            }
 
             // Not using bufferContentStream. Unlike AspNet, AspNetCore cannot guarantee the disposal
             // of the stream in the context of execution so there is no choice but to copy the stream
--- a/src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.xml
+++ b/src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.xml
@@ -1089,6 +1089,20 @@
             <param name="clrType">The type to test.</param>
             <returns>True if the type is a DateTime; false otherwise.</returns>
         </member>
+        <member name="M:Microsoft.AspNetCore.OData.Common.TypeHelper.IsDateOnly(System.Type)">
+            <summary>
+            Determine if a type is a <see cref="T:System.DateOnly"/>.
+            </summary>
+            <param name="clrType">The type to test.</param>
+            <returns>True if the type is a DateOnly; false otherwise.</returns>
+        </member>
+        <member name="M:Microsoft.AspNetCore.OData.Common.TypeHelper.IsTimeOnly(System.Type)">
+            <summary>
+            Determine if a type is a <see cref="T:System.TimeOnly"/>.
+            </summary>
+            <param name="clrType">The type to test.</param>
+            <returns>True if the type is a TimeOnly; false otherwise.</returns>
+        </member>
         <member name="M:Microsoft.AspNetCore.OData.Common.TypeHelper.IsTimeSpan(System.Type)">
             <summary>
             Determine if a type is a TimeSpan.
--- a/src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.xml
+++ b/src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.xml
@@ -1089,20 +1089,6 @@
             <param name="clrType">The type to test.</param>
             <returns>True if the type is a DateTime; false otherwise.</returns>
         </member>
-        <member name="M:Microsoft.AspNetCore.OData.Common.TypeHelper.IsDateOnly(System.Type)">
-            <summary>
-            Determine if a type is a <see cref="T:System.DateOnly"/>.
-            </summary>
-            <param name="clrType">The type to test.</param>
-            <returns>True if the type is a DateOnly; false otherwise.</returns>
-        </member>
-        <member name="M:Microsoft.AspNetCore.OData.Common.TypeHelper.IsTimeOnly(System.Type)">
-            <summary>
-            Determine if a type is a <see cref="T:System.TimeOnly"/>.
-            </summary>
-            <param name="clrType">The type to test.</param>
-            <returns>True if the type is a TimeOnly; false otherwise.</returns>
-        </member>
         <member name="M:Microsoft.AspNetCore.OData.Common.TypeHelper.IsTimeSpan(System.Type)">
             <summary>
             Determine if a type is a TimeSpan.
--- a/test/Microsoft.AspNetCore.OData.E2E.Tests/Batch/DefaultODataBatchHandlerController.cs
+++ b/test/Microsoft.AspNetCore.OData.E2E.Tests/Batch/DefaultODataBatchHandlerController.cs
@@ -49,16 +49,28 @@ namespace Microsoft.AspNetCore.OData.E2E.Tests.Batch
             return Created(customer);
         }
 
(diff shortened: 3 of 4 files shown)


mikepizzo:
Do you have an added test that validates that the correct URL is executed (with $1 replaced by the locationheader returned by the request with id = 1)?  When I tried a similar fix I got back a 404 for the second request because $1/TestBs/$ref was invalid (the $1 was never resolved).

gathogojr:
> Do you have an added test that validates that the correct URL is executed (with $1 replaced by the locationheader returned by the request with id = 1)? When I tried a similar fix I got back a 404 for the second request because $1/TestBs/$ref was invalid (the $1 was never resolved).

@mikepizzo I applied the fix to the project I used to repro the issue and it works perfectly. Here's the batch payload I'm basing the test on and the CreateRef controller action is being hit. The `$1` in the URL of the 3rd sub request is replaced successfully. I might need to understand what scenario is failing so I can try it out
```
--batch_c6718d3f-c42b-4650-bbf2-acf33e38c65d
Content-Type: multipart/mixed; boundary=changeset_cd219c83-deef-4b21-a6cf-59f0b8d960e3

--changeset_cd219c83-deef-4b21-a6cf-59f0b8d960e3
Content-Type: application/http
Content-Transfer-Encoding: binary
Content-ID: 1

POST http://localhost/odata/TestAs HTTP/1.1
OData-Version: 4.0
OData-MaxVersion: 4.0
Content-Type: application/json;odata.metadata=minimal
Accept: application/json;odata.metadata=minimal
Accept-Charset: UTF-8
User-Agent: Microsoft.OData.Client/7.9.2

{"@odata.type":"#Ex113.Models.TestA","Id":1}
--changeset_cd219c83-deef-4b21-a6cf-59f0b8d960e3
Content-Type: application/http
Content-Transfer-Encoding: binary
Content-ID: 2

POST http://localhost/odata/TestBs HTTP/1.1
OData-Version: 4.0
OData-MaxVersion: 4.0
Content-Type: application/json;odata.metadata=minimal
Accept: application/json;odata.metadata=minimal
Accept-Charset: UTF-8
User-Agent: Microsoft.OData.Client/7.9.2

{"@odata.type":"#Ex113.Models.TestB","Id":1,"Value":13}
--changeset_cd219c83-deef-4b21-a6cf-59f0b8d960e3
Content-Type: application/http
Content-Transfer-Encoding: binary
Content-ID: 3

POST $1/TestBs/$ref HTTP/1.1
OData-Version: 4.0
OData-MaxVersion: 4.0
Accept: application/json;odata.metadata=minimal
Accept-Charset: UTF-8
Content-Type: application/json;odata.metadata=minimal
User-Agent: Microsoft.OData.Client/7.9.2

{"@odata.id":"$2"}
--changeset_cd219c83-deef-4b21-a6cf-59f0b8d960e3--
--batch_c6718d3f-c42b-4650-bbf2-acf33e38c65d--
```

mikepizzo:
Okay; I see the disconnect.  This *does* fix the absolute Uri issue, but there is a related issue that also needs to be addressed, which is that $<ids> can reference any request, not just in the same changeset (that was an old restriction that has been relaxed).  So, for example, the repro in https://github.com/OData/AspNetCoreOData/issues/360 would still not work, because the two requests are not grouped together in an atomicityGroup.

I have a fix for that, as well as the relative URL issue that this fixes, which I'll share in a draft PR and we can discuss...

gathogojr:
> $ids can reference any request, not just in the same changeset (that was an old restriction that has been relaxed)
@mikepizzo I wasn't aware of that. Would you know if we ever spec'ed and implemented that? Can the customer apply the atomicityGroup such that the fix for the relative Url issue works for them in the interim?

mikepizzo:
See https://github.com/OData/AspNetCoreOData/pull/482 for an alternate approach.

1) Rather than create a relativeUri by prepending "/", #482 uses the baseUri to create an absolute URI.
2) Shares ContentIdToLocationMapping across the request, not just within a changeset.

gathogojr:
> See #482 for an alternate approach.
> 
> 1. Rather than create a relativeUri by prepending "/", [Fix reference handling in $batch #482](https://github.com/OData/AspNetCoreOData/pull/482) uses the baseUri to create an absolute URI.
> 2. Shares ContentIdToLocationMapping across the request, not just within a changeset.

@gathogojr I see a potential problem here. Here is the scenario:
POST resource A => returns location (Url) for resource A - http://tempuri.org/As(1)
POST resource B => returns location (Url) for resource B - http://tempuri.org/Bs(1)
POST link A to B => Url needs to be http://tempuri.org/As(1)/Bs/$ref where `Bs` is a navigation property

If all the requests are in the same batch and you append base Uri at the point you're doing it, before we have resolved the location for resource A (that corresponds to `$1` in the Url for the /$ref request), will that give us the desired Url for the /$ref request?

gathogojr:
https://github.com/OData/AspNetCoreOData/pull/481#issuecomment-1035885558
Let's verify that with a test. I'll push one shortly then you can add it to your PR

gathogojr:
Addressed in https://github.com/OData/AspNetCoreOData/pull/482

PR #519, opened 2022-03-10
Adding support for case insensitive action names

giulianob:
I am working with an API which uses camel case OData action names but these are not found because the routing is case sensitive. For example, the following will not work:
```diff
// sample/ODataCustomizedSample/Models/EnumsEdmModel.cs:75
- var actionConfiguration = employee.Action("AddSkill");
+ var actionConfiguration = employee.Action("addSkill");
```
This fails because `OperationRoutingConvension` is case sensitive. Changing the following makes it work.
```diff
// src/Microsoft.AspNetCore.OData/Routing/Conventions/OperationRoutingConvention.cs:91
- IEnumerable<IEdmOperation> candidates = context.Model.SchemaElements.OfType<IEdmOperation>().Where(f => f.IsBound && f.Name == operationName);
+ IEnumerable<IEdmOperation> candidates = context.Model.SchemaElements.OfType<IEdmOperation>().Where(f => f.IsBound && string.Equals(f.Name, operationName, StringComparison.InvariantCultureIgnoreCase));
```

I have introduced the property `EnableActionNameCaseInsensitive` to optionally enable case insensitive action names.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Routing/Conventions/OperationRoutingConvention.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/Conventions/OperationRoutingConvention.cs
@@ -88,7 +88,12 @@ namespace Microsoft.AspNetCore.OData.Routing.Conventions
 
             // TODO: refactor here
             // If we have multiple same function defined, we should match the best one?
-            IEnumerable<IEdmOperation> candidates = context.Model.SchemaElements.OfType<IEdmOperation>().Where(f => f.IsBound && f.Name == operationName);
+            StringComparison actionNameComparison = context.Options?.RouteOptions?.EnableActionNameCaseInsensitive switch
+            {
+                true => StringComparison.InvariantCultureIgnoreCase,
+                _ => StringComparison.InvariantCulture
+            };
+            IEnumerable<IEdmOperation> candidates = context.Model.SchemaElements.OfType<IEdmOperation>().Where(f => f.IsBound && string.Equals(f.Name, operationName, actionNameComparison));
             foreach (IEdmOperation edmOperation in candidates)
             {
                 IEdmOperationParameter bindingParameter = edmOperation.Parameters.FirstOrDefault();
--- a/src/Microsoft.AspNetCore.OData/Routing/Conventions/OperationRoutingConvention.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/Conventions/OperationRoutingConvention.cs
@@ -79,7 +79,7 @@ namespace Microsoft.AspNetCore.OData.Routing.Conventions
                     return;
                 }
 
-                castTypeFromActionName = entityType.FindTypeInInheritance(context.Model, cast) as IEdmEntityType;
+                castTypeFromActionName = entityType.FindTypeInInheritance(context.Model, cast, context.Options?.RouteOptions?.EnableActionNameCaseInsensitive == true) as IEdmEntityType;
                 if (castTypeFromActionName == null)
                 {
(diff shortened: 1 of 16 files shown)


corranrogue9:
@giulianob , thanks for the contribution! Did you file an issue for this? I'm pretty happy with the change, but it looks like the update to the API is causing a test failure. Can you take a look at that?

giulianob:
@corranrogue9 Updated the API definitions. Do you want me to create an issue? I ran into the issue and created the PR.

src/Microsoft.AspNetCore.OData/Routing/Conventions/OperationRoutingConvention.cs
```diff
@@ -88,7 +88,12 @@ protected void ProcessOperations(ODataControllerActionContext context, IEdmEntit
 
             // TODO: refactor here
             // If we have multiple same function defined, we should match the best one?
-            IEnumerable<IEdmOperation> candidates = context.Model.SchemaElements.OfType<IEdmOperation>().Where(f => f.IsBound && f.Name == operationName);
+            StringComparison actionNameComparison = context.Options?.RouteOptions?.EnableActionNameCaseInsensitive switch
+            {
+                true => StringComparison.InvariantCultureIgnoreCase,
+                _ => StringComparison.InvariantCulture
+            };
```
xuzhg, line 95: why don't you use " ? : ", it seems more clear

xuzhg, line 95: why do we only apply the configuration on in operation routing convention? not in other conventions?

test/Microsoft.AspNetCore.OData.Tests/Routing/Conventions/ActionRoutingConventionTests.cs
```diff
@@ -17,14 +17,21 @@
 using Microsoft.AspNetCore.OData.Tests.Extensions;
 using Microsoft.OData.Edm;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace Microsoft.AspNetCore.OData.Tests.Routing.Conventions
 {
     public class ActionRoutingConventionTests
     {
+        private readonly ITestOutputHelper _testOutputHelper;
```
xuzhg, line 26: >_testOutputHelper [](http://example.com/codeflow?start=43&length=17)

It seems there's no place using this output?

xuzhg (changes requested):
:clock1:

src/Microsoft.AspNetCore.OData/Routing/Conventions/OperationRoutingConvention.cs
giulianob: I went through the conventions and made them case insensitive. Thanks!

test/Microsoft.AspNetCore.OData.Tests/Routing/Conventions/ActionRoutingConventionTests.cs
giulianob: Fixed

src/Microsoft.AspNetCore.OData/Routing/Conventions/OperationRoutingConvention.cs
giulianob: Changed

giulianob:
@xuzhg I resolved all of your comments. Thanks for the review.

xuzhg (approved):

Issue #420, opened 2022-01-06
problem with $apply and $orderby


MolallaComm:
It appears the group by issues discussed in https://github.com/OData/AspNetCoreOData/issues/76 are partially fixed using efcore6, .net core 6, and latest odata which is good news. If I do just a simple group by:

/odata/Item?$apply=groupby((ComponentID),aggregate($count%20as%20GroupCount))&$count=true

it works.  If I tack an $orderby on the end however, it stops working:

/odata/Item?$apply=groupby((ComponentID),aggregate($count%20as%20GroupCount))&$count=true&$orderby=ComponentID

Yields: 

The query specified in the URI is not valid. The given model does not contain the type 'Microsoft.AspNetCore.OData.Query.Wrapper.AggregationWrapper'

I can see on line 54 of QueryBinderContext.cs where it is looking for that AggregationWrapper in my model, and it obviously isn't there.  If I comment that out, it gets further, but then pukes with:

Instance property 'ComponentID' is not defined for type 'Microsoft.AspNetCore.OData.Query.Wrapper.AggregationWrapper'

It seems like the basic expression processing is smart enough to deal with AggregationWrapper, but the orderby expression processing code is not.  This worked in 3.1 as long as your controller pulled everything into memory first using something like:

```
            if (Request.QueryString.Value != null && Request.QueryString.Value.Contains("$apply"))
                return db.Item.ToList().AsQueryable();
            else
                return db.Item;
````

That fix no longer works in latest however which has me bummed as we use $apply with $orderby a lot.

xuzhg:
@MolallaComm Thanks for reporting this. 

For the query binding refactor, we just finished query except for $apply.for $apply, it's in https://github.com/OData/AspNetCoreOData/pull/400

MolallaComm:
FWIW, I tried running:

http://localhost:64771/Products?$apply=groupby((Category),aggregate($count%20as%20GroupCount))&$count=true&$orderby=Category

against #400 from KenitoInc’s repo and still get the error:

`The query specified in the URI is not valid. The given model does not contain the type 'Microsoft.AspNetCore.OData.Query.Wrapper.AggregationWrapper'.`

So it doesn't appear to fully address the issue unfortunately.

henning-krause:
I've run into the same issue today.

henning-krause:
[...]

PR #543, opened 2022-03-30
fix for issue 420 ($apply and $orderby together)

MolallaComm:
issue #420

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinderContext.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinderContext.cs
@@ -10,6 +10,7 @@ using System.Collections.Generic;
 using System.Diagnostics.Contracts;
 using System.Linq.Expressions;
 using Microsoft.AspNetCore.OData.Edm;
+using Microsoft.AspNetCore.OData.Query.Wrapper;
 using Microsoft.OData;
 using Microsoft.OData.Edm;
 using Microsoft.OData.ModelBuilder;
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinderContext.cs
```diff
@@ -43,17 +44,24 @@ public class QueryBinderContext
         /// <param name="model">The Edm model.</param>
         /// <param name="querySettings">The query setting.</param>
         /// <param name="clrType">The current element CLR type in this context (scope).</param>
-        public QueryBinderContext(IEdmModel model, ODataQuerySettings querySettings, Type clrType)
+        public QueryBinderContext(IEdmModel model, ODataQuerySettings querySettings, Type clrType, Type beforeAggregationClrType = null)
```
xuzhg, line 47: >beforeAggregationClrType [](http://example.com/codeflow?start=104&length=24)

Let's don't add this in the constructor, move it as internal property set.

src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinderContext.cs
```diff
@@ -125,7 +133,17 @@ public QueryBinderContext(QueryBinderContext context, ODataQuerySettings querySe
...
+
+        /// <summary>
+        /// Gets the CLR type of the element before the aggregation wrapper override.
+        /// </summary>
+        public Type BeforeAggregationClrType { get; }
+
+        /// <summary>
+        /// Gets the CLR type of the element after the aggregation wrapper override.
+        /// </summary>
+        public Type AfterAggregationClrType { get; }
```
xuzhg, line 146: Let's move all of these internal, It could be changed in the apply refactor.

src/Microsoft.AspNetCore.OData/Query/ODataQueryContext.cs
```diff
@@ -140,6 +140,8 @@ public DefaultQuerySettings DefaultQuerySettings
         /// </summary>
         public Type ElementClrType { get; internal set; }
 
+        public Type BeforeAggregationClrType { get; internal set; }
```
xuzhg, line 143: make it as internal

xuzhg (approved):
:shipit:

xuzhg:
I'd like to take this changes to unblock customers temporarily. But, As I see, it will be changed in the $apply refactor.
So, please make the changes as internal as much as possible. 

Thanks.

habbes:
The tests are failing because of changes to the public API not updated in the baseline files. Was this PR expected to change the Public API? @xuzhg requested changes to be applied to `internal` API.

@MolallaComm could you add test cases to this PR?

xuzhg:
Merged at https://github.com/OData/AspNetCoreOData/pull/763.

Close this PR. thanks.

Issue #567, opened 2022-04-15
Slash character ('/') is not proper handled at function or action in quoted string value segment


wjmirror:
**Assemblies affected**
Microsoft.AspNetCore.OData

**Describe the bug**
Slash character ('/') is not proper handled at function or action in quoted string value segment

**Reproduce steps**

1. In **ODataRoutingSample** project,  PeopleController, initialize the IList<Person> as below.  see I changed the first person's LastName as "Zha/ngg" .  
```c#
private static IList<Person> _persons = new List<Person>
        {
            new Person
            {
                FirstName = "Goods",
                LastName = "Zha/ngg",
            },
            new Person
            {
                FirstName = "Magazine",
                LastName = "Jingchan",
            },
            new Person
            {
                FirstName = "Fiction",
                LastName = "Hollewye"
            },
        };
``` 
2. Run the sample. 
Get: /People(FirstName='Goods',LastName='Zha%2Fngg') 

It returns  

```json
{
"@odata.context": "https://localhost:64772/$metadata#Edm.String",
"value": "Not found person with FirstName = Goods and LastName = Zha%2Fngg"
}
``` 

Get: /People(FirstName='Goods',LastName='Zha/ngg')
It return HTTP 404

**Data Model**
ODataRoutingSample.Models.Person

**EDM (CSDL) Model**

**Request/Response**
Get: /People(FirstName='Goods',LastName='Zha%2Fngg') 
Get: /People(FirstName='Goods',LastName='Zha/ngg')

**Expected behavior**
Expect to Return
```json
{
"@odata.context": "https://localhost:64772/$metadata#People/$entity",
"FirstName": "Goods",
"LastName": "Zha/ngg"
}
```

**Screenshots**
Get: /People(FirstName='Goods',LastName='Zha/ngg')
![image](https://user-images.githubusercontent.com/31735283/163607081-119ce9bd-cb0f-4a52-a5d0-9353025b0efd.png) 

Get: /People(FirstName='Goods',LastName='Zha%2Fngg') 
![image](https://user-images.githubusercontent.com/31735283/163607166-f7a8b9d0-9f6f-470a-ab46-ad7296cd6ce1.png)

**Additional context**
N/A

julealgon:
A raw `/` character in a quote string should not work as per the docs, but your alternative, using the encoded value, _should_. Looks like there should be some decoding process happening prior to attempting to compare the actual values here.

xuzhg:
[...]

PR #572, opened 2022-04-21
Issue #567: Escape character , for example ('/') is not proper handle at key or function in quoted string

xuzhg:
#567

When send a request contains the escaped string in the string key or function parameter as follows
`Get: /People(FirstName='Goods',LastName='Zha%2Fngg')`

We should unescape it before going to the parsing step.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Routing/Template/SegmentTemplateHelpers.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/Template/SegmentTemplateHelpers.cs
@@ -58,6 +58,7 @@ namespace Microsoft.AspNetCore.OData.Routing.Template
                 {
                     string strValue = rawValue as string;
                     string newStrValue = context.GetParameterAliasOrSelf(strValue);
+                    newStrValue = Uri.UnescapeDataString(newStrValue);
                     if (newStrValue != strValue)
                     {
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.OData/Routing/Template/SegmentTemplateHelpers.cs
```diff
@@ -58,6 +58,7 @@ internal static class SegmentTemplateHelpers
                 {
                     string strValue = rawValue as string;
                     string newStrValue = context.GetParameterAliasOrSelf(strValue);
+                    newStrValue = Uri.UnescapeDataString(newStrValue);
```
corranrogue9, line 61: Is it possible with something like an open type for a property to be named with a character that needs to be escaped/unescaped? I'm really wondering if this new unescape logic should be applied to all segments, not just key and function segments

corranrogue9 (approved):

src/Microsoft.AspNetCore.OData/Routing/Template/SegmentTemplateHelpers.cs
xuzhg: It's not possible because the property name is an identifier, and an identifier should be a simple string.

PR #574, opened 2022-04-22
Issue #349: Per-request query options are not preserved with the new $batch execution

xuzhg:
#349

Customers may retrieve the query from the main request ($batch) that will create IQueryFeature into the Main HttpContext, Once created, it will cache in the main HttpContext. We should remove IQueryFeature and don't pass that to the subrequest. Otherwise, the sub-request query string will never be used.


xuzhg:
@[garegina](https://github.com/garegina)

garegina (approved):

Issue #568, opened 2022-04-16
Error when function/action name contains keyword "On"


appvity-danhpham:
**Assemblies affected**
.NET 6.0
Microsoft.AspNetCore.OData 8.0.*

**Describe the bug**
Error when function/action name contains keyword "On"
Ex: 
/odata/SEChat/SE.GetStatusOnLineOfflineUser => error
/odata/SEChat/SE.StatusLineOfflineUserOn => error
/odata/SEChat/SE.OnGetStatusLineOfflineUser => success
/odata/SEChat/SE.GetStatusLineOfflineUser => success
/odata/SEChat/SE.GetStatusonLineOfflineUser => success

**EDM (CSDL) Model**
```C#
var builder = new ODataConventionModelBuilder();
            builder.Namespace = "SE";
            builder.ContainerName = "SE";
            builder.EnableLowerCamelCase();
builder.EntitySet<ChatInfo>("SEChat");
builder.EntityType<ChatInfo>().Collection.Function("GetStatusOnLineOfflineUser").ReturnsCollection<spv2_Org_GetOnlineOfflineUser_Result>().Parameter<int>("userId");
return builder.GetEdmModel();
```
**Program.cs**
```C#
builder.Services.AddControllers().AddOData(opt => {
    opt.Conventions.Remove(opt.Conventions.OfType<MetadataRoutingConvention>().First());
    opt.AddRouteComponents("odata", EdmModel.GetEdmModel())
        .Count()
        .OrderBy()
        .Filter()
        .Select()
        .Expand()
        .SkipToken()
        .SetMaxTop(300);
    opt.RouteOptions.EnableNonParenthesisForEmptyParameterFunction = true;
});
```
**Controller**
```C#
        [HttpGet]
        [EnableQuery]
        public IActionResult GetStatusOnLineOfflineUser(int userId)
        {
            //Code
            return Ok();
        }
```
**Request/Response**
```C#
Request: http://localhost:3481/odata/SEChat/SE.GetStatusOnLineOfflineUser(userId=1)
Response: 
Microsoft.OData.ODataException: The key value (SE.GetStatusOnLineOfflineUser(userId=1)) from request is not valid. The key value should be format of type 'Edm.Int32'.
 ---> Microsoft.OData.ODataException: Expected literal type token but found token 'SE'.
   at Microsoft.OData.UriParser.ExpressionLexerLiteralExtensions.ReadLiteralToken(ExpressionLexer expressionLexer)
[...]

PR #576, opened 2022-04-22
issue #568: Error when function/action name contains keyword "On"

xuzhg:
issue #568

Describe the bug
when function/action name contains keyword "On", the operation convention can't correct parse the action name. 
/odata/SEChat/SE.GetStatusOnLineOfflineUser => error
/odata/SEChat/SE.StatusLineOfflineUserOn => error

What is in the PR:
1) Search the operation using the whole action name first, if found, use it.
2) Split the action name, and search again.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Routing/Conventions/OperationRoutingConvention.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/Conventions/OperationRoutingConvention.cs
@@ -177,7 +177,11 @@ namespace Microsoft.AspNetCore.OData.Routing.Conventions
             out IEdmEntityType castTypeFromActionName, out bool isOnCollection)
         {
             // OperationNameOnCollectionOfEntityType
-            string operationName = SplitActionName(actionName, out string cast, out isOnCollection);
+            StringComparison caseComparision = context.Options?.RouteOptions?.EnableActionNameCaseInsensitive == true ?
+                StringComparison.OrdinalIgnoreCase :
+                StringComparison.Ordinal;
+
+            string operationName = SplitActionName(actionName, out string cast, out isOnCollection, caseComparision);
 
             castTypeFromActionName = null;
             if (cast != null)
@@ -206,8 +210,10 @@ namespace Microsoft.AspNetCore.OData.Routing.Conventions
         /// <param name="actionName">The input action name.</param>
         /// <param name="cast">The out of cast type name.</param>
         /// <param name="isOnCollection">The out of collection binding flag.</param>
+        /// <param name="comparison">The case comparision flag.</param>
         /// <returns>The operation name.</returns>
-        internal static string SplitActionName(string actionName, out string cast, out bool isOnCollection)
+        internal static string SplitActionName(string actionName, out string cast, out bool isOnCollection,
+            StringComparison comparison = StringComparison.Ordinal)
         {
             Contract.Assert(actionName != null);
 
@@ -218,7 +224,7 @@ namespace Microsoft.AspNetCore.OData.Routing.Conventions
             cast = null;
             isOnCollection = false;
             string operation;
-            int index = actionName.LastIndexOf("OnCollectionOf", StringComparison.Ordinal);
+            int index = actionName.LastIndexOf("OnCollectionOf", comparison);
             if (index > 0)
             {
                 operation = actionName.Substring(0, index);
@@ -227,7 +233,7 @@ namespace Microsoft.AspNetCore.OData.Routing.Conventions
                 return operation;
             }
 
-            index = actionName.LastIndexOf("On", StringComparison.Ordinal);
+            index = actionName.LastIndexOf("On", comparison);
             if (index > 0)
             {
                 operation = actionName.Substring(0, index);
--- a/src/Microsoft.AspNetCore.OData/Routing/Conventions/OperationRoutingConvention.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/Conventions/OperationRoutingConvention.cs
@@ -76,11 +76,6 @@ namespace Microsoft.AspNetCore.OData.Routing.Conventions
                 candidates = FindCandidates(context, entityType, actionName, out castTypeFromActionName, out isOnCollection);
             }
 
-            if (candidates == null || candidates.Length == 0)
-            {
-                return;
-            }
-
             foreach (IEdmOperation edmOperation in candidates)
             {
                 IEdmOperationParameter bindingParameter = edmOperation.Parameters.FirstOrDefault();
@@ -192,13 +187,13 @@ namespace Microsoft.AspNetCore.OData.Routing.Conventions
                     // Early return for the following cases:
                     // - {OperationName}On
                     // - {OperationName}OnCollectionOf
-                    return null;
+                    return Array.Empty<IEdmOperation>();
                 }
 
                 castTypeFromActionName = entityType.FindTypeInInheritance(context.Model, cast, context.Options?.RouteOptions?.EnableActionNameCaseInsensitive == true) as IEdmEntityType;
                 if (castTypeFromActionName == null)
                 {
-                    return null;
+                    return Array.Empty<IEdmOperation>();
                 }
             }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.OData/Routing/Conventions/OperationRoutingConvention.cs
```diff
@@ -190,7 +207,7 @@ internal static string SplitActionName(string actionName, out string cast, out b
             cast = null;
             isOnCollection = false;
             string operation;
-            int index = actionName.IndexOf("OnCollectionOf", StringComparison.Ordinal);
+            int index = actionName.LastIndexOf("OnCollectionOf", StringComparison.Ordinal);
```
corranrogue9, line 210: Do these comparisons also need to be updated to honor `context.Options?.RouteOptions?.EnableActionNameCaseInsensitive`?

src/Microsoft.AspNetCore.OData/Routing/Conventions/OperationRoutingConvention.cs
```diff
@@ -65,31 +65,39 @@ protected void ProcessOperations(ODataControllerActionContext context, IEdmEntit
...
-                    // - {OperationName}OnCollectionOf
-                    return;
-                }
+                // OperationNameOnCollectionOfEntityType
+                string operationName = SplitActionName(actionName, out string cast, out isOnCollection);
 
-                castTypeFromActionName = entityType.FindTypeInInheritance(context.Model, cast, context.Options?.RouteOptions?.EnableActionNameCaseInsensitive == true) as IEdmEntityType;
-                if (castTypeFromActionName == null)
+                castTypeFromActionName = null;
+                if (cast != null)
```
corranrogue9, line 81: Do I understand correctly that these early `return`s are an optimization and not actually involved with addressing the bug? In other words, wouldn't it make sense to factor the code this way:

```
string operationName = SplitActionName(actionName, out string cast, out isOnCollection);
castTypeFromActionName = null;
if (cast != null)
{
  if (cast.Length == 0)
  {
    return;
  }

  castTypeFromActionName = entityType.FindTypeInInheritance(context.Model, cast, context.Options?.RouteOptions?.EnableActionNameCaseInsensitive == true) as IEdmEntityType;
  if (castTypeFromActionName == null)
  {
    return;
  }
}

IEdmOperation[] candidates = FindCandidates(context, actionName);
if (candidates.Length == 0)
{
  candidates = FindCandidates(context, operationName);
}
```

src/Microsoft.AspNetCore.OData/Routing/Conventions/OperationRoutingConvention.cs
```diff
@@ -172,6 +180,15 @@ protected void ProcessOperations(ODataControllerActionContext context, IEdmEntit
             }
         }
 
+        private static IEdmOperation[] FindCandidates(ODataControllerActionContext context, string operationName)
+        {
+            // TODO: refactor here
+            // If we have multiple same function defined, we should match the best one?
+
+            StringComparison actionNameComparison = context.Options?.RouteOptions?.EnableActionNameCaseInsensitive == true ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal;
+            return context.Model.SchemaElements.OfType<IEdmOperation>().Where(f => f.IsBound && f.Name.Equals(operationName, actionNameComparison)).ToArray();
```
corranrogue9, line 189: Why `.ToArray()`? The code was already running without it, and I don't see how adding it improves anything

xuzhg: we will do use it:
1) check the count
2) iterator
So, it could be an improvement to call toArray().

src/Microsoft.AspNetCore.OData/Routing/Conventions/OperationRoutingConvention.cs
julealgon: > Do I understand correctly that these early returns are an optimization and not actually involved with addressing the bug?

If that's really the case, it would be nice to avoid making this change in the same PR IMHO. Or at the very least, the change should be done in a separate commit in the branch.

src/Microsoft.AspNetCore.OData/Routing/Conventions/OperationRoutingConvention.cs
```diff
@@ -65,31 +65,39 @@ protected void ProcessOperations(ODataControllerActionContext context, IEdmEntit
...
 
-            // OperationNameOnCollectionOfEntityType
-            string operationName = SplitActionName(actionName, out string cast, out bool isOnCollection);
-
+            bool isOnCollection = false;
             IEdmEntityType castTypeFromActionName = null;
-            if (cast != null)
+
+            // Let's find the operations using the action first,
+            // If not founding, let's split the action name, search again
```
julealgon, line 72: These comments imply to me we have 2 strategies, and a fallback/priotity between them.

It would be nice to model these as 2 distinct functions and then have an explicit fallback using `??` or even a small chain of responsibility implementation (though that might be a bit overkill here).

src/Microsoft.AspNetCore.OData/Routing/Conventions/OperationRoutingConvention.cs
julealgon, line 189: Long lines like this are really hard to read.

I'd suggest reformatting like:
```suggestion
            return context.Model.SchemaElements
                .OfType<IEdmOperation>()
                .Where(f => f.IsBound && f.Name.Equals(operationName, actionNameComparison))
                .ToArray();
```

src/Microsoft.AspNetCore.OData/Routing/Conventions/OperationRoutingConvention.cs
xuzhg: Refactor codes, please take a look. Thanks.

src/Microsoft.AspNetCore.OData/Routing/Conventions/OperationRoutingConvention.cs
xuzhg: #Resolved.

xuzhg: #Resolved.

src/Microsoft.AspNetCore.OData/Routing/Conventions/OperationRoutingConvention.cs
```diff
@@ -65,31 +65,22 @@ protected void ProcessOperations(ODataControllerActionContext context, IEdmEntit
...
+                // Let's split the action name and use part of it to search again.
+                candidates = FindCandidates(context, entityType, actionName, out castTypeFromActionName, out isOnCollection);
+            }
 
-                castTypeFromActionName = entityType.FindTypeInInheritance(context.Model, cast, context.Options?.RouteOptions?.EnableActionNameCaseInsensitive == true) as IEdmEntityType;
-                if (castTypeFromActionName == null)
-                {
-                    return;
-                }
+            if (candidates == null || candidates.Length == 0)
```
julealgon, line 79: Can we make it so that none of the `FindCandidates` overloads ever return `null`, and instead make it return an empty array (for instance, using `Array.Empty<T>`)?

Handling null collections like this is usually ill-advised unless there is a very clear semantic value difference between null and empty (which I don't think is the case here).

We could then simplify this back to:
```suggestion
            if (candidates.Length == 0)
```

xuzhg: #Resolved.

corranrogue9 (dismissed):

src/Microsoft.AspNetCore.OData/Routing/Conventions/OperationRoutingConvention.cs
xuzhg: #Resolved.

corranrogue9 (approved):

PR #579, opened 2022-04-27
Support FromBody parameters with actions and functions

### LINKED ISSUE #571: Cannot use either `[FromBody] Foo entity` or `[FromBody] Delta<Foo> entity` with `[HttpPost]` method
**Assemblies affected**
`Microsoft.AspNetCore.OData` version 8.0.8

**Describe the bug**
In our project, we have an endpoint (OData bound Function) is responsible for two operations in one place:
- To **update** (an existing entity) [PATCH] 
- And **clone** to new one [POST]

From the above point, we implemented that endpoint & still working on `Microsoft.AspNetCore.OData` v7.4.1 until migrating to v8.0.8. 
We cannot use that way anymore because of `NullReferenceException` on `DeltaOfT` parameter.

**Data Model**
- I will reproduce an entity class & controller from scratch.

**Product.cs**
```
public enum ProductType
{
    Old = 1,
    LikeNew,
    NewSeal
}

public class Product
{
    public string Id { get; set; }

    public ProductType ProductType { get; set; }
}
```

**ProductConfigurationBuilder.cs**
```
public class ProductConfiguration : IEdmConfigurationV1
{
    public ProductConfiguration()
    {
    }

    public void Configure(ODataConventionModelBuilder builder)
    {
        var config = builder.EntitySet<Product>(nameof(Product));
        var entity = config.EntityType;

        entity.HasKey(p => p.Id);
        entity.Property(p => p.Name);
        entity.EnumProperty(p => p.ProductType).Select(SelectExpandType.Allowed);

        entity
            .Select() // Product?$select
            .Filter() // Product?$filter
            .Count() // Product/$count
            .Page(); // Product?$top= & $skip

        var importAction = entity.Action("Import");
        importAction.ReturnsFromEntitySet<Product>(nameof(Product));

        var moveProductToOtherShopAction = entity.Action("MoveProductToOtherShop");
        moveProductToOtherShopAction .ReturnsFromEntitySet<Product>(nameof(Product));
        
    }
}
```

**ProductController.cs**
```
using Microsoft.AspNetCore.Http;
using Microsoft.AspNetCore.OData.Deltas;
using Microsoft.AspNetCore.OData.Formatter;
using Microsoft.AspNetCore.OData.Query;
using Microsoft.AspNetCore.OData.Routing.Attributes;
using SampleOData.V8.Entity.Sales;

namespace SampleOData.V8.OData.Controllers.V1;

public class ProductController : ODataController
{
    public ProductController()
    {
    }

    // ... other endpoints

    [HttpPost]
    public async Task<IActionResult> Import([FromODataUri] string key,
            [FromBody] Delta<Product> value, ODataQueryOptions<Product> options)
    {
        value.TryGetPropertyValue(nameof(Product.ProductType), out object type);

        if ((ProductType)type == ProductType.LikeNew)
        {
            return UnprocessableEntity("Stock is full");
        }

        if (!ModelState.IsValid)
        {
            return BadRequest(ModelState);
        }

        if (options.IfMatch == null)
        {
            return StatusCode(StatusCodes.Status428PreconditionRequired);
        }

        var data = await Task.FromResult(new Product());
        return Ok(data);
    }

    [HttpPost]
    public async Task<IActionResult> MoveProductToOtherShop([FromODataUri] string key,
            [FromBody] Product product, ODataQueryOptions<Product> options)
    {
        if (product.ProductType == ProductType.LikeNew)
        {
            return UnprocessableEntity("Out of stock");
        }

        if (!ModelState.IsValid)
        {
            return BadRequest(ModelState);
        }

        if (options.IfMatch == null)
        {
            return StatusCode(StatusCodes.Status428PreconditionRequired);
        }

        var data = await Task.FromResult(new Product());
        return Ok(data);
    }
}
```

**Request/Response**
POSTMAN 
- Use one of curl below with `--header` & `--data-raw` argument in the next following line

```
curl --location --request POST 'http://localhost:12000/api/v1.0/Product('\''6256754c161cb552ecd0b878'\'')/Import' \
curl --location --request POST 'http://localhost:12000/api/v1.0/Product('\''6256754c161cb552ecd0b878'\'')/MoveProductToOtherShop' \


--header 'If-Match: W/"JzYzNzg1NDMwMDkyMjYxJw=="' \
--header 'Content-Type: application/json' \
--data-raw '{
    "id": "6256754c161cb552ecd0b878",
    "name": "My Table",
    "productType": 2
}'
```

**Expected behavior**
Should work as you guys did on v7.4.1

**Screenshots**
image

**Stack trace**
image

gathogojr:
Fixes #571 

The target navigation source is unresolved at the time of initializing an `ActionSegmentTemplate` and a `FunctionSegmentTemplate`. For that reason, the `NavigationSource` property of the `ODataSegmentTemplate` is initialized to `null`. Subsequently when we create an `OperationSegment` from the `TryTranslate` method of either of the two templates, the `EdmType` and `EntitySet` properties of the segment are initialized to null. This is the reason for the error exhibited in the [reported issue](https://github.com/OData/AspNetCoreOData/issues/571).

We fix the issue in this PR by using the entity set annotation to get the target navigation source. This annotation is set when `ReturnsFromEntitySet` method (and similar variants) are called when using the model builder to initialize the Edm model.
By fixing this we dealt with a technical debt in the form of a comment left in [`ODataOutputFormatterHelper`](https://github.com/OData/AspNetCoreOData/blob/eea3a95a4831432d08fcdd5beaea0ec791d642ce/src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatterHelper.cs#L211)

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatterHelper.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatterHelper.cs
@@ -66,7 +66,7 @@ namespace Microsoft.AspNetCore.OData.Formatter
             IODataSerializer serializer = GetSerializer(type, value, request, serializerProvider);
 
             ODataPath path = request.ODataFeature().Path;
-            IEdmNavigationSource targetNavigationSource = GetTargetNavigationSource(path, model);
+            IEdmNavigationSource targetNavigationSource = path.GetNavigationSource();
             HttpResponse response = request.HttpContext.Response;
 
             // serialize a response
@@ -194,30 +194,6 @@ namespace Microsoft.AspNetCore.OData.Formatter
             }
 
             return serializer;
-        }
-
-        private static IEdmNavigationSource GetTargetNavigationSource(ODataPath path, IEdmModel model)
-        {
-            if (path == null)
-            {
-                return null;
-            }
-
-            Contract.Assert(model != null);
-
-            OperationSegment operationSegment = path.LastSegment as OperationSegment;
-            if (operationSegment != null)
-            {
-                // OData model builder uses an annotation to save the function returned entity set.
-                // TODO: we need to refactor it later.
-                ReturnedEntitySetAnnotation entitySetAnnotation = model.GetAnnotationValue<ReturnedEntitySetAnnotation>(operationSegment.Operations.Single());
-                if (entitySetAnnotation != null)
-                {
-                    return model.EntityContainer.FindEntitySet(entitySetAnnotation.EntitySetName);
-                }
-            }
-
-            return path.GetNavigationSource();
         }
 
         private static string GetRootElementName(ODataPath path)
--- a/src/Microsoft.AspNetCore.OData/Routing/Template/ActionSegmentTemplate.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/Template/ActionSegmentTemplate.cs
@@ -10,6 +10,7 @@ using System.Diagnostics.Contracts;
 using System.Linq;
 using Microsoft.OData;
 using Microsoft.OData.Edm;
+using Microsoft.OData.ModelBuilder;
 using Microsoft.OData.UriParser;
 
 namespace Microsoft.AspNetCore.OData.Routing.Template
@@ -104,7 +105,15 @@ namespace Microsoft.AspNetCore.OData.Routing.Template
                 throw Error.ArgumentNull(nameof(context));
             }
 
-            context.Segments.Add(Segment);
+            if (NavigationSource != null)
+            {
+                context.Segments.Add(Segment);
+                return true;
+            }
+
+            IEdmNavigationSource navigationSource = SegmentTemplateHelpers.GetNavigationSourceFromEdmOperation(context.Model, Action);
+            OperationSegment actionSegment = new OperationSegment(Action, navigationSource as IEdmEntitySetBase);
+            context.Segments.Add(actionSegment);
             return true;
         }
     }
--- a/src/Microsoft.AspNetCore.OData/Routing/Template/FunctionSegmentTemplate.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/Template/FunctionSegmentTemplate.cs
@@ -11,6 +11,7 @@ using System.Linq;
 using Microsoft.AspNetCore.OData.Edm;
 using Microsoft.OData;
 using Microsoft.OData.Edm;
+using Microsoft.OData.ModelBuilder;
 using Microsoft.OData.UriParser;
 
 namespace Microsoft.AspNetCore.OData.Routing.Template
@@ -144,10 +145,16 @@ namespace Microsoft.AspNetCore.OData.Routing.Template
                 throw Error.ArgumentNull(nameof(context));
             }
 
+            IEdmNavigationSource navigationSource = NavigationSource;
+            if (navigationSource == null)
+            {
+                navigationSource = SegmentTemplateHelpers.GetNavigationSourceFromEdmOperation(context.Model, Function);
+            }
+
             // If the function has no parameter, we don't need to do anything and just return an operation segment.
             if (ParameterMappings.Count == 0)
             {
-                context.Segments.Add(new OperationSegment(Function, NavigationSource as IEdmEntitySetBase));
+                context.Segments.Add(new OperationSegment(Function, navigationSource as IEdmEntitySetBase));
                 return true;
             }
 
@@ -172,7 +179,7 @@ namespace Microsoft.AspNetCore.OData.Routing.Template
                 return false;
             }
 
-            context.Segments.Add(new OperationSegment(Function, parameters, NavigationSource as IEdmEntitySetBase));
+            context.Segments.Add(new OperationSegment(Function, parameters, navigationSource as IEdmEntitySetBase));
             return true;
         }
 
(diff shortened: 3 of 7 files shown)


ElizabethOkerio:
@gathogojr  please add description of the changes you've made.

src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatterHelper.cs
```diff
@@ -66,7 +66,7 @@ public static ODataSerializerContext BuildSerializerContext(HttpRequest request)
             IODataSerializer serializer = GetSerializer(type, value, request, serializerProvider);
 
             ODataPath path = request.ODataFeature().Path;
-            IEdmNavigationSource targetNavigationSource = GetTargetNavigationSource(path, model);
+            IEdmNavigationSource targetNavigationSource = path.GetNavigationSource();
```
ElizabethOkerio, line 69: I don't think I understand why you did away with the <code>GetTargetNavigationSource</code> method. and called the <code>path.GetNavigationSource</code> directly. Does it mean the <code>path.LastSegment</code> is always null at this point?

src/Microsoft.AspNetCore.OData/Routing/Template/ActionSegmentTemplate.cs
```diff
@@ -104,7 +105,21 @@ public override bool TryTranslate(ODataTemplateTranslateContext context)
...
             }
 
-            context.Segments.Add(Segment);
+            if (NavigationSource != null)
+            {
+                context.Segments.Add(Segment);
+                return true;
+            }
+
+            IEdmNavigationSource navigationSource = null;
```
ElizabethOkerio, line 114: Can we just use the <code>NavigationSource </code> property here without introducing a new variable?

src/Microsoft.AspNetCore.OData/Routing/Template/FunctionSegmentTemplate.cs
```diff
@@ -144,10 +145,20 @@ public override bool TryTranslate(ODataTemplateTranslateContext context)
                 throw Error.ArgumentNull(nameof(context));
             }
 
+            IEdmNavigationSource navigationSource = NavigationSource;
```
ElizabethOkerio, line 148: Also here. Can we use the NavigationSource property here without introducing a new variable? shouldn't the <code>navigationSource</code> be same as <code>NavigationSource</code> throughtout the class?

src/Microsoft.AspNetCore.OData/Routing/Template/ActionSegmentTemplate.cs
```diff
@@ -104,7 +105,21 @@ public override bool TryTranslate(ODataTemplateTranslateContext context)
...
+            ReturnedEntitySetAnnotation entitySetAnnotation = context.Model?.GetAnnotationValue<ReturnedEntitySetAnnotation>(Action);
+            if (entitySetAnnotation != null)
+            {
+                navigationSource = context.Model.EntityContainer.FindEntitySet(entitySetAnnotation.EntitySetName);
+            }
+
+            var actionSegment = new OperationSegment(Action, navigationSource as IEdmEntitySetBase);
```
ElizabethOkerio, line 121: nit: use of var.

src/Microsoft.AspNetCore.OData/Routing/Template/FunctionSegmentTemplate.cs
```diff
@@ -144,10 +145,20 @@ public override bool TryTranslate(ODataTemplateTranslateContext context)
...
+            if (navigationSource == null)
+            {
+                ReturnedEntitySetAnnotation entitySetAnnotation = context.Model?.GetAnnotationValue<ReturnedEntitySetAnnotation>(Function);
+                if (entitySetAnnotation != null)
+                {
+                    navigationSource = context.Model.EntityContainer.FindEntitySet(entitySetAnnotation.EntitySetName);
+                }
```
ElizabethOkerio, line 155: There is repetition of this code block in the <code>FunctionSegmentTemplate </code> and in the <code>ActionSegmentTemplate</code>. Can it be shared or something?

src/Microsoft.AspNetCore.OData/Routing/Template/ActionSegmentTemplate.cs
gathogojr: The `NavigationSource` property (declared in `ODataSegmentTemplate`) is readonly and cannot be assigned from `TryTranslate`

src/Microsoft.AspNetCore.OData/Routing/Template/FunctionSegmentTemplate.cs
gathogojr: The `NavigationSource` property (declared in `ODataSegmentTemplate`) is readonly and cannot be assigned from `TryTranslate`

src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatterHelper.cs
gathogojr: The `GetTargetNavigationSource` method defined in this class was a workaround to the problem that we're fixing from the `TryTranslate` methods of `ActionSegmentTemplate` and `FunctionSegmentTemplate`. It's not necessary to return it after the fix that we applied

ElizabethOkerio: aah I see

ElizabethOkerio (approved):

Issue #587, opened 2022-05-03
`ArgumentNullException` when calling `AddODataNewtonsoftJson()`


prochnowc:
**Assemblies affected**
Microsoft.AspNetCore.OData.NewtonsoftJson v8.0.4

**Describe the bug**
The method `AddODataNewtonsoftJson()` calls `AddNewtonsoftJson(null)` 
Instead it should probably call `AddODataNewtonsoftJson(null)`

**Expected behavior**
The method should not throw.

xuzhg:
https://github.com/OData/AspNetCoreOData/blob/main/src/Microsoft.AspNetCore.OData.NewtonsoftJson/ODataNewtonsoftJsonMvcBuilderExtensions.cs#L58 could be a problem.

prochnowc:
@xuzhg Thank you.

PR #589, opened 2022-05-03
Fix issue #587: AddODataNewtonsoftJson should call AddODataNewtonsoftJson with null

xuzhg:
#587

Fix the issue by calling AddODataNewtonsoftJson with null.

Add the test cases.


corranrogue9:
It is normal for there to be test failures but for the build to succeed?

xuzhg:
> 


Abnormal. Where do you see 'test failures but the build succeeded"?

Issue #620, opened 2022-06-15
Empty $select/$expand throws ArgumentNullException


giulianob:
**Assemblies affected**
ASP.NET Core OData 8.0.10

**Describe the bug**
Passing an empty or whitespace $select/$expand will throw an ArgumentNullException

**Reproduce steps**
1. Run `ODataCustomizedSample`
2. `GET http://localhost:5000/convention/Employees?$select=`

```
HTTP/1.1 500 Internal Server Error
Connection: close
Date: Wed, 15 Jun 2022 17:08:52 GMT
Content-Type: text/plain
Server: Kestrel
Transfer-Encoding: chunked

System.ArgumentException: 'select' and 'expand' cannot be both null or empty.
   at Microsoft.AspNetCore.OData.Query.SelectExpandQueryOption..ctor(String select, String expand, ODataQueryContext context, ODataQueryOptionParser queryOptionParser) in AspNetCoreOData\src\Microsoft.AspNetCore.OData\Query\Query\SelectExpandQueryOption.cs:line 53
   at Microsoft.AspNetCore.OData.Query.ODataQueryOptions.BuildQueryOptions(IDictionary`2 queryParameters) in AspNetCoreOData\src\Microsoft.AspNetCore.OData\Query\ODataQueryOptions.cs:line 1022
   at Microsoft.AspNetCore.OData.Query.ODataQueryOptions.Initialize(ODataQueryContext context) in AspNetCoreOData\src\Microsoft.AspNetCore.OData\Query\ODataQueryOptions.cs:line 1138
   at Microsoft.AspNetCore.OData.Query.ODataQueryOptions..ctor(ODataQueryContext context, HttpRequest request) in AspNetCoreOData\src\Microsoft.AspNetCore.OData\Query\ODataQueryOptions.cs:line 85
   at Microsoft.AspNetCore.OData.Query.EnableQueryAttribute.OnActionExecuting(ActionExecutingContext actionExecutingContext) in AspNetCoreOData\src\Microsoft.AspNetCore.OData\Query\EnableQueryAttribute.cs:line 163
```

**Expected behavior**
$select and $expand should be ignored if they are empty.

PR #621, opened 2022-06-15
Fixing ArgumentNullException on empty select/expand

giulianob:
Fixes https://github.com/OData/AspNetCoreOData/issues/620

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/Query/SelectExpandQueryOption.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Query/SelectExpandQueryOption.cs
@@ -48,7 +48,7 @@ namespace Microsoft.AspNetCore.OData.Query
                 throw Error.ArgumentNull(nameof(context));
             }
 
-            if (string.IsNullOrEmpty(select) && string.IsNullOrEmpty(expand))
+            if (string.IsNullOrWhiteSpace(select) && string.IsNullOrWhiteSpace(expand))
             {
                 throw Error.Argument(SRResources.SelectExpandEmptyOrNull);
             }
@@ -60,7 +60,7 @@ namespace Microsoft.AspNetCore.OData.Query
 
             if (!(context.ElementType is IEdmStructuredType))
             {
-                throw Error.Argument(SRResources.SelectNonStructured, context.ElementType);
+                throw Error.Argument("context", SRResources.SelectNonStructured, context.ElementType.ToTraceString());
             }
 
             Context = context;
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNetCore.OData/Query/Query/SelectExpandQueryOption.cs
```diff
@@ -82,33 +82,19 @@ public class SelectExpandQueryOption
 
         // This constructor is intended for unit testing only.
         internal SelectExpandQueryOption(string select, string expand, ODataQueryContext context)
```
giulianob, line 84: Combined the testing and production constructors. The production constructor was never tested and the logic was not exactly the same.

corranrogue9:
I don't think the URLs that this allows are "legal" according to the [ABNF](http://docs.oasis-open.org/odata/odata/v4.01/cs01/abnf/odata-abnf-construction-rules.txt) (search for "$select"). I could be misreading it, or maybe something else you're aware of in the standard overrides it?

giulianob:
What should the behavior be? Return a 400? An internal server error should still be a bug IMO

corranrogue9:
Good point Giuliano, we triaged this today. `$select=` is *not* valid, so it should still be an error, but we agree with you that it should be `4xx` instead of `5xx`. From what I see in your tests, this PR makes it a `2xx`? Would you be willing to make the change to change this to `4xx`?

giulianob:
@corranrogue9 Fixed

image

src/Microsoft.AspNetCore.OData/Query/Validator/ODataQueryValidator.cs
```diff
@@ -139,5 +141,13 @@ private static void ValidateQueryOptionAllowed(AllowedQueryOptions queryOption,
                 throw new ODataException(Error.Format(SRResources.NotAllowedQueryOption, queryOption, "AllowedQueryOptions"));
             }
         }
+        
+        private static void ValidateNotEmptyOrWhitespace(string rawValue)
+        {
+            if (rawValue != null && string.IsNullOrWhiteSpace(rawValue))
```
corranrogue9, line 147: I'm actually pretty baffled that there's no existing `string.IsWhiteSpace` method. To double check, though, that's what you're trying to do here, right?

Any consideration for removing `rawValue != null` since that's already checked by both callers?

corranrogue9 (approved):

Nthemba (approved):

PR #630, opened 2022-06-30
Check for empty list instead of null value.

VirusQuartirus:


src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatter.cs
```diff
@@ -194,7 +194,8 @@ public override void WriteResponseHeaders(OutputFormatterWriteContext context)
             // Set the character set.
             MediaTypeHeaderValue currentContentType = GetContentType(response.Headers[HeaderNames.ContentType].FirstOrDefault());
             RequestHeaders requestHeader = request.GetTypedHeaders();
-            if (requestHeader != null && requestHeader.AcceptCharset != null)
+            // Starting from ASP .NET Core 3.0 AcceptCharset returns an empty collection instead of null.
+            if (requestHeader?.AcceptCharset.Count > 0)
```
corranrogue9, line 198:
```suggestion
            if (requestHeader?.AcceptCharset?.Count > 0)
```

Let's be compatible with both versions

src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatter.cs
```diff
@@ -194,7 +194,8 @@ public override void WriteResponseHeaders(OutputFormatterWriteContext context)
             // Set the character set.
             MediaTypeHeaderValue currentContentType = GetContentType(response.Headers[HeaderNames.ContentType].FirstOrDefault());
             RequestHeaders requestHeader = request.GetTypedHeaders();
-            if (requestHeader != null && requestHeader.AcceptCharset != null)
+            // Starting from ASP .NET Core 3.0 AcceptCharset returns an empty collection instead of null.
```
corranrogue9, line 197: Can you please add a test for this?

src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatter.cs
VirusQuartirus: AFAIR this repo is for ASP .NET Core 3+. So it can't be ran with ASP .NET Core 2.x. This change make sense for OData/WebApi repo. Correct me if I'm wrong.

test/Microsoft.AspNetCore.OData.E2E.Tests/Enums/EnumsTest.cs
```diff
@@ -178,6 +178,32 @@ public async Task QueryEntitySetCount(string requestUri, int expectedCount)
...
 
+        [Theory]
+        [InlineData("/convention/Employees/$count", true)]
+        [InlineData("/convention/Employees/$count", false)]
+        public async Task QueryEntitySetCountEncoding(string requestUri, bool sendAcceptCharset)
+        {
+            // Arrange
+            await ResetDatasource();
+            HttpClient client = CreateClient();
+			client.DefaultRequestHeaders.AcceptCharset.Clear();
```
gathogojr, line 189:
```suggestion
            HttpClient client = CreateClient();
            client.DefaultRequestHeaders.AcceptCharset.Clear();
```

test/Microsoft.AspNetCore.OData.E2E.Tests/Enums/EnumsTest.cs
```diff
@@ -178,6 +178,32 @@ public async Task QueryEntitySetCount(string requestUri, int expectedCount)
...
+            // Act
+            HttpResponseMessage response = await client.GetAsync(requestUri);
+
+            // Assert
+            Assert.True(response.IsSuccessStatusCode);
+            var blob = await response.Content.ReadAsByteArrayAsync();
+            Assert.Equal("utf-8", response.Content.Headers.ContentType.CharSet, StringComparer.OrdinalIgnoreCase);
+            Assert.False((blob[0] == 0xEF) && (blob[1] == 0xBB) && (blob[2] == 0xBF));
+            var count = Encoding.UTF8.GetString(blob);
+            Assert.Equal(3, int.Parse(count));
```
gathogojr, line 204:
```suggestion
            var count = Encoding.UTF8.GetString(blob);
            Assert.Equal(3, int.Parse(count));
            Assert.False((blob[0] == 0xEF) && (blob[1] == 0xBB) && (blob[2] == 0xBF));
```

I think the assertion for count should come first to mitigate possible `index out of range" exception

VirusQuartirus: Agree

gathogojr (approved):
✅

src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatter.cs
habbes: I'd also lean in favor of the `AcceptCharset?.Count` check

habbes: But the standard lib explicitly marks the properties that can be null as nullable in newer .NET versions e.g.: https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.headers.requestheaders.cachecontrol?view=aspnetcore-6.0

I don't think that was the case for .NET Core 3+ so it's hard to tell from just looking at the property whether it's safe to assume it can't be null.

habbes (approved):

odero (approved):

PR #643, opened 2022-07-22
Resolve ContentID in odata.bind annotation

gathogojr:
### Issues

*This pull request fixes https://github.com/OData/odata.net/issues/2442.*

### Description

Resolve ContentID in odata.bind annotation

- [ ] *Test cases added*
- [ ] *Build and test with one-click build and test script passed*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
@@ -14,7 +14,9 @@ using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.Serialization;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.OData.Deltas;
 using Microsoft.AspNetCore.OData.Edm;
 using Microsoft.AspNetCore.OData.Extensions;
@@ -22,6 +24,7 @@ using Microsoft.AspNetCore.OData.Formatter.Value;
 using Microsoft.AspNetCore.OData.Formatter.Wrapper;
 using Microsoft.AspNetCore.OData.Routing;
 using Microsoft.AspNetCore.OData.Routing.Parser;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.OData;
 using Microsoft.OData.Edm;
 using Microsoft.OData.UriParser;
@@ -33,6 +36,8 @@ namespace Microsoft.AspNetCore.OData.Formatter.Deserialization
     /// </summary>
     public class ODataResourceDeserializer : ODataEdmTypeDeserializer
     {
+        private static readonly Regex ContentIdReferencePattern = new Regex(@"\$\d", RegexOptions.Compiled);
+
         /// <summary>
         /// Initializes a new instance of the <see cref="ODataResourceDeserializer"/> class.
         /// </summary>
@@ -376,7 +381,7 @@ namespace Microsoft.AspNetCore.OData.Formatter.Deserialization
             }
 
             IList<ODataItemWrapper> nestedItems;
-            var referenceLinks = resourceInfoWrapper.NestedItems.OfType<ODataEntityReferenceLinkWrapper>().ToArray();
+            ODataEntityReferenceLinkWrapper[] referenceLinks = resourceInfoWrapper.NestedItems.OfType<ODataEntityReferenceLinkWrapper>().ToArray();
             if (referenceLinks.Length > 0)
             {
                 // Be noted:
@@ -578,7 +583,7 @@ namespace Microsoft.AspNetCore.OData.Formatter.Deserialization
 
             IEdmStructuredTypeReference structuredType = edmType.AsStructured();
 
-            var nestedReadContext = new ODataDeserializerContext
+            ODataDeserializerContext nestedReadContext = new ODataDeserializerContext
             {
                 Path = readContext.Path,
                 Model = readContext.Model,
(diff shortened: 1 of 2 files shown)


ElizabethOkerio:
@gathogojr is this valid `"Parent@odata.bind": "$1"`. Should the writer write something like this?

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -870,26 +875,35 @@ private static IList<ODataProperty> CreateKeyProperties(Uri id, ODataDeserialize
...
+                HttpRequest request = readContext.Request;
+                IServiceProvider requestContainer = request.GetRouteServices();
+                Uri resolvedId = id;
+
+                string idOriginalString = id.OriginalString;
+                if (ContentIdReferencePattern.IsMatch(idOriginalString))
                 {
-                    string serviceRoot = readContext.Request.CreateODataLink();
-                    serviceRootUri = new Uri(serviceRoot, UriKind.Absolute);
+                    // We can expect request.ODataBatchFeature() to not be null?
```
habbes, line 886: This seems like a question. Does that mean it's not guaranteed not to be null? Maybe you should confirm to avoid possible null-reference exceptions.

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -870,26 +875,35 @@ private static IList<ODataProperty> CreateKeyProperties(Uri id, ODataDeserialize
...
                 }
 
-                var request = readContext.Request;
-                IEdmModel model = readContext.Model;
-
-                // TODO: shall we use the DI to inject the path parser?
-                DefaultODataPathParser pathParser = new DefaultODataPathParser();
+                Uri serviceRootUri = new Uri(request.CreateODataLink());
+                IODataPathParser pathParser = requestContainer?.GetService<IODataPathParser>();
+                if (pathParser == null) // Seem like IODataPathParser is NOT injected into DI container by default
```
habbes, line 895:
```suggestion
                if (pathParser == null) // Seems like IODataPathParser is NOT injected into DI container by default
```

test/Microsoft.AspNetCore.OData.E2E.Tests/Batch/ContentIdToLocationMappingTests.cs
```diff
@@ -0,0 +1,200 @@
...
+        {
+            return Created(new Uri($"{Request.Scheme}://{Request.Host}{Request.Path}/{parent.ParentId}"), parent);
+        }
+    }
+
+    public class ContentIdToLocationMappingChildrenController : ODataController
+    {
+        public ActionResult Post([FromBody] ContentIdToLocationMappingChild child)
+        {
+            Assert.Equal(123, child.Parent.ParentId);
```
habbes, line 157: This is where we assert that the id was bound correctly?

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -870,26 +875,35 @@ private static IList<ODataProperty> CreateKeyProperties(Uri id, ODataDeserialize
...
             {
-                Uri serviceRootUri = null;
-                if (id.IsAbsoluteUri)
+                IEdmModel model = readContext.Model;
+                HttpRequest request = readContext.Request;
+                IServiceProvider requestContainer = request.GetRouteServices();
+                Uri resolvedId = id;
+
+                string idOriginalString = id.OriginalString;
+                if (ContentIdReferencePattern.IsMatch(idOriginalString))
```
ElizabethOkerio, line 884: is this Regex Matching? Is it possible to use `substring` or `indexOf` to do the matching instead of `Regex`. I remember some discussions where this `Regex` was being discouraged.

xuzhg (dismissed):

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
gathogojr: @habbes Based on the the way [`ODataBatchFeature` extension method is implemented here](https://github.com/OData/AspNetCoreOData/blob/69eec03c7003fe12d92cdc619efdc16781683694/src/Microsoft.AspNetCore.OData/Extensions/HttpContextExtensions.cs#L47), it can never be null. I'll remove the question mark

test/Microsoft.AspNetCore.OData.E2E.Tests/Batch/ContentIdToLocationMappingTests.cs
gathogojr: @gathogojr Correct

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
gathogojr: @ElizabethOkerio We use [similar logic in `ODataMessageWrapper` class](https://github.com/OData/AspNetCoreOData/blob/f675f6c6abe2cd10bfa00ce8c4f7dc83b49e1293/src/Microsoft.AspNetCore.OData/Formatter/ODataMessageWrapper.cs#L147). We're looking for a very particular pattern here - a `$` followed by an integer. I think using `Substring` and `Index` would result into more complex logic

gathogojr:
> @gathogojr is this valid `"Parent@odata.bind": "$1"` in a request body?. Should the writer write something like this?

Yes. Consider the sample payload below that the customer provided in the issue. You should be able to establish a parent-child relationship in the manner illustrated
```
{
  "requests": [
    {
      "id": "1",
      "method": "POST",
      "url": "http://localhost:5254/odata/Parents",
      "headers": {
        "OData-Version": "4.0",
        "Content-Type": "application/json;odata.metadata=minimal",
        "Accept": "application/json;odata.metadata=minimal"
      },
      "body": {}
    },
    {
      "id": "2",
      "method": "POST",
      "url": "http://localhost:5254/odata/Children",
      "headers": {
        "OData-Version": "4.0",
        "Content-Type": "application/json;odata.metadata=minimal",
        "Accept": "application/json;odata.metadata=minimal"
      },
      "body": {
        "Parent@odata.bind": "$1"
      }
    }
  ]
}
```

ElizabethOkerio (approved):

ElizabethOkerio (approved):

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
habbes: On an unrelated note, let's avoid substring in favor `Span<char>.Slice()` where possible, to avoid string allocations.

habbes (approved):

Issue #644, opened 2022-07-23
Overloaded properties cannot be expanded


Dean-ZhenYao-Wang:
**Assemblies affected**
Microsoft.AspNetCore.OData 8.0.10

**Describe the bug**
When my model is non-inherited and the attributes are not overloaded, I can get the expanded data normally.
However, when my model changes to inherit from the parent model and overloads an attribute in the parent model, the overloaded attribute of the current model can no longer be expanded.

**Reproduce steps**
The simplest set of steps to reproduce the issue. If possible, reference a commit that demonstrates the issue.

**Data Model**
Please share your Data model, for example, your C# class.

**EDM (CSDL) Model**
`<?xml version="1.0" encoding="utf-8"?>
  <edmx:Edmx Version="4.0" xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx">
    <edmx:DataServices>
      <Schema Namespace="BaseDB" xmlns="http://docs.oasis-open.org/odata/ns/edm">
        <EntityType Name="人员">
          <Key>
            <PropertyRef Name="Key" />
          </Key>
          <Property Name="ID" Type="Edm.String" />
          <Property Name="密码" Type="Edm.String" />
          <Property Name="姓名" Type="Edm.String" />
          <Property Name="性别" Type="BaseDB.Enum.性别Enum" Nullable="false" />
          <Property Name="邮件" Type="Edm.String" />
          <Property Name="手机" Type="Edm.String" />
          <Property Name="Code" Type="Edm.String" />
          <Property Name="头像" Type="Edm.String" />
          <Property Name="年龄" Type="Edm.Int32" />
          <Property Name="生日" Type="Edm.DateTimeOffset" />
          <Property Name="所属的组织列表" Type="Collection(BaseDB.人员_组织)" />
          <Property Name="Key" Type="Edm.Guid" Nullable="false" />
          <Property Name="说明" Type="Edm.String" />
          <Property Name="CreateUser" Type="Edm.Guid" Nullable="false" />
          <Property Name="CreateDate" Type="Edm.DateTimeOffset" Nullable="false" />
          <Property Name="UpdateUser" Type="Edm.Guid" />
          <Property Name="UpdateDate" Type="Edm.DateTimeOffset" />
          <Property Name="Remark" Type="Edm.String" />
          <NavigationProperty Name="所属人员类列表" Type="Collection(BaseDB.人员_人员类)" />
[...]

PR #645, opened 2022-07-23
Fixed an issue where attributes of the same name cannot be expanded when subclasses inherit and hide attributes of the parent class

Dean-ZhenYao-Wang:


Dean-ZhenYao-Wang:
#644

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -231,7 +231,11 @@ public virtual Expression CreatePropertyValueExpression(QueryBinderContext conte
             PropertyInfo propertyInfo = source.Type.GetProperty(propertyName, BindingFlags.DeclaredOnly);
             if (propertyInfo == null)
             {
-                propertyInfo = source.Type.GetProperty(propertyName);
+                /*
+                 * History of code:
+                 * propertyInfo = source.Type.GetProperty(propertyName);
+                */
+                propertyInfo = source.Type.GetProperties().Where(m => m.Name.Equals(propertyName)).FirstOrDefault();
```
xuzhg, line 238: Is it safe to call "FirstOrDefault()" to get the correct property info?

Dean-ZhenYao-Wang: After testing, it is found that this modification is not a problem. When the property you are looking for does not exist, the program produces the same result as if it had not been modified.

Dean-ZhenYao-Wang: > Is it safe to call "FirstOrDefault()" to get the correct property info?

After testing, this modification was found not to be a problem. When the property you are looking for does not exist, the program produces the same result as if it had not been modified.

xuzhg, line 238: Would you please add some comments for your change in the source file?

xuzhg (approved):

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
Dean-ZhenYao-Wang: > Would you please add some comments for your change in the source file?

Code comments have been added to the new submission

xuzhg (approved):

PR #647, opened 2022-07-25
Get value of property with data member consideration

ttruatl:
### Issues
When DataMember is applied in an entity, the `entityType.Key()` returns a DataMember name and the `MemberExpression propertyAccess = Expression.Property(odataItParameter, propertyName);` was crashed because the DataMember name couldn't be found in the actual object.

For more details, I setup `pageSize=100` in my ODataConfig, and  I applied DataMember on the key of an entity
```
        [DataMember(Name = "Id")]
        [Key]
```
When I do the `$expand` on a collection of the entity, the function `entityType.Key()` in the [SelectExpandBinder.cs](https://github.com/OData/AspNetCoreOData/blob/f675f6c6abe2cd10bfa00ce8c4f7dc83b49e1293/src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs#L1194) returns my DataMember name `id`, then it could not be found at the line ` Expression.Property(odataItParameter, propertyName)` because it's not actual a property name of the object.

Relate issue: [#2443](https://github.com/OData/WebApi/issues/2443)

### Description

The change is to find the actual property name from the DataMemeber name if the data member name is existing. Otherwise, use the property name from the input.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/ExpressionHelpers.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/ExpressionHelpers.cs
@@ -9,7 +9,6 @@ using System;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
-using System.Runtime.Serialization;
 using Microsoft.AspNetCore.OData.Common;
 using Microsoft.AspNetCore.OData.Edm;
 using Microsoft.AspNetCore.OData.Query.Container;
--- a/src/Microsoft.AspNetCore.OData/Query/ExpressionHelpers.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/ExpressionHelpers.cs
@@ -281,11 +281,8 @@ namespace Microsoft.AspNetCore.OData.Query
 
         public static LambdaExpression GetPropertyAccessLambda(Type type, string propertyName)
         {
-            var dataMemberProps = type.GetProperties().Where(p => Attribute.IsDefined(p, typeof(DataMemberAttribute)));
-            var actualPropName = dataMemberProps
-                                    .SingleOrDefault(p => ((DataMemberAttribute)Attribute.GetCustomAttribute(p, typeof(DataMemberAttribute))).Name?.ToLower() == propertyName.ToLower())?.Name ?? propertyName;
             ParameterExpression odataItParameter = Expression.Parameter(type, "$it");
-            MemberExpression propertyAccess = Expression.Property(odataItParameter, actualPropName);
+            MemberExpression propertyAccess = Expression.Property(odataItParameter, propertyName);
             return Expression.Lambda(propertyAccess, odataItParameter);
         }
     }
--- a/src/Microsoft.AspNetCore.OData/Query/ExpressionHelpers.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/ExpressionHelpers.cs
@@ -9,6 +9,7 @@ using System;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
+using System.Runtime.Serialization;
 using Microsoft.AspNetCore.OData.Common;
 using Microsoft.AspNetCore.OData.Edm;
 using Microsoft.AspNetCore.OData.Query.Container;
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.OData/Query/ExpressionHelpers.cs
```diff
@@ -280,8 +281,11 @@ public static Expression Default(Type type)
 
         public static LambdaExpression GetPropertyAccessLambda(Type type, string propertyName)
         {
+            var dataMemberProps = type.GetProperties().Where(p => Attribute.IsDefined(p, typeof(DataMemberAttribute)));
+            var actualPropName = dataMemberProps
+                                    .SingleOrDefault(p => ((DataMemberAttribute)Attribute.GetCustomAttribute(p, typeof(DataMemberAttribute))).Name?.ToLower() == propertyName.ToLower())?.Name ?? propertyName;
             ParameterExpression odataItParameter = Expression.Parameter(type, "$it");
-            MemberExpression propertyAccess = Expression.Property(odataItParameter, propertyName);
```
xuzhg, line 284: That's weird, because if you use OData model builder, the DataMember is considered in the Edm model.

So, You don't need to find any attributes. You can call "GetClrPropertyName" on the model to get the actual CLR property name. at: https://github.com/OData/AspNetCoreOData/blob/main/src/Microsoft.AspNetCore.OData/Edm/EdmModelAnnotationExtensions.cs#L166

So, my suggestion is 
1) Leave GetPropertyAccessLambda unchanged, we assume the propertyName coming is correct
2) at the upper layer (where calling this method), we can use the model and the IEdmProperty to get the actual CRL property name.

xuzhg (changes requested):

ttruatl:
@xuzhg have you had a chance to look at the related issue [#2443](https://github.com/OData/WebApi/issues/2443)?

xuzhg:
> @xuzhg have you had a chance to look at the related issue [#2443](https://github.com/OData/WebApi/issues/2443)?

What do you mean? 

My suggestion is to call 'GetClrPropertyName' before calling into this method to get the real property name? Are you using non-Edm scenario?

KenitoInc:
@ttruatl 
I have fixed the code as per @xuzhg suggestion.
Kindly add tests

KenitoInc:

KenitoInc:
@ttruatl 
I have added tests.
KIndly fix the CLA agreement so that we can resolve this PR

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/SelectExpandBinderTest.cs
```diff
@@ -1964,6 +2013,30 @@ public class QueryOrder
...
+        public IList<QueryProductTag> Tags { get; set; }
+    }
+
+    public class QueryProductTag
+    {
+        [Key]
+        public int Id { get; set; }
+
+        public string Name { get; set; }
+    }
```
gathogojr, line 2039: nit:
```suggestion
    }

```

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/SelectExpandBinderTest.cs
```diff
@@ -109,6 +114,49 @@ public void Bind_ReturnsIEdmObject_WithRightEdmType(string select)
...
+            {
+                NavigationSource = context.NavigationSource
+            };
+
+            // Act
+            SelectExpandBinder binder = new SelectExpandBinder();
+            IQueryable queryable = binder.ApplyBind(products, selectExpand.SelectExpandClause, queryBinderContext);
+
+            // Assert
+            Assert.NotNull(queryable);
```
gathogojr, line 157: Is this assertion really sufficient?

KenitoInc: I have updated tests

KenitoInc:
Resolved by #980

PR #657, opened 2022-08-08
add omit-values supports

xuzhg:
OData specs has the following at: http://docs.oasis-open.org/odata/odata/v4.01/cs01/part1-protocol/odata-v4.01-cs01-part1-protocol.html#_Toc505771137
```txt
 8.2.8.6 [Preference omit-values](http://docs.oasis-open.org/odata/odata/v4.01/cs01/part1-protocol/odata-v4.01-cs01-part1-protocol.html#sec_Preferenceomitvalues)
The omit-values preference specifies values that MAY be omitted from a response payload. Valid values are nulls or defaults.

If nulls is specified, then the service MAY omit properties containing null values from the response, in which case it MUST specify the Preference-Applied response header with omit-values=nulls.

If defaults is specified, then the service MAY omit properties containing default values from the response, including nulls for properties that have no other defined default value. Nulls MUST be included for properties that have a non-null default value defined. If the service omits default values it MUST specify the Preference-Applied response header with omit-values=defaults.

Properties with null or default values MUST be included in delta payloads, if modified.

The response to a POST operation MUST include any properties not set to their default value, and the response to a PUT/PATCH operation MUST include any properties whose values were changed as part of the operation.

The omit-values preference does not affect a request payload. 
```

This PR is used to discuss the service side support for omit_values prefer header.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
@@ -11,6 +11,7 @@ using System.Linq;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.OData.Abstracts;
 using Microsoft.AspNetCore.OData.Common;
+using Microsoft.AspNetCore.OData.Formatter;
 using Microsoft.AspNetCore.OData.Formatter.Deserialization;
 using Microsoft.AspNetCore.OData.Query;
 using Microsoft.Extensions.DependencyInjection;
@@ -69,6 +70,44 @@ namespace Microsoft.AspNetCore.OData.Extensions
             }
 
             return request.HttpContext.ODataOptions();
+        }
+
+        /// <summary>
+        /// Returns the <see cref="OmitValuesKind"/> from the request.
+        /// </summary>
+        /// <param name="request">The <see cref="HttpRequest"/> instance to access.</param>
+        /// <returns>The <see cref="OmitValuesKind"/> from the request.</returns>
+        public static OmitValuesKind GetOmitValuesKind(this HttpRequest request)
+        {
+            if (request == null)
+            {
+                throw Error.ArgumentNull(nameof(request));
+            }
+
+            // The 'Prefer' header from client has the top priority
+            string preferHeader = RequestPreferenceHelpers.GetRequestPreferHeader(request.Headers);
+            if (preferHeader != null)
+            {
+                // simply use the string comparision case-insensitive
+                if (preferHeader.Contains("omit-values=nulls", StringComparison.OrdinalIgnoreCase))
+                {
+                    return OmitValuesKind.Nulls;
+                }
+
+                if (preferHeader.Contains("omit-values=defaults", StringComparison.OrdinalIgnoreCase))
+                {
+                    return OmitValuesKind.Defaults;
+                }
+            }
+
+            // If there's no setting from client, we move to use the configuration on the OData options.
+            ODataOptions options = request.ODataOptions();
+            if (options == null)
+            {
+                return OmitValuesKind.Unknown;
+            }
+
+            return options.OmitValuesKind;
         }
 
         /// <summary>
--- a/src/Microsoft.AspNetCore.OData/Formatter/ResponsePreferenceHelpers.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/ResponsePreferenceHelpers.cs
@@ -1,0 +1,55 @@
+//-----------------------------------------------------------------------------
+// <copyright file="ResponsePreferenceHelpers.cs" company=".NET Foundation">
+//      Copyright (c) .NET Foundation and Contributors. All rights reserved.
+//      See License.txt in the project root for license information.
+// </copyright>
+//------------------------------------------------------------------------------
+
+using System.Linq;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.OData.Extensions;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNetCore.OData.Formatter
+{
+    internal static class ResponsePreferenceHelpers
+    {
+        public const string PreferAppliedHeaderName = "Preference-Applied";
+
+        public static void SetResponsePreferAppliedHeader(HttpResponse response)
+        {
+            if (response == null)
+            {
+                throw Error.ArgumentNull(nameof(response));
+            }
+
+            HttpRequest request = response.HttpContext.Request;
+
+            OmitValuesKind omitValuesKind = request.GetOmitValuesKind();
+            if (omitValuesKind == OmitValuesKind.Unknown)
+            {
+                return;
+            }
+
+            string prefer_applied = null;
+            if (response.Headers.TryGetValue(PreferAppliedHeaderName, out StringValues values))
+            {
+                // If there are many "Preference-Applied" headers, pick up the first one.
+                prefer_applied = values.FirstOrDefault();
+            }
+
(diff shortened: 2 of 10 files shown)


src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
```diff
@@ -71,6 +72,44 @@ public static ODataOptions ODataOptions(this HttpRequest request)
             return request.HttpContext.ODataOptions();
         }
 
+        /// <summary>
+        /// Returns the <see cref="OmitValuesKind"/> from the request.
+        /// </summary>
+        /// <param name="request">The <see cref="HttpRequest"/> instance to access.</param>
+        /// <returns>The <see cref="OmitValuesKind"/> from the request.</returns>
+        public static OmitValuesKind GetOmitValuesKind(this HttpRequest request)
```
chrisspre, line 80: Should this indicate that it is a header value, or even a Prefer Header.  Something like GetOmitValueHeaderKind

src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
```diff
@@ -71,6 +72,44 @@ public static ODataOptions ODataOptions(this HttpRequest request)
...
+            if (request == null)
+            {
+                throw Error.ArgumentNull(nameof(request));
+            }
+
+            // The 'Prefer' header from client has the top priority
+            string preferHeader = RequestPreferenceHelpers.GetRequestPreferHeader(request.Headers);
+            if (preferHeader != null)
+            {
+                // simply use the string comparision case-insensitive
```
chrisspre, line 91:
```suggestion
                // use case insensitive string comparison
```

src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
xuzhg: Emm, actually it's not only from request header, it's also from ODataOption setting if there's no setting.

But, I am curious which one is a higher priority. @mikepizzo

julealgon: What about `GetOmitValuesPreferenceKind`, or even just `GetOmitValuesPreference` with a rename on the enum to `OmitValuesPreference`?

src/Microsoft.AspNetCore.OData/Formatter/ResponsePreferenceHelpers.cs
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Microsoft.AspNetCore.OData.Formatter
+{
+    internal static class ResponsePreferenceHelpers
+    {
+        public const string PreferAppliedHeaderName = "Preference-Applied";
+
+        public static void SetResponsePreferAppliedHeader(HttpResponse response)
+        {
+            if (response == null)
```
julealgon, line 21:
```suggestion
            if (response is null)
```

src/Microsoft.AspNetCore.OData/Formatter/ResponsePreferenceHelpers.cs
```diff
@@ -0,0 +1,55 @@
...
+                throw Error.ArgumentNull(nameof(response));
+            }
+
+            HttpRequest request = response.HttpContext.Request;
+
+            OmitValuesKind omitValuesKind = request.GetOmitValuesKind();
+            if (omitValuesKind == OmitValuesKind.Unknown)
+            {
+                return;
+            }
```
julealgon, line 32: Again, not a big fan of having this kind of logic. Can't we just make it so that `GetOmitValuesKind` returns a nullable, then use `null` as the indicator for "no value found", which is common for get-like methods?

src/Microsoft.AspNetCore.OData/Formatter/ResponsePreferenceHelpers.cs
```diff
@@ -0,0 +1,55 @@
...
+            {
+                // If there are many "Preference-Applied" headers, pick up the first one.
+                prefer_applied = values.FirstOrDefault();
+            }
+
+            string omitValuesHead = omitValuesKind == OmitValuesKind.Nulls ?
+                "omit-values=nulls" :
+                "omit-values=defaults";
+
+            if (prefer_applied == null)
```
julealgon, line 45:
```suggestion
            if (prefer_applied is null)
```

src/Microsoft.AspNetCore.OData/Formatter/ResponsePreferenceHelpers.cs
```diff
@@ -0,0 +1,55 @@
...
+            {
+                response.Headers[PreferAppliedHeaderName] = omitValuesHead;
```
julealgon, line 47: Instead of manually accessing each header like this, have you considered creating something like the native `GetTypedHeaders`, where each header is modeled as its own property with strong typing?
https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.headerdictionarytypeextensions.gettypedheaders?view=aspnetcore-6.0

It could be called `GetTypedODataHeaders` or something similar.

src/Microsoft.AspNetCore.OData/Formatter/ResponsePreferenceHelpers.cs
```diff
@@ -0,0 +1,55 @@
...
+            }
+
+            string prefer_applied = null;
+            if (response.Headers.TryGetValue(PreferAppliedHeaderName, out StringValues values))
+            {
+                // If there are many "Preference-Applied" headers, pick up the first one.
+                prefer_applied = values.FirstOrDefault();
+            }
+
+            string omitValuesHead = omitValuesKind == OmitValuesKind.Nulls ?
```
julealgon, line 41:
```suggestion
            string omitValuesHeaderValue = omitValuesKind == OmitValuesKind.Nulls ?
```

src/Microsoft.AspNetCore.OData/Formatter/ResponsePreferenceHelpers.cs
```diff
@@ -0,0 +1,55 @@
...
+                "omit-values=nulls" :
+                "omit-values=defaults";
```
julealgon, line 43: Instead of handling these with raw strings, have you considered creating a small model and/or factory method for it?

Imagine something like
```csharp
public sealed class PreferHeader
{
    public static PreferHeader OmitValues(OmitValuesKind kind)
    {
        return new("omit-values", kind.ToText())
    }
}
```
This would remove the need to ever handle raw strings. You can parse values into this model and use factory methods to create strongly typed instances for each preference OData has.

src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
```diff
@@ -71,6 +72,44 @@ public static ODataOptions ODataOptions(this HttpRequest request)
...
+                // use case insensitive string comparison
+                if (preferHeader.Contains("omit-values=nulls", StringComparison.OrdinalIgnoreCase))
+                {
+                    return OmitValuesKind.Nulls;
+                }
+
+                if (preferHeader.Contains("omit-values=defaults", StringComparison.OrdinalIgnoreCase))
+                {
+                    return OmitValuesKind.Defaults;
+                }
```
Nthemba, line 100: Is there a possibility for both 'omit-values=defaults' and 'omit-values=nulls' to be added as headers in a request?  In that case won't one of them be ignored?

Nthemba (commented):
Add some tests

PR #677, opened 2022-08-26
Update alternate key sample to support core alternatekey term

xuzhg:
Update alternate key sample to support core alternatekey term.

### DIFF
--- a/sample/ODataAlternateKeySample/Controllers/CustomersController.cs
+++ b/sample/ODataAlternateKeySample/Controllers/CustomersController.cs
@@ -50,7 +50,8 @@ namespace ODataAlternateKeySample.Controllers
         }
 
         // Alternate key: SSN
-        [HttpGet("odata/Customers(SSN={ssn})")]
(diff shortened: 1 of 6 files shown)


xuzhg:
@julealgon

gathogojr:
Maybe you should change `core_c_r` to `core_c_or_r` so it's easy to relate to the existing `c_or_r` (CountryOrRegion)

gathogojr:
Maybe you should change `CoreSN` to `CoreSSN` so it's easy to relate it to the existing `SSN`

xuzhg:
@gathogojr Thanks for your suggestion. Actually, I attentionally named them as it. What I thought is that it can be 'any' qualified alias. Is it ok for you?

gathogojr:
> @gathogojr Thanks for your suggestion. Actually, I attentionally named them as it. What I thought is that it can be 'any' qualified alias. Is it ok for you?

If it was intentional and not inadvertent that's okay...

gathogojr (approved):
:shipit:

Nthemba (approved):

PR #702, opened 2022-09-20
Updated Sample to add CRMRecordID, Odata as required

boris-barash:
Sorry Accidental pull request, Please delete


boris-barash:
Accident

Issue #727, opened 2022-10-26
[JsonIgnore] attribute ignored when using $expand


mattperdeck:
My API end points look like this:
```
        [HttpGet("List")]
        [EnableQuery]
        public override ActionResult<IQueryable<ApiScopeListV>> List()
        {
		    ....
            return query;
        }
```
The view models use JsonProperty attributes to rename JSON response fields, and JsonIgnore attributes to ensure some fields are not sent in the JSON response:
```
    public partial class ApiScopeListV
    {
        [JsonProperty("ScopeId")]
        public Guid ExternalId { get; set; }

        [JsonIgnore]
        public int Id { get; set; }

        [StringLength(250)]
        public string ScopeName { get; set; }

        [Column(TypeName = "money")]
        public decimal? ScopeLocalDiscount { get; set; }

        ....
    }
```
This works as expected as long as I don't use $expand in my OData queries. 

However, when I use $expand, the JsonProperty attributes are still honoured, but the JsonIgnore attributes are ignored. 

So with the ApiScopeListV objects above, without using $expand, the Id field is not present in the JSON response (as expected). But with $expand, that Id field does appear in the JSON response (not expected).

This bug happens both when using [Newtonsoft.Json.JsonIgnore] and [System.Text.Json.Serialization.JsonIgnore]. When trying [Newtonsoft.Json.JsonIgnore], I used AddODataNewtonsoftJson during startup to tell OData to use Newtonsoft.Json.

This same problem is referred to by @gordon-matt in 
https://github.com/OData/AspNetCoreOData/issues/424#issuecomment-1009565102

julealgon:
Can you provide full sample payloads with and without expansion for reference?

mattperdeck:
I just created a pull request to fix this issue: https://github.com/OData/AspNetCoreOData/pull/728

To answer your question, below I included:

- Response using $expand
- Response when not using $expand
- Versions of the OData library used
- EF model for parent item ApiInitiativeListV (including [JsonIgnore] attributes)
- EF model for child item ApiDeliverableListV (including [JsonIgnore] attributes)

==============================================================================
Response using $expand, with url
https://localhost/PublicApi/v1/Initiative/List?$top=5&$expand=Deliverables

Note presence of fields Id, InternalScopeId, InternalScopeId, etc.
```
[
    {
        "Deliverables": [],
        "InitiativeId": "a42b7982-22fe-4d5a-ab91-2b405f6e5580",
        "Id": 3112,
[...]

PR #728, opened 2022-10-27
fixes issue 727 [JsonIgnore] attribute ignored when using $expand

mattperdeck:
Fixes https://github.com/OData/AspNetCoreOData/issues/727

**Cause of the bug**

I found that when using $expand, the MapProperty methods of the default implementations for IPropertyMapper (for System.Text.Json and Newtonsoft.Json) contain code to interpret attributes applied to fields in the view model on their own. That is, they do not rely on the standard libraries to serialise the object, apply attributes, etc. 

Also, the only attribute they support is the property to rename a field - JsonProperty for Newtonsoft.Json, JsonPropertyName for System.Text.Json. Every other attribute would be ignored, including JsonIgnore. This explains why the bug is happening.

**Solution in this pull request**

1. Updated the MapProperty methods, so they check whether the field has the JsonIgnore attribute, and if so, return null (to indicate to ignore the field).
2. Updated the ToDictionary.... methods calling MapProperty so they no longer throw when MapProperty returns null. Instead, they simply do not add the field to the dictionairy. Also updated comments and error messages to reflect this change in the contract of the MapProperty method.
3. Updated affected unit tests, making sure that the new contract for MapProperty etc. is tested properly. Added unit tests for Newtonsoft Json version of MapProperty.

### DIFF
--- a/src/Microsoft.AspNetCore.OData.NewtonsoftJson/JsonPropertyNameMapper.cs
+++ b/src/Microsoft.AspNetCore.OData.NewtonsoftJson/JsonPropertyNameMapper.cs
@@ -49,6 +49,13 @@ namespace Microsoft.AspNetCore.OData.NewtonsoftJson
 
             IEdmProperty property = _type.Properties().Single(s => s.Name == propertyName);
             PropertyInfo info = GetPropertyInfo(property);
+
+            JsonIgnoreAttribute jsonIgnore = GetJsonIgnore(info);
+            if (jsonIgnore != null)
+            {
+                return null;
+            }
+
             JsonPropertyAttribute jsonProperty = GetJsonProperty(info);
             if (jsonProperty != null && !string.IsNullOrWhiteSpace(jsonProperty.PropertyName))
             {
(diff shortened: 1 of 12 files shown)


src/Microsoft.AspNetCore.OData.NewtonsoftJson/JsonPropertyNameMapper.cs
```diff
@@ -49,6 +49,13 @@ public string MapProperty(string propertyName)
 
             IEdmProperty property = _type.Properties().Single(s => s.Name == propertyName);
             PropertyInfo info = GetPropertyInfo(property);
+
+            JsonIgnoreAttribute jsonIgnore = GetJsonIgnore(info);
```
corranrogue9, line 53: Is this really the direction we want to go? Are we planning to support *all* `System.Text.Json` attributes?

corranrogue9: @xuzhg  @mikepizzo

corranrogue9 (changes requested):
We need to decide our policy on which attributes we are supporting

xuzhg (approved):

PR #729, opened 2022-10-27
Add matchesPattern support

### LINKED ISSUE #331: $filter does not support matchesPattern
I noticed that matchesPattern is not an allowed function. Can it be added?

matchesPattern a function in the OdataV4 url conventions 
http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#_Toc31360992

but not listed in https://github.com/OData/AspNetCoreOData/blob/master/src/Microsoft.AspNetCore.OData/Query/AllowedFunctions.cs

Arithmomaniac:
Fixes #331 

Requires bumps to dependencies because depends on https://github.com/OData/odata.net/pull/2519


xuzhg:
@Arithmomaniac Looks good to me.

mikepizzo (approved):
:shipit:

mikepizzo:
CLA Bot was down so unable to sign, but Arithmomaniac has already signed cla for other contributions, so I'm going to go ahead and merge.

Issue #730, opened 2022-10-27
Calling endpoint with $compute when Compute is not allowed in settings, still parses the whole expression


dnperfors:
**Assemblies affected**
ASP.NET Core OData 8.0.11

**Describe the bug**
I have an endpoint where I explicitly disallow all query options, but when I pass make a request using $compute, it still parses the expression.

**Reproduce steps**
Using the `ODataRoutingSample`:
- On the `PeopleController` change the [EnableQuery] on the get to `[EnableQuery(AllowedQueryOptions = AllowedQueryOptions.None)]`

**Request/Response**
- perform the request `/people/$compute=test add p12m`

Reponse tells me that the compute expression is invalid

**Expected behavior**
Expected the error "The query specified in the URI is not valid. Query option 'Compute' is not allowed."

xuzhg:
@dnperfors please take a look the PR. Thanks.

dnperfors:
For me this specific issue is fixed with the change in the PR. However, I can imagine that once I want to support the $compute, I also want to have the parsing validation done as well, without the need of implementing it myself.

xuzhg:
report it for forward expression validation.

xuzhg:
@dnperfors Please take a look.

PR #735, opened 2022-11-01
Enable $compute allowed settings

xuzhg:
issue #730

Enable to validate $compute if AllowedQueryOptions has correct setting.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/Validator/ODataQueryValidator.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Validator/ODataQueryValidator.cs
@@ -36,18 +36,13 @@ namespace Microsoft.AspNetCore.OData.Query.Validator
             // Validate each query options
             if (options.Compute != null)
             {
-                if (options.Compute.ComputeClause != null)
-                {
-                    ValidateQueryOptionAllowed(AllowedQueryOptions.Compute, validationSettings.AllowedQueryOptions);
-                }
+                ValidateQueryOptionAllowed(AllowedQueryOptions.Compute, validationSettings.AllowedQueryOptions);
+                options.Compute.Validate(validationSettings);
             }
 
             if (options.Apply != null)
             {
-                if (options.Apply.ApplyClause != null)
-                {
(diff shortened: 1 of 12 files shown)


src/Microsoft.AspNetCore.OData/Query/Validator/ODataQueryValidator.cs
```diff
@@ -36,18 +36,13 @@ public virtual void Validate(ODataQueryOptions options, ODataValidationSettings
             // Validate each query options
             if (options.Compute != null)
             {
-                if (options.Compute.ComputeClause != null)
-                {
-                    ValidateQueryOptionAllowed(AllowedQueryOptions.Compute, validationSettings.AllowedQueryOptions);
-                }
+                ValidateQueryOptionAllowed(AllowedQueryOptions.Compute, validationSettings.AllowedQueryOptions);
+                options.Compute.Validate(validationSettings);
```
dnperfors, line 40: This change leads to not parsing the ComputeClause during this validation. Is that desirable?

xuzhg: Actually,  in ODataQueryValidator, we validate whether the query option is allowed or not. 
In each Query validator, for example FilterQueryValidator, ComputeQueryValidator, we valid the content. 

For $compute clause, what expect error message do you want to get if the $compute clause is not valid?

dnperfors: If that is the way, other validators work as well, than that is fine with me.
If the ComputeQueryValidator will parse the string and returns error message when the expression is incorrect, than that would be good enough for me, just like the filter errors.

KenitoInc (approved):
LGTM

src/Microsoft.AspNetCore.OData/Query/Validator/ODataQueryValidator.cs
xuzhg: I will merge this PR and Let me try adding the expression validation in another PR.

Issue #767, opened 2022-12-13
Feature ask: Need the have ability to set custom query validator into SelectExpandQueryValidator


SamGuoMsft:
**Assemblies affected**
Which assemblies and versions are known to be affected e.g.: ASP.NET Core OData 8.x

**Describe the bug**
@xuzhg, based on our discussion, I am just listing three issues which I hope we can improve:
1. Can't set custom query validator into SelectExpandQueryValidator, e.g.: The construct directly create the FilterQueryValidator, rather than expose as parameter, so customer can use DI or explicitly set it. Source code [link](https://github.com/OData/AspNetCoreOData/blob/69eec03c7003fe12d92cdc619efdc16781683694/src/Microsoft.AspNetCore.OData/Query/Validator/SelectExpandQueryValidator.cs#:~:text=DefaultQuerySettings%20_defaultQuerySettings%3B-,///%20%3Csummary%3E,%7D,-///%20%3Csummary).
2. Even SelectExpandQueryValidator allow us pass custom FilterQueryValidator via constructor, it will still not work due to the current code does some kind of [hack](https://github.com/OData/AspNetCoreOData/blob/3db5e5fb94df2968b1b354bc37a47907b89892be/src/Microsoft.AspNetCore.OData/Query/Validator/FilterQueryValidator.cs#:~:text=internal%20virtual%20void,DefaultQuerySettings%20querySettings)) to create an internal virtual method to allow SelectExpandQueryValidator to set local properties like defaultValidationSetting. So even I create my custom filter query validator, I still can't set these fields and it will cause null reference issue eventually.
3. In the current FilterQueryValidator, if customer doesn't register EDM model, the validation will not honor the model bound filter attribute due to we don't set property and structured type if no ODataPath, source code is [here](https://github.com/OData/AspNetCoreOData/blob/3db5e5fb94df2968b1b354bc37a47907b89892be/src/Microsoft.AspNetCore.OData/Query/Validator/FilterQueryValidator.cs#:~:text=if%20(filterQueryOption.,%7D), but if people use filter under expand clause, then the model bound filter attribute start working due to it invoke the hacky internal method which I mentioned in 2nd issue, and the internal method will set the property and structured type based on navigation property. So, it is not consistent.

**Reproduce steps**
The simplest set of steps to reproduce the issue. If possible, reference a commit that demonstrates the issue.

**Data Model**
Please share your Data model, for example, your C# class.

**EDM (CSDL) Model**
[...]

PR #772, opened 2022-12-13
Fixes #771: add/use interface for query validator

### LINKED ISSUE #771: Add/use interface for query validator
**Assemblies affected**
Which assemblies and versions are known to be affected e.g. ASP.NET Core OData 8.1

Introduce the interfaces for all odata query validator
use the interfaces to register the validators

xuzhg:
Add interfaces for query validator

This is one of step for issue #767

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
@@ -153,7 +153,7 @@ namespace Microsoft.AspNetCore.OData.Query
         /// <summary>
         /// Gets or sets the query validator.
         /// </summary>
-        public ODataQueryValidator Validator { get; set; }
+        public IODataQueryValidator Validator { get; set; }
 
         /// <summary>
         /// Check if the given query option is an OData system query option using $-prefix-required theme.
--- a/src/Microsoft.AspNetCore.OData/Query/Query/ComputeQueryOption.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Query/ComputeQueryOption.cs
@@ -113,7 +113,7 @@ namespace Microsoft.AspNetCore.OData.Query
         /// <summary>
         /// Gets or sets the $compute Query Validator.
         /// </summary>
-        public ComputeQueryValidator Validator { get; set; }
+        public IComputeQueryValidator Validator { get; set; }
 
         /// <summary>
         /// Validate the $compute query based on the given <paramref name="validationSettings"/>. It throws an ODataException if validation failed.
--- a/src/Microsoft.AspNetCore.OData/Query/Validator/Interfaces/IFilterQueryValidator.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Validator/Interfaces/IFilterQueryValidator.cs
@@ -1,0 +1,23 @@
+//-----------------------------------------------------------------------------
+// <copyright file="IFilterQueryValidator.cs" company=".NET Foundation">
+//      Copyright (c) .NET Foundation and Contributors. All rights reserved.
+//      See License.txt in the project root for license information.
+// </copyright>
+//------------------------------------------------------------------------------
+
+namespace Microsoft.AspNetCore.OData.Query.Validator
+{
+    /// <summary>
+    /// Provide the interface used to validate a <see cref="FilterQueryOption"/>
+    /// based on the <see cref="ODataValidationSettings"/>.
+    /// </summary>
+    public interface IFilterQueryValidator
+    {
+        /// <summary>
+        /// Validates the OData query.
+        /// </summary>
+        /// <param name="filterQueryOption">The $filter query.</param>
+        /// <param name="validationSettings">The validation settings.</param>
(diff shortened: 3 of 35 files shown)


src/Microsoft.AspNetCore.OData/Query/Validator/Interfaces/IFilterQueryValidator.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// based on the <see cref="ODataValidationSettings"/>.
+    /// </summary>
+    public interface IFilterQueryValidator
+    {
+        /// <summary>
+        /// Validates the OData query.
+        /// </summary>
+        /// <param name="filterQueryOption">The $filter query.</param>
+        /// <param name="validationSettings">The validation settings.</param>
+        void Validate(FilterQueryOption filterQueryOption, ODataValidationSettings validationSettings);
```
corranrogue9, line 21: This doesn't have a return type, so which is expected to happen if the query is invalid?

xuzhg: Good question. 
So far, we'd expect to throw exceptions if 'bad' happens.

But, I'd 're-think' it in the next step.

src/Microsoft.AspNetCore.OData/Query/Validator/Interfaces/IODataQueryValidator.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// <summary>
+    /// Provide the interface used to validate a <see cref="ODataQueryOptions"/>
+    /// based on the <see cref="ODataValidationSettings"/>.
+    /// </summary>
+    public interface IODataQueryValidator
+    {
+        /// <summary>
+        /// Validates the OData query.
+        /// </summary>
+        /// <param name="options">The OData query to validate.</param>
```
gathogojr, line 19:
```suggestion
        /// <param name="options">The OData query options to validate.</param>
```

gathogojr (dismissed):
Left a few minor suggestions otherwise LGTM

src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
```diff
@@ -153,7 +153,7 @@ public ODataQueryOptions(ODataQueryContext context, HttpRequest request)
         /// <summary>
         /// Gets or sets the query validator.
         /// </summary>
-        public ODataQueryValidator Validator { get; set; }
+        public IODataQueryValidator Validator { get; set; }
```
mikepizzo, line 156: Isn't this a breaking change? Setting the property will still work, but any existing code that calls this property getter will expect an ODataQueryValidator, and will break if we change the return type to IODataQueryValidator.

src/Microsoft.AspNetCore.OData/Query/Query/ComputeQueryOption.cs
```diff
@@ -113,7 +113,7 @@ public ComputeClause ComputeClause
         /// <summary>
         /// Gets or sets the $compute Query Validator.
         /// </summary>
-        public ComputeQueryValidator Validator { get; set; }
+        public IComputeQueryValidator Validator { get; set; }
```
mikepizzo, line 116: Same as above; aren't all of these breaking changes?

src/Microsoft.AspNetCore.OData/Query/Validator/OrderByQueryValidator.cs
```diff
@@ -89,14 +89,14 @@ public virtual void Validate(OrderByQueryOption orderByOption, ODataValidationSe
             }
         }
 
-        internal static OrderByQueryValidator GetOrderByQueryValidator(ODataQueryContext context)
+        internal static IOrderByQueryValidator GetOrderByQueryValidator(ODataQueryContext context)
```
mikepizzo, line 92: Why would these not still return the concrete types?

mikepizzo (commented):
@xuzhg  -- This PR looks like it introduces breaking changes. Let's discuss before we release.

src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
xuzhg: No, It's not a breaking change.  ODataQueryValidator implements the IODataQueryValidator interface.
The existing codes still can work.

src/Microsoft.AspNetCore.OData/Query/Query/ComputeQueryOption.cs
xuzhg: No, It's not a breaking change.  ComputeQueryValidator implements the IComputeQueryValidator interface.
The existing codes still can work.

src/Microsoft.AspNetCore.OData/Query/Validator/OrderByQueryValidator.cs
xuzhg: If we return concrete type, we can't let users extend/customize it.

PR #777, opened 2022-12-19
Update azure pipeline

xuzhg:


pipelines/azure_pipelines_nightly.yml
```diff
@@ -40,6 +40,11 @@ stages:
       inputs:
         version: 5.x
```
ElizabethOkerio, line 41: should version be 6.x here?

pipelines/azure_pipelines.yml
```diff
@@ -31,6 +31,11 @@ jobs:
     inputs:
       version: 5.x
```
ElizabethOkerio, line 32: 6.x?

ElizabethOkerio (approved):

pipelines/azure_pipelines.yml
xuzhg: I added a new task to install 6.x. Later, I'd like to remove .net 5.

pipelines/azure_pipelines_nightly.yml
xuzhg: I added a new task to install 6.x. Later, I'd like to remove .net 5.

gathogojr (approved):

PR #783, opened 2022-12-21
Avoid generating wrapper for simple element types

bdebaere:
Currently a wrapper is generated for any IEnumerable other than byte[] and char[]. This can be optimized by avoiding the wrapper for any IEnumerable with a value type or string type as element type.


KenitoInc:
@bdebaere Add tests

bdebaere:
@KenitoInc Can you point me to where the current tests are for byte[] and char[] so I can add them there?

KenitoInc:
> @KenitoInc Can you point me to where the current tests are for byte[] and char[] so I can add them there?

All `SelectExpandBinder` tests are [here](https://github.com/OData/AspNetCoreOData/blob/main/test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/SelectExpandBinderTest.cs). If there are not tests for `byte[]` and `char[]`, add them.

KenitoInc:
Tests are failing

PR #795, opened 2023-01-05
Move Get Query validator as ODataQueryContext extension methods

xuzhg:
#issue #767

This is part of refactoring of issue #767.


src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.xml
```diff
@@ -14581,3 +14630,19 @@
         </member>
     </members>
 </doc>
+ummary>
```
gathogojr, line 14633: @xuzhg Is something wrong here?

xuzhg: It seems VS didn't fresh/update this XML in time every build.

I updated it. Please take a look for others. @gathogojr 

I will move to next part based on this changes.

gathogojr (approved):
:shipit:

PR #800, opened 2023-01-10
Enable custom query validato into SelectExpandQueryValidator

xuzhg:
issue #767

Refactor SelectExpandQueryValidator by:
1) Remove the private field
2) Create/use the validator context
3) Separate each part validation into a protected virtual, so the customer can customize each part.
4) Remove "very bad" internal virtual method in FilterQueryValidator
5) Other changes.


src/Microsoft.AspNetCore.OData/Query/Validator/SelectExpandQueryValidator.cs
```diff
@@ -77,163 +71,421 @@ public virtual void Validate(SelectExpandQueryOption selectExpandQueryOption, OD
...
-                        !expandItem.LevelsOption.IsMaxLevel &&
-                        (expandItem.LevelsOption.Level > Int32.MaxValue ||
-                        expandItem.LevelsOption.Level + currentDepth > maxDepth))))
+            foreach (SelectItem selectItem in selectExpandClause.SelectedItems)
+            {
+                if (selectItem is ExpandedNavigationSelectItem expandedSelectItem)
                 {
-                    throw new ODataException(
-                        Error.Format(SRResources.MaxExpandDepthExceeded, maxDepth, "MaxExpansionDepth"));
+                    // $expand=Nav
```
gathogojr, line 91: Do we need to retain this commented line?

src/Microsoft.AspNetCore.OData/Query/Validator/SelectExpandQueryValidator.cs
```diff
@@ -77,163 +71,421 @@ public virtual void Validate(SelectExpandQueryOption selectExpandQueryOption, OD
...
+            isExpandable = EdmHelpers.IsExpandable(property.Name, validatorContext.Property, validatorContext.StructuredType, edmModel, out expandConfiguration);
+            if (isExpandable)
+            {
+                int maxDepth = expandConfiguration.MaxDepth;
+                if (maxDepth > 0 && (validatorContext.RemainingDepth == null || maxDepth < validatorContext.RemainingDepth))
+                {
+                    remainDepth = maxDepth;
                 }
+
+                //if (expandConfiguration.MaxDepth > 0 && currentDepth >= expandConfiguration.MaxDepth)
```
gathogojr, line 177: Should this commented out code be removed?

src/Microsoft.AspNetCore.OData/Query/Validator/SelectExpandQueryValidator.cs
```diff
@@ -77,163 +71,421 @@ public virtual void Validate(SelectExpandQueryOption selectExpandQueryOption, OD
...
+        /// <param name="expandCountItem">The expand count item.</param>
+        /// <param name="validatorContext">The validator context.</param>
+        protected virtual void ValidateExpandedCountSelectItem(ExpandedCountSelectItem expandCountItem, SelectExpandValidatorContext validatorContext)
         {
-            if (topOption != null)
+            // So far, No default validation logic here.
+        }
+
+        /// <summary>
+        /// Validates one expand count. For example, ~/Customers?$expand=Nav/$ref
```
gathogojr, line 240:
```suggestion
        /// Validates one expand reference. For example, ~/Customers?$expand=Nav/$ref
```

src/Microsoft.AspNetCore.OData/Query/Validator/SelectExpandQueryValidator.cs
```diff
@@ -77,163 +71,421 @@ public virtual void Validate(SelectExpandQueryOption selectExpandQueryOption, OD
...
+        /// <param name="validatorContext">The validator context.</param>
+        protected virtual void ValidateExpandedReferenceSelectItem(ExpandedReferenceSelectItem expandReferItem, SelectExpandValidatorContext validatorContext)
+        {
+            // So far, No default validation logic here.
+        }
+
+        /// <summary>
+        /// Validates $select. For example, ~/Customers?$select=Prop($select=SubProp;$top=2)
+        /// </summary>
+        /// <param name="pathSelectItem"></param>
```
gathogojr, line 252: Supply the parameter description

src/Microsoft.AspNetCore.OData/Query/Validator/SelectExpandQueryValidator.cs
```diff
@@ -77,163 +71,421 @@ public virtual void Validate(SelectExpandQueryOption selectExpandQueryOption, OD
...
+            ValidateSkip(pathSelectItem.SkipOption, subValidatorContext);
+
+            // Validate the nested $count within $select
+            ValidateCount(pathSelectItem.CountOption, subValidatorContext);
+
+            // Validate the nested $search within $select
+            ValidateSearch(pathSelectItem.SearchOption, subValidatorContext);
+
+            // Validate the nested $compute within $select
+            ValidateCompute(pathSelectItem.ComputeOption, subValidatorContext);
```
gathogojr, line 323: I assume it's not valid to have `$levels` and `$apply` nested in a `$select` expression?

src/Microsoft.AspNetCore.OData/Query/Validator/SelectExpandQueryValidator.cs
```diff
@@ -77,163 +71,421 @@ public virtual void Validate(SelectExpandQueryOption selectExpandQueryOption, OD
...
-            if (orderByClause != null)
+            // So far, No default validation logic here.
+        }
+
+        /// <summary>
+        /// Validates $filter within $select or $expand
+        /// </summary>
+        /// <param name="filterClause">The nested $filter clause.</param>
+        /// <param name="validatorContext">The validator context.</param>
+        protected virtual void ValidateFilter(FilterClause filterClause, SelectExpandValidatorContext validatorContext)
```
gathogojr, line 371: For this method and the related ones, would it have been better to named them `ValidateNestedXXX`?

src/Microsoft.AspNetCore.OData/Query/Validator/SelectExpandQueryValidator.cs
```diff
@@ -77,163 +71,421 @@ public virtual void Validate(SelectExpandQueryOption selectExpandQueryOption, OD
...
-            if (filterClause != null)
+            if (orderByClause != null)
             {
-                _filterQueryValidator.Validate(property, structuredType, filterClause, validationSettings, edmModel, _defaultQuerySettings);
+                // TODO: OrderByModelLimitationsValidator is used already. but we should use IOrderbyQueryValidator to validate.
+                // Should change it later.
+                OrderByModelLimitationsValidator orderByQueryValidator =
+                   new OrderByModelLimitationsValidator(validatorContext.Context, validatorContext.Context.DefaultQuerySettings.EnableOrderBy);
+
+                orderByQueryValidator.TryValidate(validatorContext.Property, validatorContext.StructuredType, orderByClause, false);
```
gathogojr, line 410: Does `TryValidate` return a boolean value of `true`/`false` and should we be checking the return value?

src/Microsoft.AspNetCore.OData/Query/Validator/SelectExpandQueryValidator.cs
```diff
@@ -77,163 +71,421 @@ public virtual void Validate(SelectExpandQueryOption selectExpandQueryOption, OD
...
-            FilterClause filterClause, ODataValidationSettings validationSettings)
+        /// <summary>
+        /// Validates $orderby within $select or $expand
+        /// </summary>
+        /// <param name="orderByClause">The nested $orderby clause.</param>
+        /// <param name="validatorContext">The validator context.</param>
+        protected virtual void ValidateOrderby(OrderByClause orderByClause, SelectExpandValidatorContext validatorContext)
         {
-            if (filterClause != null)
+            if (orderByClause != null)
```
gathogojr, line 403: Just to observe that in some methods, you're doing:
```csharp
if (xxxOption == null)
{
    return;
}

// Validation logic here
```
Here and in some other methods you're nesting the validation logic within the `if` block...

src/Microsoft.AspNetCore.OData/Query/Validator/Interfaces/ISelectExpandQueryValidator.cs
```diff
@@ -19,5 +27,7 @@ public interface ISelectExpandQueryValidator
         /// <param name="selectExpandQueryOption">The $select and $expand query.</param>
         /// <param name="validationSettings">The validation settings.</param>
         void Validate(SelectExpandQueryOption selectExpandQueryOption, ODataValidationSettings validationSettings);
+
+       // Task ValidateAsync(SelectExpandClause selectExpandClause, ODataValidationSettings validationSettings);
```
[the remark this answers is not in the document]
xuzhg: Oh, Yes. Forgot it. Will update it.

src/Microsoft.AspNetCore.OData/Query/Validator/SelectExpandQueryValidator.cs
xuzhg: Yes. Thanks.

src/Microsoft.AspNetCore.OData/Query/Validator/SelectExpandQueryValidator.cs
xuzhg: #Resolved.

src/Microsoft.AspNetCore.OData/Query/Validator/SelectExpandQueryValidator.cs
xuzhg: No, They are not.

src/Microsoft.AspNetCore.OData/Query/Validator/SelectExpandQueryValidator.cs
xuzhg: I original thought we don't need 'nested' since they are 'protected virtual' within SelectExpandValidator.

But, I am ok to rename them.

src/Microsoft.AspNetCore.OData/Query/Validator/SelectExpandQueryValidator.cs
xuzhg: That's the old codes/existing codes that I will refactor later (next PR).

src/Microsoft.AspNetCore.OData/Query/Validator/SelectExpandQueryValidator.cs
xuzhg: It depends on the "number of codes line".  #won't fix.

src/Microsoft.AspNetCore.OData/Query/Validator/SelectExpandQueryValidator.cs
gathogojr: Given your explanation it would have been okay to retain them as they were...

xuzhg: #Resolved. Never mind, I updated them.

src/Microsoft.AspNetCore.OData/Query/Validator/FilterQueryValidator.cs
```diff
@@ -51,26 +51,13 @@ public virtual void Validate(FilterQueryOption filterQueryOption, ODataValidatio
                 throw Error.ArgumentNull(nameof(settings));
             }
 
-            if (filterQueryOption.Context.Path != null)
-            {
-                _property = filterQueryOption.Context.TargetProperty;
-                _structuredType = filterQueryOption.Context.TargetStructuredType;
-            }
-
+            _property = filterQueryOption.Context.TargetProperty;
```
habbes, line 54: I don't understand this change. Why was the if statement there in the first place?

src/Microsoft.AspNetCore.OData/Query/Validator/SelectExpandValidatorContext.cs
```diff
@@ -0,0 +1,44 @@
...
+    public class SelectExpandValidatorContext : QueryValidatorContext
+    {
+        /// <summary>
+        /// The top level $select and $expand query option.
+        /// </summary>
+        public SelectExpandQueryOption SelectExpand { get; set; }
+
+        /// <summary>
+        /// The remaining depth on property.
+        /// It's weird logic in current implementation. Need to improve it later.
```
habbes, line 22: What implementation does this comment refer to? Why is it weird? If it needs to be implemented late, consider creating an issue to track it.

src/Microsoft.AspNetCore.OData/Query/Validator/SelectExpandValidatorContext.cs
```diff
@@ -0,0 +1,44 @@
...
+        public SelectExpandValidatorContext Clone()
+        {
+            return new SelectExpandValidatorContext
+            {
+                SelectExpand = this.SelectExpand,
+                Context = this.Context,
+                ValidationSettings = this.ValidationSettings,
+                Property = this.Property,
+                StructuredType = this.StructuredType,
+                RemainingDepth = this.RemainingDepth,
```
habbes, line 39: Is `RemainingDepth` property value set by the customer or by the framework? At which point is it calculated?

gathogojr (approved):
:shipit:

src/Microsoft.AspNetCore.OData/Query/Validator/SelectExpandValidatorContext.cs
xuzhg: It's weird because the 'RemainingDepth' is working with the query setting configuration on the property. I can't understand such logic. 
Maybe it's correct, but I do not fully understand this. 
Let me update the description.

src/Microsoft.AspNetCore.OData/Query/Validator/SelectExpandValidatorContext.cs
xuzhg: In the SelectExpandQueryValidator.cs at line 192 at [SelectExpandQueryValidator.cs](https://github.com/OData/AspNetCoreOData/pull/800/files/7084be03534687098241c672594b6d1de75621ed#diff-b78354d7f70f4489d29dd669ad0f467ed341fe3451928e6767529720bf6593d6)

It's calculated at Line 170 - 176.  
The codes read the query setting on property (Model query setting). Model query setting is something 

```C#
public class Customer
{

       [Expand(MaxDepth=3)]
       public Order MyOrder {get;set;}
}
````

PR #817, opened 2023-01-20
fix: correctly find the target of a containment navigation property

mjameson-se:
This fixes an issue where the NavigationPropertySegment would have the wrong NavigationSource, resulting in an arcane error in certain scenarios because the associated entity type was wrong. The modified method is effectively a copy from [odata.net EdmExtensionMethods](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/EdmExtensionMethods.cs#L32), and it isn't clear why this line deviated.


gathogojr:
@mjameson-se To make it easier to understand the issue we're fixing here, would you mind creating an issue with repro steps demonstrating the scenario(s) which the existing code wasn't working as expected? Would you also mind adding a test to this PR covering the change that you made?

PR #820, opened 2023-01-24
Refactor FilterQueryValidator to make it thread safe

### LINKED ISSUE #819: Make FilterQueryValidator thread safe
**Assemblies affected**

**Describe the bug**
Currently, FilterQueryValidator has some private fields that make it not thread-safe.

We should refactor it to make FilterQueryValidator thread-safe.

xuzhg:
Fixes #819

Refactor FilterQueryValidator to make it thread safe.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/Validator/FilterQueryValidator.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Validator/FilterQueryValidator.cs
@@ -7,10 +7,10 @@
 
 using System.Diagnostics.Contracts;
 using System.Runtime.CompilerServices;
+using System.Xml;
 using Microsoft.AspNetCore.OData.Edm;
 using Microsoft.OData;
 using Microsoft.OData.Edm;
-using Microsoft.OData.ModelBuilder.Config;
 using Microsoft.OData.UriParser;
 
 namespace Microsoft.AspNetCore.OData.Query.Validator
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNetCore.OData/Query/Validator/FilterQueryValidator.cs
```diff
@@ -51,34 +37,29 @@ public virtual void Validate(FilterQueryOption filterQueryOption, ODataValidatio
...
         /// </summary>
         /// <param name="filterClause">The <see cref="FilterClause" />.</param>
-        /// <param name="settings">The validation settings.</param>
-        /// <param name="model">The EdmModel.</param>
-        /// <remarks>
-        /// Please note this method is not thread safe.
-        /// </remarks>
-        public virtual void Validate(FilterClause filterClause, ODataValidationSettings settings, IEdmModel model)
+        /// <param name="validatorContext">The validation context.</param>
+        protected virtual void ValidateFilter(FilterClause filterClause, FilterValidatorContext validatorContext)
```
KenitoInc, line 58: Won't this be a breaking change since we are removing an existing public method?

src/Microsoft.AspNetCore.OData/Query/Validator/FilterQueryValidator.cs
```diff
@@ -739,12 +720,12 @@ private void ValidateSingleValueNode(SingleValueNode node, ODataValidationSettin
             }
         }
 
-        private static void ValidateFunction(string functionName, ODataValidationSettings settings)
+        private static void ValidateFunction(string functionName, FilterValidatorContext validatorContext)
         {
-            Contract.Assert(settings != null);
+            Contract.Assert(validatorContext != null);
```
KenitoInc, line 725: nit: We should put a message in the Contract.Assert
e.g `Contract.Assert(validatorContext != null, "validatorContext != null");`

gathogojr (commented):
@xuzhg My only and main concern is the possible breaking changes in `FilterQueryValidator` class.

xuzhg:
> @xuzhg My only and main concern is the possible breaking changes in `FilterQueryValidator` class.

We are moving to 9.0. The major version will change.

src/Microsoft.AspNetCore.OData/Query/Validator/FilterQueryValidator.cs
xuzhg: @KenitoInc 
We are moving to 9.0. The major version will change.

gathogojr (approved):
:shipit:

src/Microsoft.AspNetCore.OData/Query/Validator/FilterQueryValidator.cs
xuzhg: Not necessary, i think

PR #824, opened 2023-01-26
Fixes #766, Support in memory filtering after ODataQueryOptions.ApplyTo

### LINKED ISSUE #766: Feature ask: support in memory filtering after ODataQueryOptions.ApplyTo()
I want to follow up on this very old ticket: https://github.com/OData/WebApi/issues/521

This issue was raised in 2015, and the same limitation seems still exist in the Odata.ApplyTo(). I am wondering is there a better solution than `You can do this to check for a $select wrapper, get a Dictionary of key,values and then map it your model.` at this moment?
image

Similar to the original question, I am trying to perform in memory filtering of a query, after it has been retrieved from the downstream API, and before returning it to the client.

Using ODataQueryOptions.ApplyTo(...) I can let OData do all it's query magic on the in-memory data. However if $expand or $select has been supplied, the returned IQueryable contains not the entities themselves, but SelectExpandWrapper<TElement> objects. Which it is impossible to extract the original entity from.

Ideally I would like to be able to do something like this in a controller action

```
public IQueryable<Entity> GetFeed(ODataQueryOptions<Entity> opts)
{
    IQueryable<Entity> result = opts.ApplyTo(<IEnumerable>.AsQueryable<Entity>())

    foreach (var item in result)
    {
        // perform extra filtering here
    }
    return result;
}

```

### Expected result

Any suggestions will be very helpful.

xuzhg:
Fixes #766.

If we have $select and $expand in the request query, the result is collection of ISelectExpandWrapper.

It's not easy for customers to do further. So Add 'Cast' Extension methods on IQueryable and Object.

Basic usage in the controller/action:
```C#
IQueryable query = options.ApplyTo(....);
IEnumerable<Customer> results = query.OCast<Customer>(query); // the new extension methods on IQueryable.

foreach (var a in results)
{
    // a is a partial object of Customer if we have $select and $expand.
}

```

Thanks.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Formatter/Value/TypedEdmStructuredObject.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Value/TypedEdmStructuredObject.cs
@@ -8,6 +8,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Diagnostics.Contracts;
+using System.Linq;
 using System.Reflection;
 using Microsoft.AspNetCore.OData.Common;
 using Microsoft.AspNetCore.OData.Edm;
@@ -108,14 +109,32 @@ namespace Microsoft.AspNetCore.OData.Formatter.Value
 
         private static Func<object, object> CreatePropertyGetter(Type type, string propertyName)
         {
-            PropertyInfo property = type.GetProperty(propertyName);
+            PropertyInfo propertyInfo = null;
+            var properties = type.GetProperties().Where(p => p.Name == propertyName).ToArray();
+            if (properties.Length <= 0)
+            {
+                propertyInfo = null;
+            }
+            else if (properties.Length == 1)
+            {
+                propertyInfo = properties[0];
+            }
+            else
+            {
+                // resolve 'new' modifier
+                propertyInfo = properties.FirstOrDefault(p => p.DeclaringType == type);
+                if (propertyInfo == null)
+                {
+                    propertyInfo = properties[0];
+                }
+            }
 
-            if (property == null)
+            if (propertyInfo == null)
             {
                 return null;
             }
 
-            var helper = new PropertyHelper(property);
+            var helper = new PropertyHelper(propertyInfo);
 
             return helper.GetValue;
         }
--- a/src/Microsoft.AspNetCore.OData/Query/IQueryableODataExtension.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/IQueryableODataExtension.cs
@@ -1,0 +1,218 @@
+//-----------------------------------------------------------------------------
+// <copyright file="IQueryableODataExtension.cs" company=".NET Foundation">
+//      Copyright (c) .NET Foundation and Contributors. All rights reserved.
+//      See License.txt in the project root for license information.
+// </copyright>
+//------------------------------------------------------------------------------
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
+using System.Linq;
+using System.Reflection;
+using Microsoft.AspNetCore.OData.Common;
+using Microsoft.AspNetCore.OData.Edm;
+using Microsoft.AspNetCore.OData.Formatter.Deserialization;
+using Microsoft.AspNetCore.OData.Query.Container;
+using Microsoft.AspNetCore.OData.Query.Wrapper;
+using Microsoft.OData;
+using Microsoft.OData.Edm;
+
+namespace Microsoft.AspNetCore.OData.Query
+{
+    /// <summary>
+    /// The cast options.
+    /// </summary>
+    public class ODataCastOptions
+    {
+        /// <summary>
+        /// Gets/sets the map provider.
+        /// </summary>
+        public Func<IEdmModel, IEdmStructuredType, IPropertyMapper> MapProvider { get; set; }
+    }
+
+    /// <summary>
+    /// Provides a set of static methods for querying data structures that implement <see cref="IQueryable"/>
+    /// </summary>
+    public static class IQueryableODataExtension
+    {
+        /// <summary>
+        /// Converts the source to the specified type.
+        /// </summary>
+        /// <typeparam name="TResult">The type to convert the source to.</typeparam>
+        /// <param name="source">The source.</param>
+        /// <param name="options">The cast options.</param>
+        /// <returns>The converted object if it's OData object. Otherwise, return same source or the default.</returns>
+        [SuppressMessage("ApiDesign", "RS0026:Do not add multiple public overloads with optional parameters", Justification = "<Pending>")]
+        public static TResult OCast<TResult>(this object source, ODataCastOptions options = null)
+        {
+            if (source is null)
+            {
+                return default;
+            }
+
+            Type sourceType = source.GetType();
+            if (source is SelectExpandWrapper wrapper)
+            {
+                return (TResult)CreateInstance(wrapper, sourceType, options);
+            }
+
+            if (typeof(TResult) == sourceType)
+            {
+                return (TResult)source;
+            }
+
+            if (typeof(TResult).IsAssignableFrom(sourceType))
+            {
+                return (TResult)Convert.ChangeType(source, typeof(TResult), CultureInfo.InvariantCulture);
+            }
+
+            return default;
+        }
+
+        /// <summary>
+        /// Converts the elements of an <see cref="IQueryable"/> to the specified type.
+        /// </summary>
+        /// <typeparam name="TResult">The type to convert the elements of source to.</typeparam>
+        /// <param name="source">The <see cref="IQueryable"/> that contains the elements to be converted.</param>
+        /// <param name="options">The cast options.</param>
+        /// <returns>Contains each element of the source sequence converted to the specified type.</returns>
+        [SuppressMessage("ApiDesign", "RS0026:Do not add multiple public overloads with optional parameters", Justification = "<Pending>")]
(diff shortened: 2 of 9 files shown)


xuzhg:
@julealgon Please take a look and share your thoughts.

src/Microsoft.AspNetCore.OData/Query/IQueryableODataExtension.cs
```diff
@@ -0,0 +1,218 @@
...
+                }
+
+                instance = Activator.CreateInstance(instanceType);
+            }
+
+            IDictionary<string, object> properties;
+            if (options != null && options.MapProvider != null)
+            {
+                properties = wrapper.ToDictionary(options.MapProvider);
+            }
```
julealgon, line 129: Prefer pattern matching for this type of check+extract:
```suggestion
            if (options is { MapProvider: not null and var provider})
            {
                properties = wrapper.ToDictionary(provider);
            }
```

src/Microsoft.AspNetCore.OData/Query/IQueryableODataExtension.cs
```diff
@@ -0,0 +1,218 @@
...
+            else
+            {
+                properties = wrapper.ToDictionary();
+            }
+
+            foreach (var property in properties)
+            {
+                string propertyName = property.Key;
+                object propertyValue = property.Value;
+                if (propertyValue == null)
```
julealgon, line 139: For consistency in new code.
```suggestion
                if (propertyValue is null)
```

src/Microsoft.AspNetCore.OData/Query/IQueryableODataExtension.cs
```diff
@@ -0,0 +1,218 @@
...
+            }
+            else
+            {
+                properties = wrapper.ToDictionary();
+            }
+
+            foreach (var property in properties)
+            {
+                string propertyName = property.Key;
+                object propertyValue = property.Value;
```
julealgon, line 138: Since `property` is a `KeyValuePair`:
```suggestion
            foreach (var (propertyName, propertyValue) in properties)
            {
```

src/Microsoft.AspNetCore.OData/Query/IQueryableODataExtension.cs
```diff
@@ -0,0 +1,218 @@
...
+                    throw new ODataException(Error.Format(SRResources.PropertyNotFound, instanceType.FullName, propertyName));
+                }
+
+                bool isCollection = TypeHelper.IsCollection(propertyInfo.PropertyType, out Type elementType);
+
+                if (isCollection)
+                {
+                    IList<object> collection = new List<object>();
+                    IEnumerable collectionPropertyValue = propertyValue as IEnumerable;
+                    foreach (var item in collectionPropertyValue)
```
julealgon, line 157: This is a mistake. `as` should not be used when the semantic is "the value _must_ be of the type", since it can return `null` and this will cause a `NullReferenceExeception` in the `foreach`.

```suggestion
                    IEnumerable collectionPropertyValue = (IEnumerable)propertyValue;
                    foreach (var item in collectionPropertyValue)
```

src/Microsoft.AspNetCore.OData/Query/IQueryableODataExtension.cs
```diff
@@ -0,0 +1,218 @@
...
+            return instance;
+        }
+
+        private static object CreateValue(object value, Type elementType, ODataCastOptions options)
+        {
+            Type valueType = value.GetType();
+
+            if (typeof(ISelectExpandWrapper).IsAssignableFrom(valueType))
+            {
+                SelectExpandWrapper subWrapper = value as SelectExpandWrapper;
```
julealgon, line 184: This looks dangerous to me. We are checking for a match with the interface `ISelectExpandWrapper`, but then casting to the concrete type.

It could be possible that the value implements the interface but is not the concrete `SelectExpandWrapper`, which would cause this to fail.

src/Microsoft.AspNetCore.OData/Query/IQueryableODataExtension.cs
```diff
@@ -0,0 +1,218 @@
...
+                propertyInfo = null;
+            }
+            else if (properties.Length == 1)
+            {
+                propertyInfo = properties[0];
+            }
+            else
+            {
+                // resolve 'new' modifier
+                propertyInfo = properties.FirstOrDefault(p => p.DeclaringType == type);
```
julealgon, line 208: Duplicated code.

src/Microsoft.AspNetCore.OData/Query/IQueryableODataExtension.cs
```diff
@@ -0,0 +1,218 @@
...
+                SelectExpandWrapper subWrapper = value as SelectExpandWrapper;
+                return CreateInstance(subWrapper, elementType, options);
+            }
+            else
+            {
+                return value;
+            }
+        }
+
+        private static PropertyInfo GetPropertyInfo(Type type, string propertyName)
```
julealgon, line 193: Looks like this could be extracted to its own separate extension method. To me this is such a generic concern that doesn't need to be inside the OData-specific extension class.

test/Microsoft.AspNetCore.OData.Tests/Query/IQueryableODataExtensionTests.cs
```diff
@@ -0,0 +1,236 @@
...
+            QueryCustomer customer = new QueryCustomer
+            {
+                Name = "A"
+            };
+
+            Assert.Null(customer.OCast<string>());
+
+            var expectCustomer = customer.OCast<QueryCustomer>();
+            Assert.Same(customer, expectCustomer);
+        }
```
julealgon, line 47: I'd strongly recommend breaking this into 3 separate tests either manually, or using data-driven tests.

src/Microsoft.AspNetCore.OData/Formatter/Value/TypedEdmStructuredObject.cs
```diff
@@ -108,14 +109,32 @@ public bool TryGetPropertyValue(string propertyName, out object value)
...
+            }
+            else
+            {
+                // resolve 'new' modifier
+                propertyInfo = properties.FirstOrDefault(p => p.DeclaringType == type);
+                if (propertyInfo == null)
+                {
+                    propertyInfo = properties[0];
+                }
+            }
```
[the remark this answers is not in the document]
xuzhg: I did use BindingFlags, but it seems it can't work. (Maybe what I tried is not enough).

bogdan-patraucean:
We need this. Any updates?

bogdan-patraucean:
We need this. Any updates??

mikepizzo:
We had a discussion on this today -- what is the expected behavior if there is a $select applied to the query on which OCast<T>() is called? should we ignore the $select and set all of the properties, or should we only set the properties that are specified in $select and leave it to the developer to somehow understand which properties have been set and which have not?

bogdan-patraucean:
I would expect the select to work. I can't use the select clause with IQueryable so I have to add the needed properties in my endpoint using the Select method.

From my understanding, this PR would fix the issue.

julealgon:
> or should we only set the properties that are specified in $select

Wouldn't that clash with `required` properties? What if the user didn't select some properties that are marked as `required` in the original model? If you then try to recreate the same object but only set the selected properties, this will fail.

src/Microsoft.AspNetCore.OData/Query/IQueryableODataExtension.cs
```diff
@@ -0,0 +1,218 @@
...
+        /// </summary>
+        /// <typeparam name="TResult">The type to convert the elements of source to.</typeparam>
+        /// <param name="source">The <see cref="IQueryable"/> that contains the elements to be converted.</param>
+        /// <param name="options">The cast options.</param>
+        /// <returns>Contains each element of the source sequence converted to the specified type.</returns>
+        [SuppressMessage("ApiDesign", "RS0026:Do not add multiple public overloads with optional parameters", Justification = "<Pending>")]
+        public static IEnumerable<TResult> OCast<TResult>(this IQueryable source, ODataCastOptions options = null)
+        {
+            if (source is null)
+            {
```
corranrogue9, line 86: this won't throw until enumeration; use `.select` instead

src/Microsoft.AspNetCore.OData/Query/IQueryableODataExtension.cs
```diff
@@ -0,0 +1,218 @@
...
+        /// <summary>
+        /// Converts the elements of an <see cref="IQueryable"/> to the specified type.
+        /// </summary>
+        /// <typeparam name="TResult">The type to convert the elements of source to.</typeparam>
+        /// <param name="source">The <see cref="IQueryable"/> that contains the elements to be converted.</param>
+        /// <param name="options">The cast options.</param>
+        /// <returns>Contains each element of the source sequence converted to the specified type.</returns>
+        [SuppressMessage("ApiDesign", "RS0026:Do not add multiple public overloads with optional parameters", Justification = "<Pending>")]
+        public static IEnumerable<TResult> OCast<TResult>(this IQueryable source, ODataCastOptions options = null)
+        {
```
corranrogue9, line 84: shouldn't we cast to iqueryable<selectexpandwrapper> and then do the operations? in cases where the queryable is not a selectexpand, what's written here will result in executing the query on the server just to find out that the method should have been a no-op

bytefish:
I am with @julealgon. I don't see how this could (or should) work with `required` properties. If a `required` property isn't present in the `$select` or `$expand`, then the only sane reason is to throw an `Exception`. The current behavior is the only one, that makes sense to me.

PR #833, opened 2023-02-03
fixes #825: add query option provider

### LINKED ISSUE #825: ODataQueryOptions is built multiple times during request process
**Assemblies affected**

**Describe the bug**
@xuzhg , based on our discussion, the ODataQueryOptions was built multple times during the request process:
1. During model binding time, the ODataQueryOptions was created once based on edm model.
2. EnableQueryAttribute action filter, when action executing, the ODataQueryOptions was created again based on edm model.

If developer writes his/her own query attribute which is inherited from EnableQueryAttribute, and override OnActionExecuting function with following logic:
- Build a customize edm model 
- Call base.OnActionExecuting

This actually causes the ODataQueryOptions from model binding are totally different than the ODataQueryOptions from EnableQueryAttributes.

We should think about to define a unified place to let customer customize EDM model and build ODataQueryOptions once and pass to every place.

**Reproduce steps**

**Data Model**

**EDM (CSDL) Model**
Please share your Edm model, for example, CSDL file.
You can send `$metadata` to get a CSDL XML content.

**Request/Response**

**Expected behavior**

**Screenshots**

**Additional context**

xuzhg:
fixes #825

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
@@ -87,6 +87,28 @@ namespace Microsoft.AspNetCore.OData.Extensions
         }
 
         /// <summary>
+        /// Gets the <see cref="IEdmModel"/> from the request container.
+        /// </summary>
+        /// <param name="request">The <see cref="HttpRequest"/> instance to extend.</param>
+        /// <returns>The <see cref="IODataQueryOptionsProvider"/> from the request container.</returns>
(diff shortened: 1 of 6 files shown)


xuzhg:
@[Sam Guo](https://github.com/SamGuoMsft) @[julealgon](https://github.com/julealgon)

Please take a look the design. Thanks.

gathogojr:
@xuzhg Update baseline files

xuzhg:
So far, I don't need this design. There's a simple solution see here at: https://github.com/OData/AspNetCoreOData/pull/846

PR #835, opened 2023-02-06
fixes #834, add DefaultQueryConfigurations class

### LINKED ISSUE #834: ASP.NET Core OData should have its own DefaultQuerySettings
**Describe the bug**
So far, we are using the 'DefaultQuerySettings' from the Model builder.  https://github.com/OData/ModelBuilder/blob/c5cd1ed8b41a7425a1a74766c2e5d8ba24870a2a/src/Microsoft.OData.ModelBuilder/Config/DefaultQuerySettings.cs

we use it here: https://github.com/OData/AspNetCoreOData/blob/main/src/Microsoft.AspNetCore.OData/ODataOptions.cs#L292

I think it's a mistake. Basically, we should create ASP.NET Core OData own 'DefaultQuerySettings', maybe use different name.

Benefit:
1) loose coupling. 
2) Most important, we can add more configuration items into settings. for example $compute, $search, etc. But all of these are not required at model builder side.

xuzhg:
fixes #834

1. Add DefaultQueryConfigurations class
2. Use it in ODataOptions (a public api changes)
3. change the related test codes.

Benefit:
1) It's a mistake to use DefaultQuerySetting from ModelBuilder ( I think)
2) Now, we can add other configuration, for example $compute, $search.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Edm/EdmHelpers.cs
+++ b/src/Microsoft.AspNetCore.OData/Edm/EdmHelpers.cs
@@ -137,10 +137,10 @@ namespace Microsoft.AspNetCore.OData.Edm
         }
 
         public static bool IsTopLimitExceeded(IEdmProperty property, IEdmStructuredType structuredType,
-           IEdmModel edmModel, int top, DefaultQuerySettings defaultQuerySettings, out int maxTop)
+           IEdmModel edmModel, int top, DefaultQueryConfigurations defaultQueryConfs, out int maxTop)
         {
             maxTop = 0;
-            ModelBoundQuerySettings querySettings = edmModel.GetModelBoundQuerySettings(property, structuredType, defaultQuerySettings);
+            ModelBoundQuerySettings querySettings = edmModel.GetModelBoundQuerySettings(property, structuredType, defaultQueryConfs);
             if (querySettings != null && top > querySettings.MaxTop)
             {
                 maxTop = querySettings.MaxTop.Value;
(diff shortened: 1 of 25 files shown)


ElizabethOkerio (dismissed):

ElizabethOkerio:
there are so many tests failing though

src/Microsoft.AspNetCore.OData/Edm/EdmHelpers.cs
```diff
@@ -137,10 +137,10 @@ public static IEnumerable<IEdmEntityType> GetAllDerivedEntityTypes(IEdmEntityTyp
         }
 
         public static bool IsTopLimitExceeded(IEdmProperty property, IEdmStructuredType structuredType,
-           IEdmModel edmModel, int top, DefaultQuerySettings defaultQuerySettings, out int maxTop)
+           IEdmModel edmModel, int top, DefaultQueryConfigurations defaultQueryConfs, out int maxTop)
```
gathogojr, line 140: nit:
```suggestion
           IEdmModel edmModel, int top, DefaultQueryConfigurations defaultQueryConfigs, out int maxTop)
```

src/Microsoft.AspNetCore.OData/Query/Validator/FilterQueryValidator.cs
```diff
@@ -25,7 +25,7 @@ public class FilterQueryValidator : IFilterQueryValidator
     {
         private int _currentAnyAllExpressionDepth;
         private int _currentNodeCount;
-        private DefaultQuerySettings _defaultQuerySettings;
+        private DefaultQueryConfigurations _defaultQueryConfig;
```
gathogojr, line 28: nit:
```suggestion
        private DefaultQueryConfigurations _defaultQueryConfigs;
```

src/Microsoft.AspNetCore.OData/Query/DefaultQueryConfigurations.cs
```diff
@@ -0,0 +1,67 @@
...
+// <copyright file="DefaultQueryConfigurations.cs" company=".NET Foundation">
+//      Copyright (c) .NET Foundation and Contributors. All rights reserved.
+//      See License.txt in the project root for license information.
+// </copyright>
+//------------------------------------------------------------------------------
+
+namespace Microsoft.AspNetCore.OData.Query
+{
+    /// <summary>
+    /// This class describes the default settings to use during query composition.
```
gathogojr, line 11: nit:
```suggestion
    /// This class describes the default configurations to use during query composition.
```

gathogojr (approved):
Left a few naming suggestions and minor comments. Otherwise LGTM

PR #840, opened 2023-02-11
Make internal SelectExpandBinder methods public

bdebaere:
Because so many methods are marked internal it really makes expanding OData behavior quite a hassle. I implore you to take a look at how extensible EntityFrameworkCore is for example and to reconsider all these internal methods everywhere.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
@@ -44,7 +44,7 @@ namespace Microsoft.AspNetCore.OData.Query.Expressions
         /// <summary>
         /// For unit test only.
         /// </summary>
-        internal SelectExpandBinder()
+        public SelectExpandBinder()
             : this(new FilterBinder(), new OrderByBinder())
         { }
 
@@ -89,7 +89,7 @@ namespace Microsoft.AspNetCore.OData.Query.Expressions
             return projectionLambdaExpression;
         }
 
-        internal Expression ProjectAsWrapper(QueryBinderContext context, Expression source, SelectExpandClause selectExpandClause,
+        public Expression ProjectAsWrapper(QueryBinderContext context, Expression source, SelectExpandClause selectExpandClause,
             IEdmStructuredType structuredType, IEdmNavigationSource navigationSource, OrderByClause orderByClause = null,
             ComputeClause computeClause = null,
             long? topOption = null,
@@ -336,7 +336,7 @@ namespace Microsoft.AspNetCore.OData.Query.Expressions
 
         // Generates the expression
         //      source => new Wrapper { Instance = source, Container = new PropertyContainer { ..expanded properties.. } }
-        internal Expression ProjectElement(QueryBinderContext context, Expression source, SelectExpandClause selectExpandClause, IEdmStructuredType structuredType, IEdmNavigationSource navigationSource)
+        public Expression ProjectElement(QueryBinderContext context, Expression source, SelectExpandClause selectExpandClause, IEdmStructuredType structuredType, IEdmNavigationSource navigationSource)
         {
             Contract.Assert(source != null);
 
@@ -476,7 +476,7 @@ namespace Microsoft.AspNetCore.OData.Query.Expressions
         /// <param name="propertiesToExpand">The out properties to expand at current level, could be null.</param>
         /// <param name="autoSelectedProperties">The out auto selected properties to include at current level, could be null.</param>
         /// <returns>true if the select contains dynamic property selection, false if it's not.</returns>
-        internal static IList<DynamicPathSegment> GetSelectExpandProperties(IEdmModel model, IEdmStructuredType structuredType, IEdmNavigationSource navigationSource,
+        public static IList<DynamicPathSegment> GetSelectExpandProperties(IEdmModel model, IEdmStructuredType structuredType, IEdmNavigationSource navigationSource,
             SelectExpandClause selectExpandClause,
             out IDictionary<IEdmStructuralProperty, PathSelectItem> propertiesToInclude,
             out IDictionary<IEdmNavigationProperty, ExpandedReferenceSelectItem> propertiesToExpand,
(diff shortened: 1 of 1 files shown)


xuzhg:
> Because so many methods are marked internal it really makes expanding OData behavior quite a hassle. I implore you to take a look at how extensible EntityFrameworkCore is for example and to reconsider all these internal methods everywhere.

Can you point me to any EFCore example? Thanks.

bdebaere:
@xuzhg EntityFrameworkCore has a ServiceCollection where you can add stuff to with for example `IDbContextOptionsExtension.ApplyServices(IServiceCollection services)`. OData has a similar thing where you can add or replace services with `ODataOptions.AddRouteComponents(string routePrefix, IEdmModel model, Action<IServiceCollection> configureServices)` and this is great!

However, I think you don't take it far enough. Take a look at these EntityFrameworkCore classes/interfaces:

- `RelationalQueryTranslationPreprocessor`
- `RelationalSqlTranslatingExpressionVisitor`
- `SqlServerQuerySqlGenerator`
- `IMemberTranslatorPlugin`
- `IRelationalTypeMappingSourcePlugin`

They allow extending because the necessary methods are public. Too many methods/classes are internal in OData which makes extending it without a custom copy of the code very difficult sometimes.

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -44,7 +44,7 @@ public SelectExpandBinder(IFilterBinder filterBinder, IOrderByBinder orderByBind
         /// <summary>
         /// For unit test only.
         /// </summary>
-        internal SelectExpandBinder()
+        public SelectExpandBinder()
```
KenitoInc, line 47: We don't need to make this method `public` since it's only used internally with some tests.

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -89,7 +89,7 @@ public virtual Expression BindSelectExpand(SelectExpandClause selectExpandClause
             return projectionLambdaExpression;
         }
 
-        internal Expression ProjectAsWrapper(QueryBinderContext context, Expression source, SelectExpandClause selectExpandClause,
+        public Expression ProjectAsWrapper(QueryBinderContext context, Expression source, SelectExpandClause selectExpandClause,
```
KenitoInc, line 92: Make method `virtual` and add `docstrings`.

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -336,7 +336,7 @@ public class Child : Father
 
         // Generates the expression
         //      source => new Wrapper { Instance = source, Container = new PropertyContainer { ..expanded properties.. } }
-        internal Expression ProjectElement(QueryBinderContext context, Expression source, SelectExpandClause selectExpandClause, IEdmStructuredType structuredType, IEdmNavigationSource navigationSource)
+        public Expression ProjectElement(QueryBinderContext context, Expression source, SelectExpandClause selectExpandClause, IEdmStructuredType structuredType, IEdmNavigationSource navigationSource)
```
KenitoInc, line 339: Make method `virtual` and add `docstrings`.

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -476,7 +476,7 @@ internal Expression ProjectElement(QueryBinderContext context, Expression source
         /// <param name="propertiesToExpand">The out properties to expand at current level, could be null.</param>
         /// <param name="autoSelectedProperties">The out auto selected properties to include at current level, could be null.</param>
         /// <returns>true if the select contains dynamic property selection, false if it's not.</returns>
-        internal static IList<DynamicPathSegment> GetSelectExpandProperties(IEdmModel model, IEdmStructuredType structuredType, IEdmNavigationSource navigationSource,
+        public static IList<DynamicPathSegment> GetSelectExpandProperties(IEdmModel model, IEdmStructuredType structuredType, IEdmNavigationSource navigationSource,
```
KenitoInc, line 479: This method is static and cannot be overriden. How do you plan to use it?

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -829,7 +829,7 @@ public virtual Expression CreateTotalCountExpression(QueryBinderContext context,
         /// <param name="navigationProperty">The expanded navigation property.</param>
         /// <param name="expandedItem">The expanded navigation select item. It may contain the nested query options.</param>
         /// <param name="includedProperties">The container to hold the created property.</param>
-        internal void BuildExpandedProperty(QueryBinderContext context, Expression source, IEdmStructuredType structuredType,
+        public void BuildExpandedProperty(QueryBinderContext context, Expression source, IEdmStructuredType structuredType,
```
KenitoInc, line 832: Make method `virtual`.

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -909,7 +909,7 @@ public virtual Expression CreateTotalCountExpression(QueryBinderContext context,
         /// <param name="structuralProperty">The selected structural property.</param>
         /// <param name="pathSelectItem">The selected item. It may contain the nested query options and could be null.</param>
         /// <param name="includedProperties">The container to hold the created property.</param>
-        internal void BuildSelectedProperty(QueryBinderContext context, Expression source, IEdmStructuredType structuredType,
+        public void BuildSelectedProperty(QueryBinderContext context, Expression source, IEdmStructuredType structuredType,
```
KenitoInc, line 912: Make method `virtual`.

KenitoInc (changes requested):

gathogojr:
@bdebaere Did you get a chance to look at the feedback from @KenitoInc?

PR #841, opened 2023-02-16
AutoExpand does not expand beyond non-default MaxExpansionDepth

gathogojr:
Fixes https://github.com/OData/WebApi/issues/2748

We set the default `MaxExpansionDepth` when the `Validate` method of the `ODataQueryOptions` object is called. The problem currently is that when the `MaxExpansionDepth` is overridden, for instance by assigning the value from the `EnableQuery` attribute (`[EnableQuery(MaxExpansionDepth = 4)]`), we call the `Validate` method before the `AutoExpand` navigation properties have been added to the `SelectExpandQueryOption` object. For that reason, the specified `MaxExpansionDepth` ends up not being respected for the auto-expanded navigation properties.

This pull request fixes that by calling `ODataQueryOptions.AddAutoSelectExpandProperties` prior to validating the `SelectExpand` query option.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
@@ -629,6 +629,8 @@ namespace Microsoft.AspNetCore.OData.Query
                 throw Error.ArgumentNull(nameof(validationSettings));
             }
 
+            this.Context.ValidationSettings = validationSettings;
+
             if (Validator != null)
             {
                 Validator.Validate(this, validationSettings);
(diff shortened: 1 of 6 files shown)


xuzhg (approved):

habbes (dismissed):

src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
```diff
@@ -835,6 +837,10 @@ internal void AddAutoSelectExpandProperties()
                 {
                     SelectExpand.LevelsMaxLiteralExpansionDepth = originalSelectExpand.LevelsMaxLiteralExpansionDepth;
                 }
+                else if (Context.ValidationSettings != null && Context.ValidationSettings.MaxExpansionDepth > 0)
```
habbes, line 840: Out of curiosity, 0 is not a valid max expansion depth?

habbes (approved):

src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
gathogojr: @habbes What my investigation revealed is that if you set a `MaxExpansionDepth` of 0, that doesn't get applied - it defaults to configured `MaxExpansionDepth` of 2. The following won't prohibit expansion currently. That's the reason I used `Context.ValidationSettings.MaxExpansionDepth > 0` in this PR to keep it consistent with the current behavior even as I follow up confirm if the current behavior is by design

xuzhg (approved):

Issue #704, opened 2022-09-26
No response to client if exception occurs in EF  


pebo:
**Assemblies affected**
```xml 
    <PackageReference Include="Microsoft.AspNetCore.OData" Version="8.0.11" />
    <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.30.1" />
    <PackageReference Include="Microsoft.EntityFrameworkCore.Cosmos" Version="6.0.9" />
```
**Describe the bug**
When getting errors from Entity Framework there's no response to the client due to

````
Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware: Warning: The response has already started, the error page middleware will not be executed.
warn: Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware[2]
````

Do I need to setup special error handling for OData and is it possible to return some kind of 500 / Internal Error response for this?

**Reproduce steps**
1. Create an ODataController:

````csharp
    [HttpGet]
    [EnableQuery()]
    public IActionResult Get()
    {
      return Ok(_dbContext.MyEntity);
    }
````
2. Craft a request that's not supported, e.g. `http://localhost:8080/api/entity?filter=id gt '2452'`
 
**Data Model**

**EDM (CSDL) Model**

**Request/Response**
*Request:* `http://localhost:8080/api/entity?filter=id gt '2452'`
The greater than operator isn't allowed for strings so this will throw an error in entity framework.

*Response:* None

**Expected behavior**
An Odata error response is returned (http://docs.oasis-open.org/odata/odata-json-format/v4.01/odata-json-format-v4.01.html#_Toc38457793)

**Screenshots**
-

**Additional context**
ODataController:

````csharp
    [HttpGet]
    [EnableQuery()]
    public IActionResult Get()
    {
      return Ok(_dbContext.Items);
    }
````

Stack trace:

````

      Entity Framework Core 6.0.9 initialized 'MyCosmosContext' using provider 'Microsoft.EntityFrameworkCore.Cosmos:6.0.9' with options: NoTracking ServiceEndPoint= Database=cosmos-entities-test
Microsoft.EntityFrameworkCore.Infrastructure: Information: Entity Framework Core 6.0.9 initialized 'MyCosmosContext' using provider 'Microsoft.EntityFrameworkCore.Cosmos:6.0.9' with options: NoTracking ServiceEndPoint= Database=cosmos-entities-test
[...]

PR #858, opened 2023-03-10
Write the HTTP response only when all output has been formatted

wedgberto:
### Purpose
This PR "buffers" the OData message wrapper in a memory stream until the OData query has been enumerated, then copies it into the HTTP response body only when no exception has been thrown.

### Scenario

A WebAPI endpoint is serviced by an Entity Framework 6.0 IQueryable. 
```
[HttpGet]
[EnableQuery]
public IQueryable<dto> Get()
{
    return this.dtoService.GetDtos();
}
```
When the SQL query is executed but the configured SQL command timeout is exceeded, our exception handling middleware successfully catches the SqlException, but is unable to set the HTTP status code to 500 because the HTTP response body is already started.

The endpoint returns a 200 OK HTTP status code and a body containing malformed JSON:
```
{"@odata.count":2,"value":[
```

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatterHelper.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatterHelper.cs
@@ -67,19 +67,20 @@ namespace Microsoft.AspNetCore.OData.Formatter
 
             ODataPath path = request.ODataFeature().Path;
             IEdmNavigationSource targetNavigationSource = path.GetNavigationSource();
-            HttpResponse response = request.HttpContext.Response;
+            var responseBody = new MemoryStream();
+            var responseHeaders = request.HttpContext.Response.Headers;
 
             // serialize a response
             string preferHeader = RequestPreferenceHelpers.GetRequestPreferHeader(requestHeaders);
             string annotationFilter = null;
             if (!string.IsNullOrEmpty(preferHeader))
             {
-                ODataMessageWrapper messageWrapper = ODataMessageWrapperHelper.Create(response.Body, response.Headers);
+                ODataMessageWrapper messageWrapper = ODataMessageWrapperHelper.Create(responseBody, responseHeaders);
                 messageWrapper.SetHeader(RequestPreferenceHelpers.PreferHeaderName, preferHeader);
                 annotationFilter = messageWrapper.PreferHeader().AnnotationFilter;
             }
 
-            IODataResponseMessageAsync responseMessage = ODataMessageWrapperHelper.Create(new StreamWrapper(response.Body), response.Headers, request.GetRouteServices());
+            IODataResponseMessageAsync responseMessage = ODataMessageWrapperHelper.Create(new StreamWrapper(responseBody), responseHeaders, request.GetRouteServices());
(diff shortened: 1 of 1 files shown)


wedgberto:
This might resolve: #704

wedgberto:

src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatterHelper.cs
```diff
@@ -150,6 +151,7 @@ public static ODataSerializerContext BuildSerializerContext(HttpRequest request)
                 }
 
                 await serializer.WriteObjectAsync(value, type, messageWriter, writeContext).ConfigureAwait(false);
+                await request.HttpContext.Response.Body.WriteAsync(responseBody.ToArray());
```
Airex, line 154: do we really need to create array of unknown size here?

src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatterHelper.cs
```diff
@@ -67,19 +67,20 @@ public static ODataSerializerContext BuildSerializerContext(HttpRequest request)
 
             ODataPath path = request.ODataFeature().Path;
             IEdmNavigationSource targetNavigationSource = path.GetNavigationSource();
-            HttpResponse response = request.HttpContext.Response;
+            var responseBody = new MemoryStream();
```
Airex, line 70: Assume to use memory stream  pooling

wedgberto: Do you mean Microsoft.IO.RecyclableMemoryStream ?

src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatterHelper.cs
wedgberto: i used `responseBody.CopyTo()` initially, but it didn't work how I expected so I resorted to this. I'll find a better way

wedgberto: > 

Ha, I was missing rewinding the buffer response body stream back to position zero

mikepizzo (changes requested):
We do not want to buffer in memory by default -- this can have a significant performance impact for large responses, and we do everything possible to make sure that we are able to stream a response without buffering.

In order to support efficient streaming, the [OData protocol]( http://docs.oasis-open.org/odata/odata-json-format/v4.01/odata-json-format-v4.01.html#sec_InStreamError) specifically describes instream error scenarios, calling for the response body to be malformed JSON.

src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatterHelper.cs
```diff
@@ -21,6 +21,7 @@
 using Microsoft.AspNetCore.OData.Formatter.Value;
```
wachugamaina, line 21: Could we add a test for this functionality ?

PR #872, opened 2023-03-27
Fix skiptoken paging edge case when combined with  involving nullable datetime property

gathogojr:
Related to https://github.com/OData/AspNetCoreOData/pull/859. Fixes skiptoken paging edge case when combined with  involving nullable datetime property

In the case of a `DateTime` property, it is mapped to `DateTimeOffset` since `DateTime` is mapped to `Edm.DateTimeOffset` in the Edm. This fix ensures proper handling of null when the skip token value of the `DateTime` property is null. The PR also adds tests for the edge case.


src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
```diff
@@ -329,7 +330,7 @@ private static IQueryable ApplyToCore(IQueryable query, ODataQuerySettings query
                         Expression condition = ExpressionBinderHelper.CreateBinaryExpression(
                             binaryOperator: BinaryOperatorKind.Equal,
                             left: property,
-                            right: parameterizeConstant ? LinqParameterContainer.Parameterize(propertyType, null) : Expression.Constant(null),
+                            right: parameterizeConstant ? LinqParameterContainer.Parameterize(property.Type, null) : Expression.Constant(null),
```
xuzhg, line 333: You reset propertyType using `propertyType = property.Type;`

why do you use `property.Type` again?

gathogojr: @xuzhg Since the Edm doesn't support `DateTime`, we map `DateTime` to `DateTimeOffset`. When later we're creating the binary expression to compare the `DateTimeOffset` value to a `DateTime` property, we convert the value back to a `DateTime` in the `CreateBinaryExpression` method at [this point](https://github.com/OData/AspNetCoreOData/blob/dd8b5cd73bd8cfa1756241171aa359452f60ae80/src/Microsoft.AspNetCore.OData/Query/Expressions/ExpressionBinderHelper.cs#L74)

Now, when the value is of type `DateTimeOffset?` and the value is **null**, everything breaks down [here](https://github.com/OData/AspNetCoreOData/blob/dd8b5cd73bd8cfa1756241171aa359452f60ae80/src/Microsoft.AspNetCore.OData/Query/Expressions/ExpressionBinderHelper.cs#L665). The `parameterizedConstantValue` is null, so `parameterizedConstantValue as DateTimeOffset?` returns null and consequently the type conversion from `DateTimeOffset?` to `DateTime?` doesn't happen and an exception gets thrown later as a result of trying to compare `DateTime?` (the type for the property) with `DateTimeOffset?` (the type for the value). I found this to be the easiest way to address the issue - by creating the binary expression with the actual property type when the value is null.

habbes (dismissed):

src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
```diff
@@ -266,7 +266,7 @@ private static IQueryable ApplyToCore(IQueryable query, ODataQuerySettings query
                 directionMap = new Dictionary<string, OrderByDirection>();
             }
 
-            IDictionary<string, Tuple<object, Type>> propertyValuePairs = PopulatePropertyValuePairs(skipTokenRawValue, context);
+            IDictionary<string, (object Value, Type Type)> propertyValuePairs = PopulatePropertyValuePairs(skipTokenRawValue, context);
```
ElizabethOkerio, line 269: what's the benefit of creating a tuple this way `(object Value, Type Type)` and what was there before? I saw using this `Tuple<object, Type>` creates a readonly structure whose elements cannot be changed. Is that the reasoning here?

ElizabethOkerio: I'm not sure whether the `Value` and `Type` are meant to overide the default field names `item1`, `item2` etc. If that's the case then I'd suggest we have more descriptive names to tell someone which value is being referred to for example. etc

gathogojr: > what's the benefit of using this format`(object Value, Type Type)` in place of what was there?

That is a `ValueTuple`. It's a struct and @habbes felt using it would help reduce allocations on the heap

src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
```diff
@@ -305,6 +305,7 @@ private static IQueryable ApplyToCore(IQueryable query, ODataQuerySettings query
                 else if (value is ODataNullValue)
                 {
                     value = null;
+                    propertyType = property.Type;
```
ElizabethOkerio, line 308: If you remove this change from here will there be any differences?

ElizabethOkerio: I see it is being used to `Parameterize`

src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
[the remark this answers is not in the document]
gathogojr: @ElizabethOkerio Use of `property.Type` here is deliberate due to the explanation provided here https://github.com/OData/AspNetCoreOData/pull/872#discussion_r1150068097

ElizabethOkerio (approved):

src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
```diff
@@ -405,11 +406,11 @@ private static IQueryable ApplyToCore(IQueryable query, ODataQuerySettings query
         /// <param name="value">The skiptoken string value.</param>
         /// <param name="context">The <see cref="ODataQueryContext"/> which contains the <see cref="IEdmModel"/> and some type information</param>
         /// <returns>Dictionary with property name and property value in the skiptoken value.</returns>
-        internal static IDictionary<string, Tuple<object, Type>> PopulatePropertyValuePairs(string value, ODataQueryContext context)
+        internal static IDictionary<string, (object PropertyValue, Type PropertyType)> PopulatePropertyValuePairs(string value, ODataQueryContext context)
         {
             Contract.Assert(context != null);
 
-            IDictionary<string, Tuple<object, Type>> propertyValuePairs = new Dictionary<string, Tuple<object, Type>>();
+            IDictionary<string, (object PropertyValue, Type PropertyType)> propertyValuePairs = new Dictionary<string, (object PropertyValue, Type PropertyType)>();
```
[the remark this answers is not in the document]
gathogojr: @KenitoInc `object` is being used here because the property value can be of varying types. For that reason, we cannot swap `object` with `ValueTuple`

KenitoInc (approved):

PR #880, opened 2023-03-31
tsharp/fix next page issue

tsharp:
There is an issue with client specified top sizes. When it is explicitly specified and meets the `top > pageSize` condition, a null url will be returned which is invalid. This code seems to have a purpose to right size the top statement.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Extensions/GetNextPageHelper.cs
+++ b/src/Microsoft.AspNetCore.OData/Extensions/GetNextPageHelper.cs
@@ -53,10 +53,6 @@ namespace Microsoft.AspNetCore.OData.Extensions
                             {
                                 value = (top - pageSize).ToString(CultureInfo.InvariantCulture);
                             }
-                            else
-                            {
-                                return null;
-                            }
                         }
                         break;
                     case "$skip":
--- a/test/Microsoft.AspNetCore.OData.Tests/Extensions/HttpRequestExtensionsTests.cs
+++ b/test/Microsoft.AspNetCore.OData.Tests/Extensions/HttpRequestExtensionsTests.cs
@@ -10,7 +10,10 @@ using Microsoft.AspNetCore.OData.Extensions;
 using Microsoft.AspNetCore.OData.Tests.Commons;
 using Moq;
 using System;
+using System.Collections.Generic;
+using System.Text;
 using Xunit;
+using Xunit.Sdk;
 
 namespace Microsoft.AspNetCore.OData.Tests.Extensions
 {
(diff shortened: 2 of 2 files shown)


tsharp:
@microsoft-github-policy-service agree company="Microsoft"

tsharp:
@gathogojr Right. so if you specify a $top in the request url the `@odata.nextLink` will be null instead of using the reduced page size. You can reproduce this issue by reverting the change I made and running my updated unit tests.

For example, `http://localhost/odata/example` would produce a next link but `http://localhost/odata/example?$top=5` will not.

gathogojr:
> Using your example above, if you specify the `$top=3` parameter, the following is produced:
> 
> ```json
> {
>     "@odata.context": "https://localhost:7280/$metadata#Customers",
>     "value": [
>         {
>             "Id": 1,
>             "Name": "Customer 1"
>         },
>         {
>             "Id": 2,
>             "Name": "Customer 2"
>         },
>         {
>             "Id": 3,
>             "Name": "Customer 3"
>         }
>     ]
> }
> ```

@tsharp That is by design https://github.com/OData/AspNetCoreOData/blob/9f1547e92a6bbed936dcae705491049014873be7/src/Microsoft.AspNetCore.OData/Extensions/GetNextPageHelper.cs#L51

tsharp:
@gathogojr According to the specification - this is not expected behavior and has regressed from v7 to v8 version of the OData Library.

https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_SystemQueryOptiontop

`The $top system query option specifies a non-negative integer n that limits the number of items returned from a collection. The service returns the number of available items up to but not greater than the specified value n.`

gathogojr:
@tsharp
> this is not expected behavior and has regressed from v7 to v8 version of the OData Library.

Can you confirm that this is a regression... Because from where I sit both v7 and v8 behave the same way. You can fire up a quick repro using Microsoft.AspNetCore.OData 7.6.5 version to confirm.

tsharp:
> @tsharp
> 
> > this is not expected behavior and has regressed from v7 to v8 version of the OData Library.
> 
> Can you confirm that this is a regression... Because from where I sit both v7 and v8 behave the same way. You can fire up a quick repro using Microsoft.AspNetCore.OData 7.6.5 version to confirm.

Fair point - I am unable to reproduce the issue as well in 7.4.*, 7.6.*, etc - but the other statements still align with the expected behavior of the `$top` query. See the documentation links to the OData spec in the previous comments.

Are we saying that the page size **must not** be controlled by the client or that the implementation details have some long-standing nuance that sets this precedence? It seems kind of random to not return a next link if the `$top` is specified in a query, no? maybe those implementation details might be better placed in a different spot? This would allow the developer to still call the `Request.GetNextPageLink` and have a desired output with supported code.

Either way, what do you suggest we do next here?

gathogojr:
@tsharp

About the regression, I just wanted us to start from correct premise.

Regarding the way forward, we have both server-controlled (via `PageSize`) and client-controlled (via `$top`) at play here. I want to understand well the reasoning behind the current implementation and the proposed fix to be sure we're covering as many angles as we can think of.

For instance, let's say the service is configured to return pages of 50 items, but the client would like to retrieve pages of 3 items at a time. Would it be right to fulfil the client's expectations in all scenarios? If we do, the server has to handle 17 roundtrips (17*3=51) to return 50 items that it had no problem returning in one go. What if the 50 items configuration was the server-side balance between too little and too much?

I don't think we can just consider `$top` without taking server-controlled paging configuration into account.
Allow me time to review the spec and also discuss it with different people.

We could also move the conversation to the [Discussions](https://github.com/OData/AspNetCoreOData/discussions) page to get views from the OData community.

cc. @xuzhg @julealgon @mikepizzo

tsharp:
That is a fair point as well - let's move this to discussions for now.

tsharp:
Based on the feedback and discussion it's unlikely that this will become a thing - closing the pr.

PR #904, opened 2023-05-02
Enable declared Untyped and undeclared Untyped

xuzhg:
1. Can read declared Untyped and undeclared Untyped, include collection
2. Can write declared Untyped and undeclared Untyped, include collection


sample/ODataRoutingSample/Controllers/PeopleController.cs
```diff
@@ -55,5 +58,67 @@ public IActionResult Get(string keyFirstName, string keyLastName)
...
+            {
+
+            }
+
+            // person.Data = 42; // ODataPrimitiveValue
+            //person.Other = new Address
+            //{
+            //    City = "Redmond",
+            //    Street = "148TH AVE NE"
+            //};
```
mikepizzo, line 81: are these commented lines things that are not supported today but should be supported in the future, or things that are supported but not part of this test, or?

sample/ODataRoutingSample/Controllers/PeopleController.cs
```diff
@@ -55,5 +58,67 @@ public IActionResult Get(string keyFirstName, string keyLastName)
...
+                "A string Value",
+                Color.Red,
+                AnyEnum.E1,
+                true,
+                new Address { City = "Issaquah", Street = "Klahanie Way" },
+                new EdmUntypedObject
+                {
+                    { "a1", "abc" }
+                },
+                // So far, it can't support it?
```
mikepizzo, line 107: are untyped collections supported with this PR (but not collections of structured types?)

sample/ODataRoutingSample/Models/Person.cs
```diff
@@ -17,5 +18,28 @@ public class Person
...
+
+        public object Data { get; set; } // Edm.Untyped
+
+        public object Other { get; set; }
+
+        public IList<object> Infos { get; set; } // Collection(Edm.Untyped)
+
+        public IList<object> Sources { get; set; } // Collection(Edm.Untyped)
+
+        public AwsCondition Condition { get; set; }
```
mikepizzo, line 30: >AwsCondition [](http://example.com/codeflow?start=15&length=12)

Use a more generic name (maybe something like CustomProperties?) in the public sample.

sample/ODataRoutingSample/Models/Person.cs
```diff
@@ -17,5 +18,28 @@ public class Person
...
         // [Column(Order = 2)]
         public string LastName { get; set; }
+
+        public object Data { get; set; } // Edm.Untyped
+
+        public object Other { get; set; }
+
+        public IList<object> Infos { get; set; } // Collection(Edm.Untyped)
+
+        public IList<object> Sources { get; set; } // Collection(Edm.Untyped)
```
mikepizzo, line 28: why do we need two untyped properties and two untyped collection properties?

src/Microsoft.AspNetCore.OData/Common/TypeHelper.cs
```diff
@@ -242,6 +248,33 @@ Type collectionInterface
...
+        internal static bool IsDictionary(Type clrType, out Type keyType, out Type valueType)
+        {
+            keyType = null;
+            valueType = null;
+            if (clrType == null)
+            {
+                return false;
+            }
+
+            // Do we need conside IDictionary interface? ==> No, let's skip it.
```
mikepizzo, line 260: >conside [](http://example.com/codeflow?start=26&length=7)

typo: "consider"

src/Microsoft.AspNetCore.OData/Common/TypeHelper.cs
```diff
@@ -242,6 +248,33 @@ Type collectionInterface
             return false;
         }
 
+        internal static bool IsDictionary(Type clrType, out Type keyType, out Type valueType)
```
mikepizzo, line 251: >IsDictionary [](http://example.com/codeflow?start=29&length=12)

This method is going to be expensive -- we need to be careful where we use it, and might consider caching the result if this is called repeatedly for the same type.

src/Microsoft.AspNetCore.OData/Common/TypeHelper.cs
```diff
@@ -242,6 +248,33 @@ Type collectionInterface
...
+                return false;
+            }
+
+            // Do we need conside IDictionary interface? ==> No, let's skip it.
+            Type collectionInterface
+                = clrType.GetInterfaces()
+                    .Union(new[] { clrType })
+                    .FirstOrDefault(
+                        t => t.IsGenericType
+                             && t.GetGenericTypeDefinition() == typeof(IDictionary<,>));
```
mikepizzo, line 266: consider rewriting this logic without relying on LINQ for better performance.

src/Microsoft.AspNetCore.OData/Edm/EdmHelpers.cs
```diff
@@ -20,6 +20,100 @@ namespace Microsoft.AspNetCore.OData.Edm
...
+        {
+            if (edmTypeReference == null)
+            {
+                return null;
+            }
+
+            if (edmTypeReference is IEdmStructuredTypeReference)
+            {
+                return (IEdmStructuredTypeReference)edmTypeReference;
+            }
```
mikepizzo, line 33: better would be either:

if (edmTypeReference is IEdmStructuredTypeReference structuredTypeReference)
{
    return structuredTypeReference;
}

or, if you can use that (requires C# 7), you can do:

IEdmStructuredTypeReference structuredTypeReference = edmTypeReference as IEdmStructuredTypeReference;
if(structuredTypeReference != null)
{
   return structuredTypeReference
}

either one will avoid the redundant type check.

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -590,6 +609,14 @@ private object ReadNestedResourceInline(ODataResourceWrapper resourceWrapper, IE
                 TimeZone = readContext.TimeZone
             };
 
+            if (edmType.IsUntyped())
+            {
+                nestedReadContext.ResourceType = typeof(EdmUntypedObject);
```
mikepizzo, line 614: >typeof(EdmUntypedObject) [](http://example.com/codeflow?start=49&length=24)

should we have a static for this to return the same instance of typeof(EdmUntypedObject), rather than call typeof() each time?

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -680,10 +706,17 @@ private object ReadNestedResourceInline(ODataResourceWrapper resourceWrapper, IE
...
-            IEdmStructuredTypeReference structuredType = edmType.AsCollection().ElementType().AsStructured();
+            var elementType = edmType.AsCollection().ElementType();
+            IEdmStructuredTypeReference structuredType = elementType.ToStructuredTypeReference();
+
             ODataDeserializerContext nestedReadContext = readContext.CloneWithoutType();
 
-            if (readContext.IsNoClrType)
+            if (elementType.IsUntyped())
+            {
+                //
```
mikepizzo, line 716: >// [](http://example.com/codeflow?start=16&length=2)

did you intend to add a comment here?

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceSetDeserializer.cs
```diff
@@ -139,17 +142,164 @@ public virtual IEnumerable ReadResourceSet(ODataResourceSetWrapper resourceSet,
...
+            {
+                throw new SerializationException(Error.Format(SRResources.TypeCannotBeDeserialized, edmType.FullName()));
+            }
+
+            IEdmTypeReference setItemElementType = edmType.AsCollection().ElementType();
+            IEdmStructuredTypeReference structuredType = setItemElementType.ToStructuredTypeReference();
+
+            ODataDeserializerContext nestedReadContext = readContext.CloneWithoutType();
+
+            if (setItemElementType.IsUntyped())
```
mikepizzo, line 235: >setItemElementType [](http://example.com/codeflow?start=16&length=18)

should this be setItemElementType or structuredType?

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -785,6 +791,89 @@ private async Task WriteDynamicComplexPropertyAsync(object propertyValue, IEdmTy
...
+                    continue;
+                }
+
+                // Ok, it's an resource or collection value.
+                if (actualType != null && actualType.IsStructuredOrUntyped())
+                {
+                    ODataNestedResourceInfo nestedResourceInfo
+                        = CreateUntypedNestedResourceInfo(structuralProperty, propertyValue, actualType, null/*not used now*/, resourceContext);
+
+                    if (nestedResourceInfo != null)
```
mikepizzo, line 836: >nestedResourceInfo [](http://example.com/codeflow?start=24&length=18)

when would this be null?

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -785,6 +791,89 @@ private async Task WriteDynamicComplexPropertyAsync(object propertyValue, IEdmTy
             await serializer.WriteObjectInlineAsync(propertyValue, edmType, writer, nestedWriteContext).ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// Writing the declared Edm.Untyped properties.
+        /// </summary>
+        private async Task WriteUntypedPropertiesAsync(SelectExpandNode selectExpandNode, ResourceContext resourceContext, ODataWriter writer)
```
mikepizzo, line 797: >WriteUntypedPropertiesAsync [](http://example.com/codeflow?start=27&length=27)

so untyped properties are written after typed primitive properties? why don't we just write them as we come across them?

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1183,6 +1279,85 @@ public virtual ODataStreamPropertyInfo CreateStreamProperty(IEdmStructuralProper
...
+            {
+                throw new SerializationException(
+                    Error.Format(SRResources.TypeCannotBeSerialized, structuralProperty.Type.FullName()));
+            }
+
+            ODataProperty odataProperty =
+                serializer.CreateProperty(propertyValue, actualType, structuralProperty.Name, writeContext);
+
+            // The following codes are not-needed but ODL has the problem to write ODataPrimitiveValue for 'Edm.Untyped' declared property
+            // This's the workaround and when ODL fixes the issue, we should remove these codes.
```
mikepizzo, line 1347: >     // This's the workaround and when ODL fixes the issue, we should remove these codes. [](http://example.com/codeflow?start=7&length=89)

do we have an ODL issue logged for this? maybe put a reference to it here for tracking.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataSerializerContext.cs
```diff
@@ -294,6 +312,23 @@ public IEdmNavigationProperty NavigationProperty
             }
         }
 
+        internal IEdmTypeReference TryGetEdmType(object instance, Type type)
```
mikepizzo, line 315: >TryGetEdmType [](http://example.com/codeflow?start=35&length=13)

the name of this makes me think it will return null if it can't get an enum type, so everywhere it's used I ask myself "what if it's null?  However, I think it's intended to *always* return an EdmType or an untyped value, so maybe name something like GetEdmTypeOrUntyped?

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -176,6 +192,138 @@ await WriteObjectInlineAsync(graph, resourceSetType, writer, writeContext)
...
+                // Let's directly use untyped collection serialization no matter what type this collection is.
+                itemEdmType = EdmUntypedHelpers.NullableUntypedCollectionReference;
+                await WriteResourceSetItemAsync(item, itemEdmType, parentSetType, writer, writeContext).ConfigureAwait(false);
+                return;
+            }
+
+            if (itemEdmType != null && itemEdmType.IsStructuredOrUntypedStructured())
+            {
+                await WriteResourceItemAsync(item, itemEdmType, parentSetType, writer, writeContext).ConfigureAwait(false);
+            }
```
mikepizzo, line 236: should we just check itemEdmType != null once and then nest the other checks within the if itemEdmTYpe != null?

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -176,6 +192,138 @@ await WriteObjectInlineAsync(graph, resourceSetType, writer, writeContext)
...
+            else
+            {
+                ODataEnumSerializer enumSerializer = writeContext?.Request?.GetRouteServices()?.GetRequiredService<ODataEnumSerializer>();
+                if (enumSerializer != null)
+                {
+                    ODataEnumValue oDataEnumValue = enumSerializer.CreateODataEnumValue(enumValue, enumType.AsEnum(), writeContext);
+
+                    // We can't write the 'ODataEnumValue' directly by calling write methods on writer.
+                    // Let's switch to use 'string' and write it as primitive value.
+                    // Once ODL supports, let's call the correct writing methods.
```
mikepizzo, line 289: >                  // Once ODL supports, let's call the correct writing methods. [](http://example.com/codeflow?start=2&length=79)

Can we create an issue in ODL and link it here?

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -176,6 +192,138 @@ await WriteObjectInlineAsync(graph, resourceSetType, writer, writeContext)
...
+        /// <param name="itemSetValue">The collection value.</param>
+        /// <param name="itemSetType">The nested collection Edm type.</param>
+        /// <param name="parentSetType">The parent collection edm type.</param>
+        /// <param name="writer">The writer.</param>
+        /// <param name="writeContext">The writer context.</param>
+        /// <returns>Task.</returns>
+        protected virtual async Task WriteResourceSetItemAsync(object itemSetValue, IEdmTypeReference itemSetType, IEdmTypeReference parentSetType,
+            ODataWriter writer, ODataSerializerContext writeContext)
+        {
+            IODataEdmTypeSerializer resourceSetSerializer = SerializerProvider.GetEdmTypeSerializer(itemSetType);
```
mikepizzo, line 307: >itemSetType [](http://example.com/codeflow?start=100&length=11)

what if itemSetType is null? do we need a null check or assert?

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -176,6 +192,138 @@ await WriteObjectInlineAsync(graph, resourceSetType, writer, writeContext)
...
+        /// <param name="resourceValue">The resource value.</param>
+        /// <param name="resourceType">The nested resource Edm type.</param>
+        /// <param name="parentSetType">The parent collection edm type.</param>
+        /// <param name="writer">The writer.</param>
+        /// <param name="writeContext">The writer context.</param>
+        /// <returns>Task.</returns>
+        protected virtual async Task WriteResourceItemAsync(object resourceValue, IEdmTypeReference resourceType, IEdmTypeReference parentSetType,
+            ODataWriter writer, ODataSerializerContext writeContext)
+        {
+            IODataEdmTypeSerializer resourceSerializer = SerializerProvider.GetEdmTypeSerializer(resourceType);
```
mikepizzo, line 323: >resourceType [](http://example.com/codeflow?start=97&length=12)

what if resourceType is null?

src/Microsoft.AspNetCore.OData/Formatter/Serialization/SelectExpandNode.cs
```diff
@@ -526,6 +526,8 @@ private void AddStructuralProperty(IEdmStructuralProperty structuralProperty, Pa
             }
             else
             {
+                // TODO: Do we support the query option on property whose type is 'Edm.Untyped' or 'Collection(Edm.Untyped)'?
```
mikepizzo, line 529: protocol supports this. also supports querying on dynamic properties of open types.  our support for declared untyped should be the same as for dynamic.

src/Microsoft.AspNetCore.OData/Routing/ODataPathExtensions.cs
```diff
@@ -124,6 +124,23 @@ public static string GetPathString(this IList<ODataPathSegment> segments)
...
         }
 
+        internal static bool IsUntypedPropertyPath(this ODataPath path)
+        {
+            if (path == null)
+            {
+                return false;
+            }
+
+            // TODO: do we need take the type cast into consideration?
```
mikepizzo, line 134: >// TODO: do we need take the type cast into consideration? [](http://example.com/codeflow?start=12&length=58)

unless the type cast types to Edm.Untyped, the presence of the type cast should indicate that it's not untyped.

mikepizzo (changes requested):
:clock1:

src/Microsoft.AspNetCore.OData/Formatter/Serialization/IUntypedResourceMapper.cs
```diff
@@ -0,0 +1,102 @@
...
+            IEnumerable<PropertyInfo> propeInfos =
+                    originalType.GetProperties(BindingFlags.Public | BindingFlags.Instance)
+                        .Where(prop => prop.GetIndexParameters().Length == 0 && prop.GetMethod != null);
+
+            foreach (PropertyInfo propInfo in propeInfos)
+            {
+                JsonIgnoreAttribute jsonIgnore = GetJsonIgnore(propInfo);
+                if (jsonIgnore != null)
+                {
+                    return null;
```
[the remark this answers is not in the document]
xuzhg: Great catch.

src/Microsoft.AspNetCore.OData/Formatter/Value/EdmObjectExtensions.cs
```diff
@@ -7,6 +7,7 @@
 
 using Microsoft.AspNetCore.OData.Deltas;
 using Microsoft.OData.Edm;
+using Microsoft.Spatial;
```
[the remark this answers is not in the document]
xuzhg: Removed

src/Microsoft.AspNetCore.OData/Routing/ODataPathExtensions.cs
xuzhg: Thanks for inputs.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
xuzhg: [https://github.com/OData/odata.net/issues/2659](https://github.com/OData/odata.net/issues/2659)

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
xuzhg: basically, it's protected method, we can guarantee the input is not null. But, I can add a check to make sure it's valid.

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceSetDeserializer.cs
xuzhg: should be use 'setItemElementType'

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
xuzhg: ODL fixed it using this PR:
[https://github.com/OData/odata.net/pull/2664](https://github.com/OData/odata.net/pull/2664)

src/Microsoft.AspNetCore.OData/Common/TypeHelper.cs
xuzhg: The test 'clrType' could be lots of it. If we cache it, it could be huge. And This test could be fast and only for untyped scenario, so, i think we don't need to cache it.

src/Microsoft.AspNetCore.OData/Common/TypeHelper.cs
xuzhg: Rewrote it.

sample/ODataRoutingSample/Models/Person.cs
xuzhg: Renamed.

sample/ODataRoutingSample/Models/Person.cs
xuzhg: One set for 'reading' from payload and writing them again in the response.

the other set is only 'setting' the value in the controller, and writing them in the response.

sample/ODataRoutingSample/Controllers/PeopleController.cs
xuzhg: No, Here lists the different values that customer can assign to 'object' to represent a resource. I updated with comments in the code.

sample/ODataRoutingSample/Controllers/PeopleController.cs
xuzhg: A resource in a nested collection of a 'untyped' collection doesn't support.  See [https://github.com/OData/odata.net/issues/2661](https://github.com/OData/odata.net/issues/2661). And I want to fix it soon with ODL.

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
xuzhg: typeof() is a compile behavior, so,The typeof is an operator keyword which is used to get a type at the compile-time. So, it's ok here to use typeof.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
xuzhg: We don't know how to write the untyped properties until we know the real type of the property value.
And it goes to different writing process based on the type kind. For example, if the real property value is a primitive, enum or collection of them, we should create a 'ODataProperty' and save it into the 'ODataResource.Properties'.

if the real property value is a collection of resource, or resource, it goes to write using nested resource info.

I don't want to pollute the 'non-untyped' writing process, So, keep untyped in its own writing method.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1080,6 +1169,13 @@ private IEnumerable<ODataProperty> CreateStructuralPropertyBag(SelectExpandNode
                         continue;
                     }
 
+                    if (structuralProperty.Type != null &&
+                        (structuralProperty.Type.IsUntyped() || structuralProperty.Type.IsCollectionUntyped()))
+                    {
+                        // skip it here, we use a different method to write all 'declared' untyped properties
+                        continue;
```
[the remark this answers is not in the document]
xuzhg: Addressed in another review comments. In a summary, 
We don't know the real type of the untyped property value until we get the value. And different type goes to different writing logic. it's primitive, or it's nested resource info? 

So, same as the 'Stream' property, use its own logic to write.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataSerializerProvider.cs
```diff
@@ -143,6 +145,21 @@ public virtual IODataSerializer GetODataPayloadSerializer(Type type, HttpRequest
...
+                // (Known primitive, known enum, known strucutred or collection are handled above.)
+                // One of such case is that it's an untyped property (or dynamic property) query request
+                // and the value type is an unknown to OData Edm model.
+                if (path.IsUntypedPropertyPath())
+                {
+                    edmType = TypeHelper.IsDictionary(type, out _, out _) ?
+                        (IEdmTypeReference)EdmUntypedStructuredTypeReference.NullableTypeReference :
+                        (TypeHelper.IsCollection(type) ?
+                            (IEdmTypeReference)EdmUntypedHelpers.NullableUntypedCollectionReference :
+                            (IEdmTypeReference)EdmUntypedStructuredTypeReference.NullableTypeReference);
```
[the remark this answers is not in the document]
xuzhg: Created.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataSerializerContext.cs
```diff
@@ -294,6 +312,23 @@ public IEdmNavigationProperty NavigationProperty
...
+        {
+            try
+            {
+                return GetEdmType(instance, type);
+            }
```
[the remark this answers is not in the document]
xuzhg: refactored, please take a look again.

src/Microsoft.AspNetCore.OData/Common/TypeHelper.cs
mikepizzo: I raised the concern because the code uses reflection, which is a concern for performance-critical code.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataSerializerContext.cs
```diff
@@ -294,51 +312,60 @@ public IEdmNavigationProperty NavigationProperty
...
+                        IEdmTypeReference actualType = Model.GetEdmTypeReference(instance.GetType());
+                        if (actualType != null && actualType != edmType)
+                        {
+                            edmType = actualType;
+                        }
                     }
                 }
             }
 
+            if (edmType == null && isUntyped)
```
mikepizzo, line 363: >edmType == null && isUntyped [](http://example.com/codeflow?start=16&length=28)

so if edmType is (somehow) null, and isUntyped if false, will we now return null where previously we would throw an exception?

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataSerializerContext.cs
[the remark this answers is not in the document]
mikepizzo: much better -- thanks!

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
mikepizzo: so this code gets removed when we update our dependency to the latest ODL with this fix?

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataSerializerContext.cs
mikepizzo: :thumbsup:

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
mikepizzo: doesn't look like it was addressed, but it's not a big deal either.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
mikepizzo: :thumbsup:

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
mikepizzo: I get that there is different logic for untyped versus typed, but I'm not sure why we can't invoke that separate logic when iterating through the list of typed properties, rather than doing the iteration twice. Not a big deal, but we may want to revisit this in the future, if customers complain about re-ordering properties.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
xuzhg: I updated the comments

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
xuzhg: basically, it's protected method, we can guarantee the input is not null. But, I can add a check to make sure it's valid.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataSerializerContext.cs
xuzhg: No, if 'isUntyped==false', it is the previous logic. Nothing changed.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
xuzhg: Already resolved. 1) Remove itemEdmtype != null for all. 2) Check itemEdmType at the beginning.

mikepizzo (approved):
:shipit:

PR #905, opened 2023-05-02
#907: Enable read the untyped collection

### LINKED ISSUE #885: Enable dynamic complex property without @odata.type
**Assemblies affected**

**Reproduce steps**

Consider the following json request payload,  the type of 'condition' is an open type, and it accepts dynamic properties.

where, `StringEqualsIgnoreCase` is dynamic complex type property

```json
{
   "condition": {
         "StringEqualsIgnoreCase": {
               "awsTag":[
                         "finance",
                          "hr",
                         "legal"
                    ]
                },
                "key1": "value1",
                 "key2": [
                   "value2",
                   "value3"
                ],
                "key3": "value4"
       }
}
```

The above payload can't pass deserialization because there's no type for
a) StringEqualsIgnoreCase
b) awsTag
C) key2

We have to let customer provide types as below, then it can work.

```json
{
  "condition": {
    "StringEqualsIgnoreCase": {
        "@odata.type":"microsoft.graph.awsCondition",
        "awsTag@odata.type":"Collection(Edm.String)",
        "awsTag":[
            "finance",
            "hr",
            "legal"
        ]
      },
      "key1": "value1",
      "key2@odata.type": "Collection(Edm.String)",
      "key2": [
          "value2",
          "value3"
      ],
      "key3": "value4"
    }
}
```

**Expected behavior**
Actually, we don't need customers to provide the types since they are dynamic properties.

**Additional context**
We can deserialize 'StringEqualsIgnoreCase' as an instance of 'ODataObject', where 'ODataObject' can be defined as:

```C#
public class ODataObject: Dictionary<string, object>
{}
```

### LINKED ISSUE #907: Reading non-resource in a resource set throw exception
**Assemblies affected**

Suppose I have the following request payload (ODL supports reading the following dynamic property)

```json
{
    "Name": "Sam",
    "ADynamicProperty": [
         42,
         "a string value",
         { "key1":"value1"}
   ] 
}
```
ODL can read the above dynamic property "ADynamicProperty".
It returns the following reader state:
```txt
ResourceStart (top-Level)
  NestedResourceInfoStart (ADynamicProperty)
     ResourceSetStart
           Primitive (42)
           Primitive (a string value)
           ResourceStart
           ResourceEnd
     ResourceSetEnd
  NestedResourceInfoEnd
ResourceEnd
```

**Expected behavior**

ASP.NET Core can handle the above scenario, it means it can handle the 'Primitive' reader state.

**Actual behavior**

There's no logic to handle 'Primitive' reader state.

xuzhg:
fixes #907.

also part of fixes #885. (No breaking changes)

Enable read the untyped collection.

For example:
```json

{
 "awsTag":[
            null,
            {
                "X1": "Red",
                "Data": {
                    "D1": 42
                }
            },
            "finance",
            "hr",
            "legal",
            43
        ]
}
```

where, awsTag could be a dynamic property, it contains primitive value item, resource value item.

ODL reads it and returns:
1) ODataReaderState.Primitive  for primitive value in collection
2) ODataReaderState.ResourceStart  for resource value in collection.
3) It also can contain collection value item again in the collection.

This PR is to handle the above scenario.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataReaderExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataReaderExtensions.cs
@@ -238,6 +238,7 @@ namespace Microsoft.AspNetCore.OData.Formatter.Wrapper
                 if (parentResourceSet != null)
                 {
                     parentResourceSet.Resources.Add(resourceWrapper);
+                    parentResourceSet.Items.Add(resourceWrapper);// in the next major release, we should only use 'Items'.
                 }
                 else if (parentDeleteResourceSet != null)
                 {
--- a/src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataReaderExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataReaderExtensions.cs
@@ -311,7 +311,7 @@ namespace Microsoft.AspNetCore.OData.Formatter.Wrapper
             if (itemsStack.Count > 0)
             {
                 ODataItemWrapper peekedWrapper = itemsStack.Peek();
-                if (peekWrapper is ODataNestedResourceInfoWrapper parentNestedResourceInfo)
+                if (peekedWrapper is ODataNestedResourceInfoWrapper parentNestedResourceInfo)
                 {
                     Contract.Assert(parentNestedResourceInfo.NestedResourceInfo.IsCollection == true, "Only collection nested properties can contain resource set as their child.");
                     Contract.Assert(parentNestedResourceInfo.NestedItems.Count == 0, "Each nested property can contain only one resource set as its direct child.");
@@ -319,7 +319,7 @@ namespace Microsoft.AspNetCore.OData.Formatter.Wrapper
                 }
                 else
                 {
-                    ODataResourceSetWrapper parentResourceSet = (ODataResourceSetWrapper)peekWrapper;
+                    ODataResourceSetWrapper parentResourceSet = (ODataResourceSetWrapper)peekedWrapper;
                     parentResourceSet.Items.Add(resourceSetWrapper);
                 }
             }
--- a/src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataReaderExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataReaderExtensions.cs
@@ -310,7 +310,7 @@ namespace Microsoft.AspNetCore.OData.Formatter.Wrapper
             ODataResourceSetWrapper resourceSetWrapper = new ODataResourceSetWrapper(resourceSet);
             if (itemsStack.Count > 0)
             {
-                ODataItemWrapper peekWrapper = itemsStack.Peek();
+                ODataItemWrapper peekedWrapper = itemsStack.Peek();
                 if (peekWrapper is ODataNestedResourceInfoWrapper parentNestedResourceInfo)
                 {
                     Contract.Assert(parentNestedResourceInfo.NestedResourceInfo.IsCollection == true, "Only collection nested properties can contain resource set as their child.");
--- a/src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataReaderExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataReaderExtensions.cs
@@ -5,7 +5,6 @@
 // </copyright>
 //------------------------------------------------------------------------------
 
-using System;
 using System.Collections.Generic;
 using System.Diagnostics.Contracts;
 using System.Threading.Tasks;
@@ -191,6 +190,16 @@ namespace Microsoft.AspNetCore.OData.Formatter.Wrapper
                     linkResourceSetWrapper.DeltaItems.Add(linkBaseWrapper);
                     break;
 
+                case ODataReaderState.Primitive:
+                    Contract.Assert(itemsStack.Count > 0, "The primitive should be a non-null primitive value within an untyped collection.");
+                    // Be noted:
+                    // 1) if a 'null' value or a resource/object in the untyped collection goes to ODataResource flow
+                    // 2) if a collection value in the untyped collection goes to ODataResourceSet flow
+                    // 3) Since it's untyped, there's no logic for 'Enum' value, it means it's treated as primitive value.
+                    ODataResourceSetWrapper resourceSetParentWrapper = (ODataResourceSetWrapper)itemsStack.Peek();
+                    resourceSetParentWrapper.Items.Add(new ODataPrimitiveWrapper((ODataPrimitiveValue)reader.Item));
+                    break;
+
                 default:
                     Contract.Assert(false, "We should never get here, it means the ODataReader reported a wrong state.");
                     break;
@@ -301,11 +310,18 @@ namespace Microsoft.AspNetCore.OData.Formatter.Wrapper
             ODataResourceSetWrapper resourceSetWrapper = new ODataResourceSetWrapper(resourceSet);
             if (itemsStack.Count > 0)
             {
-                ODataNestedResourceInfoWrapper parentNestedResourceInfo = (ODataNestedResourceInfoWrapper)itemsStack.Peek();
-                Contract.Assert(parentNestedResourceInfo != null, "this has to be an inner resource set. inner resource sets always have a nested resource info.");
-                Contract.Assert(parentNestedResourceInfo.NestedResourceInfo.IsCollection == true, "Only collection nested properties can contain resource set as their child.");
-                Contract.Assert(parentNestedResourceInfo.NestedItems.Count == 0, "Each nested property can contain only one resource set as its direct child.");
-                parentNestedResourceInfo.NestedItems.Add(resourceSetWrapper);
+                ODataItemWrapper peekWrapper = itemsStack.Peek();
+                if (peekWrapper is ODataNestedResourceInfoWrapper parentNestedResourceInfo)
+                {
+                    Contract.Assert(parentNestedResourceInfo.NestedResourceInfo.IsCollection == true, "Only collection nested properties can contain resource set as their child.");
+                    Contract.Assert(parentNestedResourceInfo.NestedItems.Count == 0, "Each nested property can contain only one resource set as its direct child.");
+                    parentNestedResourceInfo.NestedItems.Add(resourceSetWrapper);
+                }
+                else
+                {
+                    ODataResourceSetWrapper parentResourceSet = (ODataResourceSetWrapper)peekWrapper;
+                    parentResourceSet.Items.Add(resourceSetWrapper);
+                }
             }
             else
             {
--- a/src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataResourceSetWrapper.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataResourceSetWrapper.cs
@@ -22,6 +22,7 @@ namespace Microsoft.AspNetCore.OData.Formatter.Wrapper
         public ODataResourceSetWrapper(ODataResourceSet resourceSet)
         {
             Resources = new List<ODataResourceWrapper>();
+            Items = new List<ODataItemWrapper>();
             ResourceSet = resourceSet;
         }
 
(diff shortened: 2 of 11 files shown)


src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataResourceSetWrapper.cs
```diff
@@ -35,5 +36,13 @@ public ODataResourceSetWrapper(ODataResourceSet resourceSet)
...
         /// </summary>
         public IList<ODataResourceWrapper> Resources { get; }
+
+        /// <summary>
+        /// Gets the nested items of this ResourceSet.
+        /// Since we have 'Resources' to contain ODataResource items in the collection (we have to keep it avoid breaking changes).
+        /// This list is used to hold other items, for example a primitive or a collection, etc.
+        /// In the next major release, we should combine 'Resources' and 'Items' together.
+        /// </summary>
+        public IList<ODataItemWrapper> Items { get; }
```
corranrogue9, line 46: Is there supposed to be code somewhere that reads the `Items` property to actually populate the CLR type?

src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataReaderExtensions.cs
```diff
@@ -301,11 +310,18 @@ private static void ReadResourceSet(ODataReader reader, Stack<ODataItemWrapper>
             ODataResourceSetWrapper resourceSetWrapper = new ODataResourceSetWrapper(resourceSet);
             if (itemsStack.Count > 0)
             {
-                ODataNestedResourceInfoWrapper parentNestedResourceInfo = (ODataNestedResourceInfoWrapper)itemsStack.Peek();
-                Contract.Assert(parentNestedResourceInfo != null, "this has to be an inner resource set. inner resource sets always have a nested resource info.");
```
habbes, line 305: Can `itemStack` contain `null` entries? Or does this assertion (and others like it in this class) make the assumption that the cast will return `null` if the conversion fails? If the item at the top of the stack is not the expected type, an `InvalidCastException` will be thrown. If we expect null, we should consider using `itemStack.Peek() as ODataNestedResourceInfoWrapper`.
But I don't mind throwing the exception if that was by design.

habbes: I know this doesn't apply to the new code change, but I've seen it in other places in this class and thought it's worth pointing out.

src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataReaderExtensions.cs
```diff
@@ -301,11 +310,18 @@ private static void ReadResourceSet(ODataReader reader, Stack<ODataItemWrapper>
...
-                Contract.Assert(parentNestedResourceInfo.NestedResourceInfo.IsCollection == true, "Only collection nested properties can contain resource set as their child.");
-                Contract.Assert(parentNestedResourceInfo.NestedItems.Count == 0, "Each nested property can contain only one resource set as its direct child.");
-                parentNestedResourceInfo.NestedItems.Add(resourceSetWrapper);
+                ODataItemWrapper peekWrapper = itemsStack.Peek();
```
habbes, line 313: nitpick:
```suggestion
                ODataItemWrapper peekedWrapper = itemsStack.Peek();
```

src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataReaderExtensions.cs
```diff
@@ -301,11 +310,18 @@ private static void ReadResourceSet(ODataReader reader, Stack<ODataItemWrapper>
...
+                if (peekWrapper is ODataNestedResourceInfoWrapper parentNestedResourceInfo)
+                {
+                    Contract.Assert(parentNestedResourceInfo.NestedResourceInfo.IsCollection == true, "Only collection nested properties can contain resource set as their child.");
+                    Contract.Assert(parentNestedResourceInfo.NestedItems.Count == 0, "Each nested property can contain only one resource set as its direct child.");
+                    parentNestedResourceInfo.NestedItems.Add(resourceSetWrapper);
+                }
+                else
+                {
+                    ODataResourceSetWrapper parentResourceSet = (ODataResourceSetWrapper)peekWrapper;
+                    parentResourceSet.Items.Add(resourceSetWrapper);
```
habbes, line 323: Why do we assume that the parent is an `ODataResourceSetWrapper`? Is it because the itemStack.Count is > 0?

src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataResourceSetWrapper.cs
```diff
@@ -35,5 +36,13 @@ public ODataResourceSetWrapper(ODataResourceSet resourceSet)
         /// Resource set only contains resources.
         /// </summary>
         public IList<ODataResourceWrapper> Resources { get; }
+
+        /// <summary>
+        /// Gets the nested items of this ResourceSet.
+        /// Since we have 'Resources' to contain ODataResource items in the collection (we have to keep it avoid breaking changes).
+        /// This list is used to hold other items, for example a primitive or a collection, etc.
```
habbes, line 43: By this you mean that `Items` and `Resources` have no overlap? Or does `Items` also include `Resources`?

src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataResourceSetWrapper.cs
habbes: I also have a similar question:
1. Why do we need to make it public? Do we expect customers to use this?
2. If we make it public, do we expect customers be both read and update the collection?

test/Microsoft.AspNetCore.OData.Tests/Formatter/Wrapper/ODataReaderExtensionsTests.Untyped.cs
```diff
@@ -0,0 +1,450 @@
...
+                    { "null", null }
+                };
+
+                return data;
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(PrimitiveValueCases))]
+        public async Task ReadOpenResource_WithUntypedPrimitiveValue_WorksAsExpected(string value, object expect)
```
habbes, line 83: nitpick:
```suggestion
        public async Task ReadOpenResource_WithUntypedPrimitiveValue_WorksAsExpected(string value, object expected)
```

test/Microsoft.AspNetCore.OData.Tests/Formatter/Wrapper/ODataReaderExtensionsTests.Untyped.cs
```diff
@@ -0,0 +1,450 @@
...
+                "null," +
+                "{\"Key1\": \"Value1\", \"Key2\": true}," +
+                "\"A string item\"," +
+                "42" +
+            "]";
+
+            string payload =
+            "{" +
+                "\"ID\": 17," +
+                $"\"Data\": {collectionValue}," +
```
habbes, line 201: Is `Data` a complex property which is also based on an open complex type? Or is `Data` one of the dynamic properties of the top-level resource?

test/Microsoft.AspNetCore.OData.Tests/Formatter/Wrapper/ODataReaderExtensionsTests.Untyped.cs
```diff
@@ -0,0 +1,450 @@
...
+        public async Task ReadOpenResource_WithDeepUntypedCollectionValue_WorksAsExpected()
+        {
+            // Arrange
+            const int Depth = 10;
+            string collectionValue = "[42]";
+            for (int i = 0; i < Depth; i++)
+            {
+                collectionValue = $"[42, {collectionValue}]";
+            }
+            // [42, [42, [42, [42, [42, [42, [42, [42, [42, [42, [42]]]]]]]]]]]
```
habbes, line 383: Curious whether you also have a test with deep nesting of objects/complex types (at least 2 levels) which may also contain collection values?

src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataResourceSetWrapper.cs
xuzhg: It depends on the question:
Does the items order in a collection matter?
1) If it does matter, I will make Items to include all in 'Resources'.
2) If it doesn't matter, I will sperate them.

So, the idea is to replace 
```diff
-public ILIist<ODataResourceWrapper> Resources {get;}
+public ILIist<ODataItemWrapper> Resources {get;}
```

src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataResourceSetWrapper.cs
xuzhg: Since 'Resources' is open/public, we can't stop customers adding new resources into that.
Same reason, we can't stop customers to modify the 'Items'.

src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataReaderExtensions.cs
habbes: The guarantee makes sense. My comment was more on the fact that `Contact.Assert` verifies that the `parentNestedResourceInfo` is not `null`. However, if the cast would have failed, `parentNestedResourceInfo` would not have been set to null, an exception would have been thrown, and therefore the `Contact.Assert` would not have been reached.

src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataResourceSetWrapper.cs
habbes: What about the current implementation? Does the order matter? Does `Items` in the current implementation of this PR include resources?

src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataResourceSetWrapper.cs
habbes: `Items` is being introduced in this PR, so it doesn't exist in existing code. So my question was whether we need to expose it to customers and why.

src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataResourceSetWrapper.cs
xuzhg: The PR doesn't include 'Resources' items into "Items".
In my opinion, the order doesn't matter. But we can change it anytime if the order does matter.

src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataReaderExtensions.cs
xuzhg: itemsStack.Peek() could return 'null', I think.
ODataNestedResourceInfoWrapper parentNestedResourceInfo = (ODataNestedResourceInfoWrapper)null;

is valid. We can sync it more, but it's no scope of this PR. :)

test/Microsoft.AspNetCore.OData.Tests/Formatter/Wrapper/ODataReaderExtensionsTests.Untyped.cs
xuzhg: 'Data' is a declared property, whose type is 'Edm.Untyped'. 
```xml
<EntityType Name="Person" IsOpen="True" >
                <Property Name="ID" Type="Edm.Int32" />
                <Property Name="Name" Type="Edm.String" />
                <Property Name="Data" Type="Edm.Untyped" />
                <Property Name="Infos" Type="Collection(NS.Info)" />
             </EntityType>
```

test/Microsoft.AspNetCore.OData.Tests/Formatter/Wrapper/ODataReaderExtensionsTests.Untyped.cs
xuzhg: What do you mean?

this is test about the collection in collection directly.
The other is a test a collection of a resource property value and this resource is also a item of a collection.

Different scenarios.

src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataResourceSetWrapper.cs
xuzhg: Customer may override the 'CreateResourceSet' method in ODataResourceSetDeserializer to access the 'Items'.

test/Microsoft.AspNetCore.OData.Tests/Formatter/Wrapper/ODataReaderExtensionsTests.Untyped.cs
xuzhg: Thanks. let me keep it unchanged.

src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataResourceSetWrapper.cs
habbes: I see.

habbes (approved):

PR #906, opened 2023-05-03
Make CreatedODataResult inherit from ObjectResult

### LINKED ISSUE #897: AutoExpand not applied to Created(object) response.
**Assemblies affected**
`Microsoft.AspNetCore.OData` Version="8.1.1"

**Describe the bug**
`Created` method in `ODataController` does not seem to honor `AutoExpand` attribute.

**Reproduce steps**
If you POST an entity with a navigation that has the `AutoExpand` attribute and return the modified entity in the controller, the resulting json does not have the navigation populated.

**Data Model**

```csharp
public class Order
{
    public int Id { get; set; }
    public decimal Amount { get; set; }
}

public class Customer
{
    public int Id { get; set; }
    public string Name { get; set; } = string.Empty;

    [AutoExpand]
    public Order Order { get; set; } = null!;
}
```

**EDM (CSDL) Model**

```xml
<?xml version="1.0" encoding="utf-8"?>
<edmx:Edmx Version="4.0" xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx">
  <edmx:DataServices>
    <Schema Namespace="Example" xmlns="http://docs.oasis-open.org/odata/ns/edm">
      <EntityType Name="Customer">
        <Key>
          <PropertyRef Name="Id" />
        </Key>
        <Property Name="Id" Type="Edm.Int32" Nullable="false" />
        <Property Name="Name" Type="Edm.String" Nullable="false" />
        <NavigationProperty Name="Order" Type="Example.Order" Nullable="false" />
      </EntityType>
      <EntityType Name="Order">
        <Key>
          <PropertyRef Name="Id" />
        </Key>
        <Property Name="Id" Type="Edm.Int32" Nullable="false" />
        <Property Name="Amount" Type="Edm.Decimal" Nullable="false" Scale="Variable" />
      </EntityType>
    </Schema>
    <Schema Namespace="Default" xmlns="http://docs.oasis-open.org/odata/ns/edm">
      <EntityContainer Name="Container">
        <EntitySet Name="Customers" EntityType="Example.Customer" />
      </EntityContainer>
    </Schema>
  </edmx:DataServices>
</edmx:Edmx>
```

model built and registered as:

```csharp
ODataConventionModelBuilder modelBuilder = new();
modelBuilder.EntitySet<Customer>("Customers");

builder.Services
    .AddControllers()
    .AddOData(
        options =>
        {
            options.Select().Filter().OrderBy().Expand().SetMaxTop(100).Count().SkipToken();
            options.AddRouteComponents(
                "odata",
                modelBuilder.GetEdmModel());
        });
```

controller:

```csharp
public class CustomersController : ODataController
{
    [HttpPost]
    [EnableQuery(AllowedQueryOptions = AllowedQueryOptions.Select)]
    public ActionResult<Customer> PostAsync([FromBody] Customer customer)
    {
        return Created(customer);
    }
}
```

**Request/Response**

`[POST] https://localhost:44341/odata/Customers`

`Request`

```json
{
    "Name": "Customer name",
    "Order": {
        "Id": 1,
        "Amount": 10
    }
}
```

`Response`

```json
{
    "@odata.context": "https://localhost:44341/odata/$metadata#Customers/$entity",
    "Id": 0,
    "Name": "Customer name"
}
```

**Expected behavior**
I expect the resulting payload to have `Order` expanded.

**Additional context**

Please see similar issue in `WebApi`.

https://github.com/OData/WebApi/pull/2765

If I make the same change to this library, then the resulting json has `Order` properly expanded.

KenitoInc:
Fixes #897 
Port from https://github.com/OData/WebApi/pull/2765

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Results/CreatedODataResult.cs
+++ b/src/Microsoft.AspNetCore.OData/Results/CreatedODataResult.cs
@@ -22,13 +22,14 @@ namespace Microsoft.AspNetCore.OData.Results
     /// <remarks>This action result handles content negotiation and the HTTP prefer header. It generates a location
     /// header containing the edit link of the created entity and, if response has status code: NoContent, also
     /// generates an OData-EntityId header.</remarks>
-    public class CreatedODataResult<T> : ActionResult
+    public class CreatedODataResult<T> : ObjectResult
     {
         /// <summary>
         /// Initializes a new instance of the <see cref="CreatedODataResult{T}"/> class.
         /// </summary>
         /// <param name="entity">The created entity.</param>
         public CreatedODataResult(T entity)
+            : base(entity)
         {
             Entity = entity ?? throw Error.ArgumentNull(nameof(entity));
         }
--- a/test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
+++ b/test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
@@ -436,13 +436,14 @@ namespace Microsoft.AspNetCore.OData.E2E.Tests.AutoExpand
                     'Friend':{'Id': 99, 'HomeAddress': {'Street': 'Street 1', 'City': 'City 1'}}
                      }";
 
+            HttpClient client = CreateClient();
+
             StringContent stringContent = new StringContent(content: content, encoding: Encoding.UTF8, mediaType: "application/json");
 
             var expectedOrder = "Order\":{\"Id\":1,\"Choice\":{\"Id\":101,\"Amount\":10.0}}";
             var expectedFriend = "Friend\":{\"Id\":99,\"HomeAddress\":{\"Street\":\"Street 1\",\"City\":\"City 1\",\"CountryOrRegion\":null},\"Order\":null,\"Friend\":null}";
 
             //Act & Assert
-            using (HttpClient client = CreateClient())
             using (HttpRequestMessage requestForPost = new HttpRequestMessage(HttpMethod.Post, requestUri) { Content = stringContent })
             using (HttpResponseMessage response = await client.SendAsync(requestForPost))
             {
--- a/test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
+++ b/test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
@@ -428,7 +428,6 @@ namespace Microsoft.AspNetCore.OData.E2E.Tests.AutoExpand
         public async Task PostCustomer_AutoExpandNavigationProperties()
         {
             //Arrange
-
             string requestUri = "autoexpand/Customers";
 
             var content = @"{
--- a/test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
+++ b/test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Text;
 using System.Threading.Tasks;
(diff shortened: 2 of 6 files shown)


src/Microsoft.AspNetCore.OData/Results/CreatedODataResult.cs
```diff
@@ -22,13 +22,14 @@ namespace Microsoft.AspNetCore.OData.Results
     /// <remarks>This action result handles content negotiation and the HTTP prefer header. It generates a location
     /// header containing the edit link of the created entity and, if response has status code: NoContent, also
     /// generates an OData-EntityId header.</remarks>
-    public class CreatedODataResult<T> : ActionResult
+    public class CreatedODataResult<T> : ObjectResult
```
xuzhg, line 25: same question, Is it a breaking change?

ElizabethOkerio (approved):

src/Microsoft.AspNetCore.OData/Results/CreatedODataResult.cs
KenitoInc: No it isn't.
`ObjectResult` inherits from `ActionResult`

ElizabethOkerio (dismissed):

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
```diff
@@ -422,5 +423,37 @@ public async Task NonDefaultMaxExpansionDepthAppliesToAutoExpand(string entitySe
...
+            //Arrange
+            string requestUri = "autoexpand/Customers";
+
+            var content = @"{
+                    'Id':88,
+                    'Order':{ 'Id':1, 'Choice' : {'Id': 101, 'Amount': 10}},
+                    'Friend':{'Id': 99, 'HomeAddress': {'Street': 'Street 1', 'City': 'City 1'}}
+                     }";
+
+            var requestForPost = new HttpRequestMessage(HttpMethod.Post, requestUri);
```
corranrogue9, line 439: It would be good to dispose this and `stringContent`.

corranrogue9 (approved):

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
KenitoInc: Disposing `HttpRequestMessage` will dispose the `Content`.

See [this](https://github.com/microsoft/referencesource/blob/5697c29004a34d80acdaf5742d7e699022c64ecd/System/net/System/Net/Http/HttpRequestMessage.cs#L200)

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
```diff
@@ -422,5 +423,34 @@ public async Task NonDefaultMaxExpansionDepthAppliesToAutoExpand(string entitySe
...
+                    'Friend':{'Id': 99, 'HomeAddress': {'Street': 'Street 1', 'City': 'City 1'}}
+                     }";
+
+            StringContent stringContent = new StringContent(content: content, encoding: Encoding.UTF8, mediaType: "application/json");
+
+            var expectedOrder = "Order\":{\"Id\":1,\"Choice\":{\"Id\":101,\"Amount\":10.0}}";
+            var expectedFriend = "Friend\":{\"Id\":99,\"HomeAddress\":{\"Street\":\"Street 1\",\"City\":\"City 1\",\"CountryOrRegion\":null},\"Order\":null,\"Friend\":null}";
+
+            //Act & Assert
+            using (HttpClient client = CreateClient())
```
corranrogue9, line 445: Sorry, I think I commented on the wrong line number before. `stringContent` is the other disposable. I *don't* think you should dispose the client, because I think the client is shared across all of the test methods.

KenitoInc: Fixed

habbes (approved):

PR #909, opened 2023-05-04
Allow reading primitive, collection in a collection

xuzhg:
fixes #905

A different design.


xuzhg:
@julealgon also want to get your input about which one is good:

this PR vs #905

PR #912, opened 2023-05-08
fix: Matching property names based on ignoring case

### LINKED ISSUE #911: `CreateRefTo` / `DeleteRefTo` Property Casing issues
We are currently facing an issue with the latest versions (8.1.x), with unrecognized endpoints of Type `CreateRefTo{navigationProperty}` & `DeleteRefTo{navigationProperty}`.

When we define a method with the signature `public async ValueTask<IActionResult> DeleteRefToStandardDiscountAsync([FromRoute] int key, CancellationToken cancellationToken) { }`, the endpoint is not recognized. When we lower the first letter of the navigation property (`public async ValueTask<IActionResult> DeleteRefTostandardDiscountAsync([FromRoute] int key, CancellationToken cancellationToken) { }`) the method is fully recognized and functional. This is a strange behaviour, which we tracked down to
https://github.com/OData/AspNetCoreOData/blob/9225d1eba1601bf1470b7e611cdd595fa479dbfc/src/Microsoft.AspNetCore.OData/Routing/Conventions/RefRoutingConvention.cs#L104

We provide a possible fix in the next minutes.

samtrion:
Fixed #911

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Routing/Conventions/RefRoutingConvention.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/Conventions/RefRoutingConvention.cs
@@ -101,7 +101,7 @@ namespace Microsoft.AspNetCore.OData.Routing.Conventions
 
             // Find the navigation property if have
             IEdmNavigationProperty navigationProperty = null;
-            navigationProperty = declaringType.DeclaredNavigationProperties().FirstOrDefault(p => p.Name == property);
+            navigationProperty = declaringType.DeclaredNavigationProperties().FirstOrDefault(p => p.Name.Equals(property, StringComparison.OrdinalIgnoreCase));
 
             if (navigationProperty == null)
             {
--- a/src/Microsoft.AspNetCore.OData/Routing/Conventions/RefRoutingConvention.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/Conventions/RefRoutingConvention.cs
@@ -101,7 +101,18 @@ namespace Microsoft.AspNetCore.OData.Routing.Conventions
 
             // Find the navigation property if have
             IEdmNavigationProperty navigationProperty = null;
-            navigationProperty = declaringType.DeclaredNavigationProperties().FirstOrDefault(p => p.Name.Equals(property, StringComparison.OrdinalIgnoreCase));
+            navigationProperty = declaringType.DeclaredNavigationProperties().FirstOrDefault(p => p.Name.Equals(property, StringComparison.Ordinal));
+
+            if (navigationProperty is null && (context.Options?.RouteOptions?.EnablePropertyNameCaseInsensitive ?? false))
+            {
+                IEdmNavigationProperty[] navigationProperties = declaringType.DeclaredNavigationProperties().Where(p => p.Name.Equals(property, StringComparison.OrdinalIgnoreCase)).ToArray();
+                if (navigationProperties.Length > 1)
+                {
+                    throw Error.InvalidOperation(SRResources.UnableToIdentifyUniqueProperty, property);
+                }
+
+                navigationProperty = navigationProperties.FirstOrDefault();
+            }
 
             if (navigationProperty == null)
             {
--- a/src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.xml
+++ b/src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.xml
@@ -7402,6 +7402,11 @@
               Looks up a localized string similar to The OData formatter must be able to determine the metadata URL for a request in order to serialize the response..
             </summary>
         </member>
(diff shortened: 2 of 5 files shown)


samtrion:
@xuzhg Can you please review this PR?

gathogojr:
@samtrion Thanks for your contribution. Could you please a a test or tests to verify that it works correctly after the fix

samtrion:
@gathogojr done

src/Microsoft.AspNetCore.OData/Routing/Conventions/RefRoutingConvention.cs
```diff
@@ -101,7 +101,7 @@ public bool AppliesToAction(ODataControllerActionContext context)
 
             // Find the navigation property if have
             IEdmNavigationProperty navigationProperty = null;
-            navigationProperty = declaringType.DeclaredNavigationProperties().FirstOrDefault(p => p.Name == property);
+            navigationProperty = declaringType.DeclaredNavigationProperties().FirstOrDefault(p => p.Name.Equals(property, StringComparison.OrdinalIgnoreCase));
```
gathogojr, line 104: @samtrion In theory, nothing prevents someone from declaring two properties that differ only by type. For example,
```csharp
public class Order
{
    public int Id { get; set; }
}

public class Customer
{
    public int Id { get; set; }
    public List<Order> Orders { get; set; }
    public List<Order> orders { get; set; }
}
```
The metadata for the corresponding Edm model would look as follows:
```xml
<?xml version="1.0" encoding="utf-8"?>
<edmx:Edmx Version="4.0" xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx">
    <edmx:DataServices>
        <Schema Namespace="Ns.Models" xmlns="http://docs.oasis-open.org/odata/ns/edm">
            <EntityType Name="Order">
                <Key>
                    <PropertyRef Name="Id" />
                </Key>
                <Property Name="Id" Type="Edm.Int32" Nullable="false" />
            </EntityType>
            <EntityType Name="Customer">
                <Key>
                    <PropertyRef Name="Id" />
                </Key>
                <Property Name="Id" Type="Edm.Int32" Nullable="false" />
                <NavigationProperty Name="Orders" Type="Collection(Ns.Models.Order)" />
                <NavigationProperty Name="orders" Type="Collection(Ns.Models.Order)" />
            </EntityType>
        </Schema>
        <Schema Namespace="Default" xmlns="http://docs.oasis-open.org/odata/ns/edm">
            <EntityContainer Name="Container">
                <EntitySet Name="Orders" EntityType="Ns.Models.Order" />
                <EntitySet Name="Customers" EntityType="Ns.Models.Customer">
                    <NavigationPropertyBinding Path="orders" Target="Orders" />
                    <NavigationPropertyBinding Path="Orders" Target="Orders" />
                </EntitySet>
            </EntityContainer>
        </Schema>
    </edmx:DataServices>
</edmx:Edmx>
```
Is it advisable to have property names that differ only by case? Probably not! But is it valid? Most certainly!

For this reason, your change here could cause all manner of problems to customers.
To mitigate that, we need to first check whether the relevant option (`context.Options?.RouteOptions?.EnablePropertyNameCaseInsensitive` I believe) is set to true.
- If no, we should only do a case-sensitive comparison.
- If yes, we should do a case-sensitive comparison first, and if we don't find a match we do a case-insensitive comparison. If we do a case-insensitive comparison and find multiple matches, we should throw an exception.

I hope that makes sense.

samtrion: I agree, with your approach it would be much "safer" especially against cases like you mentioned. How realistic this case is is almost irrelevant here, since ignorance is usually the driver of such phenomena.

I'm going to take some time and expand the PR to include that.

src/Microsoft.AspNetCore.OData/Routing/Conventions/RefRoutingConvention.cs
```diff
@@ -101,7 +101,18 @@ public bool AppliesToAction(ODataControllerActionContext context)
...
+
+            if (navigationProperty is null && (context.Options?.RouteOptions?.EnablePropertyNameCaseInsensitive ?? false))
+            {
+                IEdmNavigationProperty[] navigationProperties = declaringType.DeclaredNavigationProperties().Where(p => p.Name.Equals(property, StringComparison.OrdinalIgnoreCase)).ToArray();
+                if (navigationProperties.Length > 1)
+                {
+                    throw Error.InvalidOperation(SRResources.UnableToIdentifyUniqueProperty, property);
+                }
+
+                navigationProperty = navigationProperties.FirstOrDefault();
```
gathogojr, line 114: nit:
```suggestion
                navigationProperty = navigationProperties.SingleOrDefault();
```

gathogojr (approved):

ElizabethOkerio (approved):

gathogojr:
@samtrion Thank you for your contribution

PR #914, opened 2023-05-10
add datamodificationexception class

ElizabethOkerio:


src/Microsoft.AspNetCore.OData/IODataInstanceAnnotationContainer.cs
```diff
@@ -0,0 +1,61 @@
...
+
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.OData
+{
+    /// <summary>
+    /// Identifies a container for holding instance annotations. A default implementation is provided.
+    /// Customers can have their own implementation.
+    /// </summary>
+    public interface IODataInstanceAnnotationContainer
```
xuzhg, line 16: Actually, I don't like this interface and the methods defined.

src/Microsoft.AspNetCore.OData/Deltas/IDeltaSetItem.cs
```diff
@@ -16,5 +16,10 @@ public interface IDeltaSetItem
         /// Gets the delta item kind.
         /// </summary>
         DeltaItemKind Kind { get; }
+
+        /// <summary>
+        /// Gets or sets the annotation container to hold transient instance annotations.
+        /// </summary>
+        IODataInstanceAnnotationContainer TransientInstanceAnnotationContainer { get; set; }
```
xuzhg, line 23: where and how can customer set this interface?

src/Microsoft.AspNetCore.OData/Deltas/Delta.cs
```diff
@@ -21,6 +21,9 @@ public abstract class Delta : DynamicObject, IDelta
         /// </summary>
         public abstract DeltaItemKind Kind { get; }
 
+        /// <inheritdoc/>
+        public IODataInstanceAnnotationContainer TransientInstanceAnnotationContainer { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
```
xuzhg, line 25: Can we mark it abstract and remove NotImplementedException()?
or at least the exception contains the message

src/Microsoft.AspNetCore.OData/Deltas/IDeltaSetItem.cs
ElizabethOkerio: when they are creating the response. If there are operations in the `DeltaSet` that failed then you need to create DataModificationExceptions. This is the time customers will need this interface.

src/Microsoft.AspNetCore.OData/IODataInstanceAnnotationContainer.cs
ElizabethOkerio: What do you suggest it be  replaced with?

PR #926, opened 2023-05-16
Fix invalid partial json generated when serverside paging is applied in multi-level containment scenarios

gathogojr:
This pull request ports [this 7.x fix](https://github.com/OData/WebApi/pull/2771) to 8.x

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Formatter/LinkGenerationHelpers.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/LinkGenerationHelpers.cs
@@ -8,6 +8,7 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Diagnostics;
 using System.Diagnostics.Contracts;
 using System.Linq;
 using Microsoft.AspNetCore.OData.Edm;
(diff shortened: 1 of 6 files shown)


ElizabethOkerio (approved):

src/Microsoft.AspNetCore.OData/Formatter/LinkGenerationHelpers.cs
```diff
@@ -69,7 +70,16 @@ public static Uri GenerateSelfLink(this ResourceContext resourceContext, bool in
                 throw Error.ArgumentNull(nameof(resourceContext));
             }
 
-            IList<ODataPathSegment> navigationPathSegments = resourceContext.GenerateBaseODataPathSegments();
+            IList<ODataPathSegment> navigationPathSegments;
+            if (resourceContext.NavigationSource is IEdmContainedEntitySet &&
+                resourceContext.NavigationSource != resourceContext.SerializerContext.Path.NavigationSource())
```
habbes, line 75: What does this second condition ensure?

KenitoInc (approved):

src/Microsoft.AspNetCore.OData/Formatter/LinkGenerationHelpers.cs
gathogojr: @habbes When `resourceContext.NavigationSource is IEdmContainedEntitySet` and `resourceContext.NavigationSource` is equal `resourceContext.SerializationContext.Path.NavigationSource()`, the `GenerateBaseODataPathSegments()` is able singly handle that scenario

Issue #918, opened 2023-05-12
OData Collection requests with $apply (aggregation) still failing when Client/Server side pagination enabled.


kairgrw:
Looks like, it's a long story from issue #420 which appears to be partially fixed but still not working for Client/Server side pagination enabled.

Our project settings:

- Framework: .NET 7.0
- EF Core: 7.0
- Language: C# 11.0
- Microsoft.AspNetCore.OData: 8.0.12 (doesn't matter, still is not working since 8.0.5)

It's reproducible on [ODataRoutingSample](https://github.com/OData/AspNetCoreOData/tree/main/sample/ODataRoutingSample): ASP.NET Core OData sample project.

Requests:
1. Orders: http://localhost:5000/v21/Orders
2. Orders with just $apply (filter + aggregate): http://localhost:5000/v21/Orders?$apply=filter((contains(title,'i')))/aggregate(id with countdistinct as f0)
3. Orders with $apply (filter + groupBy + aggregate) and $top: http://localhost:5000/v21/Orders?$apply=filter((contains(title,'i')))/groupby((title),aggregate(id with countdistinct as f0))&$top=5
4. Orders with $apply (filter + aggregate) and $top: http://localhost:5000/v21/Orders?$apply=filter((contains(title,'i')))/aggregate(id with countdistinct as f0)&$top=5

Request 1 works as expected:
```json
{"@odata.context":"http://localhost:5000/v21/$metadata#Orders","value":[{"Id":1,"Title":"Goods"},{"Id":2,"Title":"Magazine"},{"Id":3,"Title":"Fiction"}]}
```
Request 2 works as expected:
```json
{"@odata.context":"http://localhost:5000/v21/$metadata#Orders(f0)","value":[{"@odata.id":null,"f0":2}]}
```
Request 3 works as expected:
```json
{"@odata.context":"http://localhost:5000/v21/$metadata#Orders(Title,f0)","value":[{"@odata.id":null,"Title":"Fiction","f0":1},{"@odata.id":null,"Title":"Magazine","f0":1}]}
```
Request 4 fails with error:
```
[...]

PR #941, opened 2023-05-26
fix $apply without groupby

ElizabethOkerio:
This PR fixes https://github.com/OData/AspNetCoreOData/issues/918

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinderContext.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinderContext.cs
@@ -55,8 +55,8 @@ namespace Microsoft.AspNetCore.OData.Query.Expressions
 
             ElementType = Model.GetEdmTypeReference(ElementClrType)?.Definition;
 
-            // Check if element type is null and not of AggregationWrapper type.
-            if (ElementType == null && ElementClrType != typeof(AggregationWrapper))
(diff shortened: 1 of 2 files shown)


xuzhg (dismissed):

src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinderContext.cs
```diff
@@ -55,8 +55,8 @@ public QueryBinderContext(IEdmModel model, ODataQuerySettings querySettings, Typ
 
             ElementType = Model.GetEdmTypeReference(ElementClrType)?.Definition;
 
-            // Check if element type is null and not of AggregationWrapper type.
-            if (ElementType == null && ElementClrType != typeof(AggregationWrapper))
+            // Check if element type is null and not of AggregationWrapper type and not of NoGroupByAggregationWrapper type.
+            if (ElementType == null && ElementClrType != typeof(AggregationWrapper) && ElementClrType != typeof(NoGroupByAggregationWrapper))
```
KenitoInc, line 59: Explain this change

ElizabethOkerio: When we have $apply without a groupby we create a `NoGroupByAggregationWrapper.`. This check checks for the ElementClrType and throw an exception if the type is not of `AggregationWrapper`. It should also not throw an exception if the type is of `NoGroupByAggregationWrapper`.

Issue #616, opened 2022-06-08
SkipToken returning invalid next link when combined with select


karayanni:
Hi team,
we are using AspNetCore.Odata 8.0.8 for a simple get api with paging.
the Default paging uses Skip - which is causing us regressions with EFC (inefficient query)

in order to improve the performance, we are trying to use SkipToken for generating the next page link instead.

we added the skip token this way:

```cs
AddOData(opt =>
                {
                    var edmModel = GetEdmModel();
                    opt.AddRouteComponents("api", edmModel, ConfigureODataServices).SkipToken();
                    opt.RouteOptions.EnableNonParenthesisForEmptyParameterFunction = true;
                });
```
this works perfectly when sending a simple get request, however, when the request has a select filter
for example: http://localhost:1059/api/Machines?$select=id

the next link is: "@odata.nextLink": "https://localhost:1059/api/Machines?$select=id&$skiptoken=id-null"

can you help in investigating this bug?

Best,
Nader.

xuzhg:
@karayanni  the skiptoken token is calculated using the last 'key' value. The last 'key' value is 'null'? What is your response payload look like?

Would you please share your repro?

By the way, we support to customize/inject SkipTokenHandler, if the default one can't meet your requirement, maybe consider to create your handler.

karayanni:
Hi @xuzhg sorry for the alte response,

The last key is not null, however it is not included in the response.
for example the customer requested api/Students$select=Grade when the key is for is the students ID, so the response doesn't have the key but the entities themselves do have keys, all of them

and we can see that when we don't use the select=Grade the skiptoken works perfectly, yet when we add the selet filter we get the provided above next link - "https://localhost:1059/api/Machines?$select=id&$skiptoken=id-null"

karayanni:
@xuzhg friendly ping

karayanni:
@xuzhg 

Hi Sam, I'm debugging the Odata SkipToken flow locally, And I found an interesting behavior you might want to hear about

specifically, in the GenerateSkipTokenValue funciton -

in the code, we see this line
``` cs
(1)       IEdmStructuredObject obj = lastMember as IEdmStructuredObject; 
```

and if we are able to parse the object as EDM structured (obj != null), we read the properties by:

``` cs
(2)       obj.TryGetPropertyValue(clrPropertyName, out value)
```
if we are not (obj == null) we read the property from the original object with reflection:
``` cs
[...]

PR #947, opened 2023-05-31
Potential fix for issue #616

orty:
Based on issue #616 which I also encountered, I would like to propose this fix.

As I am not completely aware of side-effects that using only `edmProperty.Name` could have, I relied on a safe way to rollback to the EDM property name instead of the CLR one if the latter is not found.

https://github.com/OData/AspNetCoreOData/blob/10a246671b6312fd234151c47cc91787da5cb171/src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs#LL136C1-L145C18

I tested this change with my current project, but I suspect that some unit tests might have to be "doubled" with a `builder.EnableLowerCamelCase()` scenario to check for regressions on property retrieval when using a lowercased convention.

edit: Also tested along with the `[DataContract]` and `[DataMember]` attributes which create aliases for properties.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
@@ -135,12 +135,7 @@ namespace Microsoft.AspNetCore.OData.Query
                 bool islast = count == lastIndex;
                 if (obj != null)
                 {
-                    //if (!obj.TryGetPropertyValue(
-                    //        clrPropertyName,
-                    //        out value))
-                    //{
-                        obj.TryGetPropertyValue(edmProperty.Name, out value);
-                    //}
+                    obj.TryGetPropertyValue(edmProperty.Name, out value);
                 }
                 else
                 {
--- a/src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
@@ -133,18 +133,18 @@ namespace Microsoft.AspNetCore.OData.Query
             foreach (IEdmProperty edmProperty in propertiesForSkipToken)
             {
                 bool islast = count == lastIndex;
-                string clrPropertyName = model.GetClrPropertyName(edmProperty);
                 if (obj != null)
                 {
-                    if (!obj.TryGetPropertyValue(
-                            clrPropertyName,
-                            out value))
-                    {
+                    //if (!obj.TryGetPropertyValue(
+                    //        clrPropertyName,
+                    //        out value))
+                    //{
                         obj.TryGetPropertyValue(edmProperty.Name, out value);
-                    }
+                    //}
                 }
                 else
                 {
+                    string clrPropertyName = model.GetClrPropertyName(edmProperty);
                     value = lastMember.GetType().GetProperty(clrPropertyName).GetValue(lastMember);
                 }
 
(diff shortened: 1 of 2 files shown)


orty:
> @microsoft-github-policy-service agree

KenitoInc (changes requested):
Add tests

orty:
Hi @KenitoInc ,
I am a bit afraid of the amount of UT which would need an update to validate this scenario, but I will certainly give it a shot

test/Microsoft.AspNetCore.OData.Tests/Query/Query/DefaultSkipTokenHandlerTests.cs
```diff
@@ -74,6 +76,28 @@ public void GetNextPageLinkDefaultSkipTokenHandler_Returns_CorrectSkipTokenLink(
             Assert.Equal(expectedUri, actualUri);
         }
 
+        [Theory]
+        [InlineData("http://localhost/Customers(1)/Orders", "http://localhost/Customers(1)/Orders?$skiptoken=id-42")]
+        [InlineData("http://localhost/Customers?$expand=Orders", "http://localhost/Customers?$expand=Orders&$skiptoken=id-42")]
```
KenitoInc, line 81: Add a test with $select

orty: Did it. I did not found a way to validate the `$skiptoken` generation in a context where `EnsureStableOrdering = true`. It seemed to me that in that case, query options like the following:
`http://localhost/Customers?$select=id&$orderby=name` 
should generate this kind of token: `http://localhost/Customers?$select=id&$orderby=name&$skiptoken=id-42%2Cname-%27ZX%27`

But that is not happening in this test method. Maybe something to add to another test class ?

src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
```diff
@@ -136,7 +136,12 @@ internal static string GenerateSkipTokenValue(Object lastMember, IEdmModel model
                 string clrPropertyName = model.GetClrPropertyName(edmProperty);
                 if (obj != null)
                 {
-                    obj.TryGetPropertyValue(clrPropertyName, out value);
+                    if (!obj.TryGetPropertyValue(
+                            clrPropertyName,
+                            out value))
+                    {
+                        obj.TryGetPropertyValue(edmProperty.Name, out value);
+                    }
```
gathogojr, line 144: Thank you @orty for your pull request contribution.
I'd propose a simpler fix to this issue:
```suggestion
                    obj.TryGetPropertyValue(edmProperty.Name, out value);
```
I did a quick test and confirmed that it does the job.
If `lastMember` can be cast to `IEdmStructuredObject`, it should be okay to use the Edm property name to retrieve the property value from the object.
I didn't go to the extent of running all tests to confirm that nothing broke - let's try the change and see how it goes.

Note: We can also move the declaration for `clrPropertyName` to the `else` block since that's the only place where it'd be applicable/relevant.

orty: I did the change you suggested and I confirm it works like a charm. I went ahead and added some tests to check if an aliased model could also be handled like that, and discovered something.

In here :
https://github.com/OData/AspNetCoreOData/blob/118c2c175f49ebcf47fab64840bf4a8a675e8e8f/src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs#L431-L435

we should retrieve the proper underlying CLR type property name for the `IEdmProperty`, like it is done in the `QueryBinder.CreatePropertyAccessExpression` method, by adding `context.Model.GetClrPropertyName(property);`.

This is because, when the value-pairs dictionary has been populated, today it only works by chance that the `Expression` can be built here, in the `ApplyToCore` method
https://github.com/OData/AspNetCoreOData/blob/118c2c175f49ebcf47fab64840bf4a8a675e8e8f/src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs#L291-L296

It works with both PascalCased and LowerCamelCased propery names because of the permissivity of LINQ. When an alias is provided, it just throws an exception. With the next commit I will push this afternoon (it is 1PM here, gotta take a break ;)), the `Expression` is able to be built even with an alias.

I added all the subsequent unit tests for this change.

gathogojr (approved):
Thanks @orty for your contribution

orty:
> Thanks @orty for your contribution

You're welcome ! @KenitoInc, do you agree with the overall changes ?

KenitoInc (approved):

KenitoInc:
@orty Thanks for this PR

PR #961, opened 2023-06-19
fixes #931: Undeclared enum type value for untype value serialized as empty resource

### LINKED ISSUE #931: Undeclared enum type value for untype value serialized as empty resource
**Assemblies affected**

**Describe the bug**

**Reproduce steps**
If we have 
```C#
public enum UserType {...}
```

But this enum is not defined in OData model.

Suppose we use this C# enum to assign a value for a untyped property or dynamic property, the result is empty resource.
```C#
Person p = new Person
{
    Data = UserType.Admin  // System.Object Data {get;set;}
}
```
We get:
```json

"Data": {},
```

This is because, in "CreateUntypedPropertyValue", we can't get the Edm type for "UserType", then we consider it a resource.
So, goes to the resource serializer, owing that the enum type doesn't have public properties,
so, the final result is an empty resource.

Before the fix, the workaround is to call 'ToString'.

```C#
Person p = new Person
{
    Data = UserType.Admin.ToString()
}
```
**Data Model**

**EDM (CSDL) Model**
Please share your Edm model, for example, CSDL file.
You can send `$metadata` to get a CSDL XML content.

**Request/Response**

**Expected behavior**

**Screenshots**

**Additional context**

xuzhg:
fixes #931

If we have
```C#
public enum UserType {...}
```
But this enum is not defined in OData model.

Suppose we use this C# enum to assign a value for a untyped property or dynamic property, the result is empty resource.

Person p = new Person
{
    Data = UserType.Admin  // System.Object Data {get;set;}
}
We get:

"Data": {},

This PR is to fix this issue.

To Be discussed: Shall we write the full-name of enum member as string into OData payload? @mikepizzo

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
@@ -638,6 +638,7 @@ namespace Microsoft.AspNetCore.OData.Formatter.Serialization
 
             IEdmStructuredType structuredType = resourceContext.StructuredType;
             IEdmStructuredObject structuredObject = resourceContext.EdmObject;
+            ODataSerializerContext serializierContext = resourceContext.SerializerContext;
             object value;
             IDelta delta = structuredObject as IDelta;
             if (structuredObject is EdmUntypedObject untypedObject) // NO CLR, NO EDM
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -700,10 +701,25 @@ public virtual ODataResource CreateResource(SelectExpandNode selectExpandNode, R
...
                 if (edmTypeReference == null || edmTypeReference.IsStructuredOrUntyped())
                 {
+                    if (TypeHelper.IsEnum(propertyType))
+                    {
+                        // we don't have the Edm enum type in the model, let's write it as string.
+                        dynamicProperties.Add(new ODataProperty
+                        {
+                            Name = dynamicProperty.Key,
+
+                            // TBD: Shall we write the un-declared enum value as full-name string?
```
ElizabethOkerio, line 715: Was this determined?

ElizabethOkerio: Can it be written the same way it's written if it were in model?

xuzhg: So far, no. @mikepizzo

mikepizzo: Sorry; I thought we had discussed.

I am more concerned with whether or not we are able to round-trip the value than whether we use the full name or just the member name.

If we can round-trip using just the member name, then I suspect that most closely matches what the service implementor was trying to get. If we need to use the full name in order to set the value on write to be the enum value (as opposed to string), then I'm fine with that.

xuzhg: In the reading, we can't say whether it's Enum member, because its literal is a 'string'. It's weird to let the customer to specify a full name in the reading scenario.  So, Let me keep writing the name string to see the feedback from customers. Is it ok?

mikepizzo: Yes; so to recap, when writing we will write just the name ("Apple", not "Namespace.EnumTypeName.Apple").  When reading we will read "Apple" and set it as a string value.  If the service wants different behavior on deserialization, they can customize the deserialization logic to detect that "Apple" is actually Namespace.EnumType.Apple.

Once we start writing as "Apple", it would be a breaking change to start writing as "Namespace.EnumTypeName.Apple" (or vice-versa), so whatever we decide will be the default behavior at least for 8.x (we can add a flag to 8.x, or change behavior in 9.x or beyond, but we shouldn't unless it's a common ask that can't easily be overridden.)

xuzhg: I will update the comments and let me know if you have any other concerns for this PR?

Issue #725, opened 2022-10-22
Empty $filter in request causes service to return 500 due to ODataException


devlie:
**Assemblies affected**
ASP.NET Core OData 7.x and 8.x

**Describe the bug**
When user calls service with empty $filter, for e.g. `https://localhost/foo?$filter=` 

**Reproduce steps**
1. Create ASP.NET Core 3.1 REST API project.
2. Enable OData.
3. Call the service with empty $filter.

**Request/Response**
Request: `https://localhost/foo?$filter=` 
Response: service returns 500, with the model binding code throwing exception.

**Expected behavior**
Service returns 400. While I can handle the particular exception in my handler, it seems weird that I have to do that for simple user error like this.

**Additional context**

```csharp
fail: Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware[1]
      An unhandled exception has occurred while executing the request.
Microsoft.OData.ODataException: The value for OData query '$filter' cannot be empty.
   at Microsoft.AspNetCore.OData.Query.ODataQueryOptions.ThrowIfEmpty(String queryValue, String queryName)
   at Microsoft.AspNetCore.OData.Query.ODataQueryOptions.BuildQueryOptions(IDictionary`2 queryParameters)
   at Microsoft.AspNetCore.OData.Query.ODataQueryOptions.Initialize(ODataQueryContext context)
   at Microsoft.AspNetCore.OData.Query.ODataQueryOptions..ctor(ODataQueryContext context, HttpRequest request)
   at Microsoft.AspNetCore.OData.Query.ODataQueryOptions`1..ctor(ODataQueryContext context, HttpRequest request)
   at Microsoft.AspNetCore.OData.Query.ODataQueryParameterBindingAttribute.ODataQueryParameterBinding.CreateODataQueryOptions[T](ODataQueryContext context, HttpRequest request)
   at Microsoft.AspNetCore.OData.Query.ODataQueryParameterBindingAttribute.ODataQueryParameterBinding.BindModelAsync(ModelBindingContext bindingContext)
   at Microsoft.AspNetCore.Mvc.ModelBinding.Binders.BinderTypeModelBinder.BindModelAsync(ModelBindingContext bindingContext)
   at Microsoft.AspNetCore.Mvc.ModelBinding.ParameterBinder.BindModelAsync(ActionContext actionContext, IModelBinder modelBinder, IValueProvider valueProvider, ParameterDescriptor parameter, ModelMetadata metadata, Object value)
[...]

PR #964, opened 2023-06-21
return bad request for an empty filter

ElizabethOkerio:
This PR fixes this issue: https://github.com/OData/AspNetCoreOData/issues/725

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/EnableQueryAttribute.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/EnableQueryAttribute.cs
@@ -62,18 +62,19 @@ namespace Microsoft.AspNetCore.OData.Query
 
             actionExecutingContext.HttpContext.Items.TryAdd(nameof(RequestQueryData), requestQueryData);
 
-            ODataQueryOptions queryOptions = CreateQueryOptionsOnExecuting(actionExecutingContext);
-            if (queryOptions == null)
-            {
-                return; // skip validation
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.OData/Query/EnableQueryAttribute.cs
```diff
@@ -61,11 +61,24 @@ public override void OnActionExecuting(ActionExecutingContext actionExecutingCon
...
+                queryOptions = CreateQueryOptionsOnExecuting(actionExecutingContext);
 
-            ODataQueryOptions queryOptions = CreateQueryOptionsOnExecuting(actionExecutingContext);
-            if (queryOptions == null)
+                if (queryOptions == null)
+                {
+                    return; // skip validation
+                }
+            }
+            catch (InvalidOperationException e)
```
gathogojr, line 75: @ElizabethOkerio I think you should catch `ODataException` here and return whatever value is in the `Message` property. We know that `ODataException` can only be thrown by the OData libraries, but `InvalidOperationException` can be thrown by the any other dependencies.

ElizabethOkerio: @gathogojr I think when we are creating the ODataQueryOption we can return a `BadRequest` for all `InvalidOperationExceptions`.

gathogojr (approved):

PR #966, opened 2023-06-22
Allow suppressing request url root check during entity reference link validation

### LINKED ISSUE #965: AspNetCoreOData expects the @odata.bind addresses to begin with service url and thus breaking support for proxies
**Assemblies affected**
ASP.NET Core OData 8.x

**Describe the bug**
[ODataResourceDeserializer](https://github.com/OData/AspNetCoreOData/blob/9deb8b1ee2e26880dd8203fb5f3e347079f5b8ed/src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs#LL933C25-L933C25) expects that address in @odata.bind must start with a prefix same as the request url. However, this might not always be true. Our service has a proxy in the middle which replays the request, so the request url goes from abc.website.com/prefix/customer => localhost:21355/prefix/customer. Payload isn't edited in this case.

**Reproduce steps**
Have the @odata.bind url base differ from the url being invoked.

**Expected behavior**
The caller should either get option to skip verification of service root, or it should be default skipped.

RamjotSingh:
Fixes #965

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
@@ -17,6 +17,7 @@ using System.Runtime.Serialization;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.OData.Abstracts;
 using Microsoft.AspNetCore.OData.Deltas;
 using Microsoft.AspNetCore.OData.Edm;
 using Microsoft.AspNetCore.OData.Extensions;
@@ -25,6 +26,7 @@ using Microsoft.AspNetCore.OData.Formatter.Wrapper;
 using Microsoft.AspNetCore.OData.Routing;
 using Microsoft.AspNetCore.OData.Routing.Parser;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
 using Microsoft.OData;
 using Microsoft.OData.Edm;
 using Microsoft.OData.UriParser;
@@ -831,7 +833,7 @@ namespace Microsoft.AspNetCore.OData.Formatter.Deserialization
                 TypeName = edmPropertyType.FullName(),
             };
 
-            resource.Properties = CreateKeyProperties(refLink.EntityReferenceLink.Url, readContext) ?? Array.Empty<ODataProperty>();
+            resource.Properties = CreateKeyProperties(refLink.EntityReferenceLink.Url, readContext, isEntityReferenceLink: true) ?? Array.Empty<ODataProperty>();
             ODataResourceWrapper resourceWrapper = new ODataResourceWrapper(resource);
 
             foreach (ODataInstanceAnnotation instanceAnnotation in refLink.EntityReferenceLink.InstanceAnnotations)
@@ -859,7 +861,7 @@ namespace Microsoft.AspNetCore.OData.Formatter.Deserialization
                 return resourceWrapper;
             }
 
-            IEnumerable<ODataProperty> keys = CreateKeyProperties(resourceWrapper.Resource.Id, readContext);
+            IEnumerable<ODataProperty> keys = CreateKeyProperties(resourceWrapper.Resource.Id, readContext, isEntityReferenceLink: false);
             if (keys == null)
             {
                 return resourceWrapper;
@@ -893,9 +895,10 @@ namespace Microsoft.AspNetCore.OData.Formatter.Deserialization
         /// </summary>
         /// <param name="id">The key Id.</param>
         /// <param name="readContext">The reader context.</param>
+        /// <param name="isEntityReferenceLink">Creation is running for entity reference link.</param>
         /// <returns>The key properties.</returns>
         [SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "<Pending>")]
-        private static IList<ODataProperty> CreateKeyProperties(Uri id, ODataDeserializerContext readContext)
+        private static IList<ODataProperty> CreateKeyProperties(Uri id, ODataDeserializerContext readContext, bool isEntityReferenceLink)
         {
             Contract.Assert(id != null);
             Contract.Assert(readContext != null);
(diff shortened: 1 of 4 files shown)


RamjotSingh:
@microsoft-github-policy-service agree [company="Microsoft"]

RamjotSingh:

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -910,6 +913,8 @@ private static IList<ODataProperty> CreateKeyProperties(Uri id, ODataDeserialize
                 IEdmModel model = readContext.Model;
                 HttpRequest request = readContext.Request;
                 IServiceProvider requestContainer = request.GetRouteServices();
+                IODataFeature oDataFeature = request.ODataFeature();
```
mikepizzo, line 916: >oDataFeature [](http://example.com/codeflow?start=30&length=12)

This does not appear to be used anywhere? Either remove or use on Line 950

src/Microsoft.AspNetCore.OData/ODataOptions.cs
```diff
@@ -63,6 +63,11 @@ public class ODataOptions
         /// </summary>
         public ODataRouteOptions RouteOptions { get; } = new ODataRouteOptions();
 
+        /// <summary>
+        /// Gets or sets a value indicating whether service root URI should be validated against the resource URI during deserialization of entity reference link.
+        /// </summary>
+        public bool EnableServiceRootValidationOnEntityReferenceLinkDeserialization { get; set; } = true;
```
mikepizzo, line 69: >EnableServiceRootValidationOnEntityReferenceLinkDeserialization [](http://example.com/codeflow?start=20&length=63)

wow; that's a *really* long name...

RamjotSingh: I am open to renaming, I could not think of a name that explained the scope correctly without leaving room for confusion, so this. I am open to suggestions.

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -922,15 +927,37 @@ private static IList<ODataProperty> CreateKeyProperties(Uri id, ODataDeserialize
...
+                    // and the @odata.bind URI might be different. This can happen in cases where there are intermediate
+                    // routers (proxies) between the client and the server.
+                    string relativeResourceUri = resolvedId.PathAndQuery;
+
+                    // If path contains a Route prefix, remove it from the relative URI.
+                    if (!string.IsNullOrEmpty(request.ODataFeature().RoutePrefix))
+                    {
+                        // The path will always start with a '/'. So we are removing the route prefix as well
+                        // /routePrefix/odata/Customers(1) -> /odata/Customers(1)
+                        relativeResourceUri = relativeResourceUri.Substring(relativeResourceUri.IndexOf('/', 1));
```
mikepizzo, line 954: >IndexOf('/', 1) [](http://example.com/codeflow?start=96&length=15)

This seems to rely on the route prefix being a single segment; however, I believe the route prefix can be a multi-segment path.

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -922,15 +927,37 @@ private static IList<ODataProperty> CreateKeyProperties(Uri id, ODataDeserialize
...
+                }
+                else
+                {
+                    // Get the relative URI from the service root. This is needed since for servers, the request root URI
+                    // and the @odata.bind URI might be different. This can happen in cases where there are intermediate
+                    // routers (proxies) between the client and the server.
+                    string relativeResourceUri = resolvedId.PathAndQuery;
+
+                    // If path contains a Route prefix, remove it from the relative URI.
+                    if (!string.IsNullOrEmpty(request.ODataFeature().RoutePrefix))
```
mikepizzo, line 950: >string.IsNullOrEmpty(request.ODataFeature().RoutePrefix) [](http://example.com/codeflow?start=25&length=56)

I think this logic is flawed.  The RoutePrefix is going to be the prefix for your service, but relativeResourceUri is going to be the URI that was passed to the host service. You are deciding whether you want to trim the external uri based on your back-end implementation, which seems flawed.  You would need some other way to determine what part of the external Uri was the route and what part was the actual Uri.

mikepizzo:
Taking a step back:

There is a fundamental problem with adding this logic to the libraries, which is that the libraries don't know, for an arbitrary incoming URL, what part is the serviceRoot versus what part is the relative URL. You know that for your service, but the incoming URL is just a URL so there's no way to know.  This PR uses whether your service has a prefix to determine whether or not to remove a segment from the incoming URL, which only happens to work if your service and your host service both happen to have a single segment prefix (or not), but that doesn't extend to the general case.

The other concern with adding this PR is that it's patching the behavior in one place (creating the wrapper for @odata.bind) but, in the general case, there may be other places where urls are provided that need similar rewriting.

What is really needed is a hook to call out to your service with the incoming URL so that your service has the opportunity to normalize it before continuing.

Happily, such a hook exists.

The logic for parsing the request (that is failing when the root of the request is different than your service root) is in the UriPathParser, which is a public, non-sealed class. You can derive your own type from this class, override the ParsePathIntoSegments virtual method to make absolute urls with a different service root relative, and then pass that to base method.

~~~c#
    public class MySimpleUriPathParser : UriPathParser
    {
        public MySimpleUriPathParser(ODataUriParserSettings settings) : base(settings) {}

        public override ICollection<string> ParsePathIntoSegments(Uri requestUri, Uri serviceBaseUri)
        {
            // if the request comes from a different host
            if (requestUri.IsAbsoluteUri && requestUri.Host != serviceBaseUri.Host)
            {
                // skip the expected prefix, retaining the query string, and set the serviceBase to null
                string query = requestUri.Query;
                requestUri = new Uri("/" + String.Concat(requestUri.Segments.Skip(2)) + query, UriKind.Relative);
                serviceBaseUri = null;
            }

            return base.ParsePathIntoSegments(requestUri, serviceBaseUri);
        }
    }
~~~

You can then inject your UriPathParser through dependency injection, so it gets called when parsing the path.

~~~C#
                .AddOData(opt => opt.Count().Filter().Expand().Select().OrderBy().SetMaxTop(500)
                    .AddRouteComponents("v1", model1, services => services.AddScoped<UriPathParser,MySimpleUriPathParser>())
                );
~~~

mikepizzo:
@RamjotSingh  -- since we have a solution that doesn't require changing the existing code, I'm going to close this PR. If the solution doesn't pan out, and we are able to figure out how to modify the incoming URL in a generic way and let it apply across all incoming URLs, then we can re-open.

Thanks for your contribution!

Issue #945, opened 2023-05-30
Output set of $apply transformations is not available for subsequent transformations or other query operators


imoraca:
**Assemblies affected**
Microsoft.AspNetCore.OData v 8.2.0 (Problem persists since v 8.0.5)

**Describe the bug**
When $apply is combined with $filter or $orderby an error occurs. $apply should be evaluated first, so all dynamic properties introduced in the $apply should be available for later query options. However this doesn't seem to work. 

**Example**
```
http://localhost:5000/odata/Orders?$apply=groupby((Customer/Name), aggregate($count as OrderCount, TotalAmount with sum as TotalAmount))&$filter=TotalAmount gt 23

http://localhost:5000/odata/Orders?$apply=groupby((Customer/Name), aggregate($count as OrderCount, TotalAmount with sum as TotalAmount))/filter(TotalAmount gt 23)
```
In both cases I receive the following error:
The binary operator GreaterThan is not defined for the types 'System.Object' and 'System.Nullable`1[System.Decimal]'

When I try to combine $apply and $order I also get an error:
```
http://localhost:5000/odata/Orders?$apply=groupby((Customer/Name), aggregate($count as OrderCount, TotalAmount with sum as Total))&$orderby=Total desc
```
This time the error is:
Could not find a property named 'Total' on type 'AggregationSample.Models.Order'

I've also looked at comment https://github.com/OData/AspNetCoreOData/issues/420#issuecomment-1343281828 and I totally agree. In my example groupby transformation in $apply includes a property from the input data set, but that property is still not available for subsequent transformations or other query options.

These queries use to work prior to ver. 8.0.5. Is there any workaround for this and will it be fixed in future versions?

imoraca:
Hi, is there any news on this matter? Do you know when this functionality will be available in the new version? I'm asking as I'm currently stuck on version 8.0.4 and can't move forward without this functionality.

PR #987, opened 2023-07-12
Make output set of $apply transformations available for subsequent transformations or other query operators

ElizabethOkerio:
This PR fixes https://github.com/OData/AspNetCoreOData/issues/945, #984 

This fix makes this work: 
```url
http://localhost:5000/odata/Orders?$apply=groupby((Customer/Name), aggregate($count as OrderCount, TotalAmount with sum as Total))&$orderby=Total desc
```

Like you can use the output of $apply transformations, `Total`,  in this case to perform subsequent transformations, `$orderby` in this case. 

This is how the `ApplyClause` property looks like: 

```c#
        public ApplyClause ApplyClause
        {
            get
            {
                if (_applyClause == null)
                {
                    _applyClause = _queryOptionParser.ParseApply();
                }

                return _applyClause;
            }
        }
```

So what happens when this is called: `if (options.Apply?.ApplyClause != null)`.  if `ApplyClause` is null,which will be null at this point, then the `ParseApply` gets called which parses the apply clause and binds the dynamic properties introduced in the `$apply` transformation. This makes these properties recognizable in subsequent transformations.

Before this change, the parsing of the apply clause was not happening before the `$orderby` gets called.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/Validator/ODataQueryValidator.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Validator/ODataQueryValidator.cs
@@ -39,12 +39,9 @@ namespace Microsoft.AspNetCore.OData.Query.Validator
                 options.Compute.Validate(validationSettings);
             }
 
-            if (options.Apply != null)
+            if (options.Apply?.ApplyClause != null)
             {
-                if (options.Apply.ApplyClause != null)
-                {
-                    ValidateQueryOptionAllowed(AllowedQueryOptions.Apply, validationSettings.AllowedQueryOptions);
-                } 
+                ValidateQueryOptionAllowed(AllowedQueryOptions.Apply, validationSettings.AllowedQueryOptions);
             }
 
             if (options.Skip != null)
--- a/src/Microsoft.AspNetCore.OData/Query/Validator/ODataQueryValidator.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Validator/ODataQueryValidator.cs
@@ -41,7 +41,10 @@ namespace Microsoft.AspNetCore.OData.Query.Validator
 
             if (options.Apply != null)
             {
-                ValidateQueryOptionAllowed(AllowedQueryOptions.Apply, validationSettings.AllowedQueryOptions);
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.OData/Query/Validator/ODataQueryValidator.cs
```diff
@@ -41,7 +41,10 @@ public virtual void Validate(ODataQueryOptions options, ODataValidationSettings
 
             if (options.Apply != null)
             {
-                ValidateQueryOptionAllowed(AllowedQueryOptions.Apply, validationSettings.AllowedQueryOptions);
+                if (options.Apply.ApplyClause != null)
```
gathogojr, line 44: You could use the null propagation operator here to avoid the extra level of nesting

if (options.Apply?.ApplyClause != null)

gathogojr:
@ElizabethOkerio For my understanding, how did this regression come about and how does the seemingly innocuous code change fix the issue? Can you add an explanation in the pull request description?

src/Microsoft.AspNetCore.OData/Query/Validator/ODataQueryValidator.cs
```diff
@@ -39,7 +39,7 @@ public virtual void Validate(ODataQueryOptions options, ODataValidationSettings
                 options.Compute.Validate(validationSettings);
             }
 
-            if (options.Apply != null)
+            if (options.Apply?.ApplyClause != null)
```
gathogojr, line 42: It's strange that the check whether `ApplyClause` is null was removed yet it was there in 7.x https://github.com/OData/WebApi/blob/8c03a9077020859096a65a49bcc6e6e68429ab98/src/Microsoft.AspNet.OData.Shared/Query/Validators/ODataQueryValidator.cs#L37

gathogojr (approved):
:shipit:

gathogojr:
@ElizabethOkerio Does this pull request fix this issue too? https://github.com/OData/AspNetCoreOData/issues/984
If it does, please link it to this PR such that the issue will be closed when the pull request is merged.

Issue #390, opened 2021-11-23
Async odata endpoint not returning edm model


andrew-vdb:
Actual:
The setup use EDM model, 
when we use IQueryable, the db call will be done synchronously and return EDM model
when we use IAsyncEnumerable, the db call will be done async and return simple model..
Expected:
when we use IAsyncEnumerable, the db call will be done async and should return EDM model..

```
//This return EDM model
[HttpGet("odata/students")]
        [EnableQuery]
        public IQueryable<Model.Student> OData()
        {
            return _context.Students;
        }

//This NOT return EDM model
        [HttpGet("odata/students")]
        [EnableQuery]
        public IAsyncEnumerable<Model.Student> OData()
        {
            return _context.Students.AsAsyncEnumerable();
        }
```

andrew-vdb:
This is not supported
```
[HttpGet("odata/students")]
        [EnableQuery]
        public IAsyncQueryable<Model.Student> OData()
        {
            return _context.Students.AsAsyncEnumerable().AsAsyncQueryable();
        }
```

xuzhg:
Could be similar to https://github.com/OData/WebApi/issues/2598

andrew-vdb:
Just tested with EnableQueryAsync in another issue...

https://localhost:7040/odata/students?$count=true&$top=2

The select top will run async
The select **count** will run **sync**
and it will return edm data as expected

jr01:
See my other issue for count async support: https://github.com/OData/WebApi/issues/2269

jr01:
See this issue for count async in [EnableQueryAsync] https://github.com/OData/WebApi/issues/2325

PR #988, opened 2023-07-13
support IAsyncenumerable

ElizabethOkerio:
This PR fixes https://github.com/OData/AspNetCoreOData/issues/390

Currently, If you have such a controller action method in OData AspNetCore: 

```c#

 [EnableQuery]
 public IAsyncEnumerable<Customer> Get()
 {
      return _context.Customers.AsAsyncEnumerable();
 }
```
The request is processed asynchronously and a non-OData response is returned.

This change ensures that an OData response is returned. 

The issue was, `AspNetCore OData` could not map an `IAsyncEnumerable<>` to an `EdmCollectionType`. To the `DefaultODataTypeMapper` I've added a check for an `IAsyncEnumerable<>`. If the collection is of type `IAsyncEnumerable<>` then it is mapped to an `EdmCollectionType`.

If we map the `IAsyncEnumerable<>` to an `EdmCollectionType` then the `ODataResourceSetSerializer` is used in serializing the response. 

To the `ODataResourceSetSerializer` I've added a check for the collection type. If the type is of `IAsyncEnumerable<>` then we use the `await foreach` to loop through the resources in the collection. Otherwise, we use the normal `foreach` loop.


habbes (commented):
It doesn't appear that we are processing the `IAsyncEnumerable` asynchronously (i.e. using `await foreach`). I think that's the whole point of this PR.

habbes:
I think you should also test with `IAsyncEnumerable` objects that are not `IEnumerable` to make sure we support that scenario as well.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -185,6 +246,36 @@ await WriteObjectInlineAsync(graph, resourceSetType, writer, writeContext)
             await writer.WriteEndAsync().ConfigureAwait(false);
         }
 
+        private async Task WriteResourceSetAsync(
```
habbes, line 249: I think this method should be called `WriteResourceSetItemAsync` or `WriteResourceSetEntryAsync` since it only writes a single item.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -150,28 +165,74 @@ await WriteObjectInlineAsync(graph, resourceSetType, writer, writeContext)
...
+        private async Task WriteResourceSetAsync(IAsyncEnumerable<object> asyncEnumerable, IEdmTypeReference resourceSetType, ODataWriter writer,
+           ODataSerializerContext writeContext)
+        {
+            Contract.Assert(writer != null);
+            Contract.Assert(writeContext != null);
+            Contract.Assert(asyncEnumerable != null);
+            Contract.Assert(resourceSetType != null);
+
+            IEdmStructuredTypeReference elementType = GetResourceType(resourceSetType);
+            ODataResourceSet resourceSet = CreateResourceSet(asyncEnumerable, resourceSetType.AsCollection(), writeContext);
```
habbes, line 196: Seems like this is the only line that differs between this and the `IEnumerable` version. I wonder whether more code could be shared neatly between the two to avoid cases where we update one version and forget about the other.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -403,6 +494,81 @@ private async Task WriteUntypedResourceSetItemAsync(object item, IEdmTypeReferen
...
         }
 
+        /// <summary>
+        /// Create the <see cref="ODataResourceSet"/> to be written for the given resourceSet instance.
+        /// </summary>
+        /// <param name="resourceSetInstance">The instance representing the resourceSet being written.</param>
+        /// <param name="resourceSetType">The EDM type of the resourceSet being written.</param>
+        /// <param name="writeContext">The serializer context.</param>
+        /// <returns>The created <see cref="ODataResourceSet"/> object.</returns>
+        public virtual ODataResourceSet CreateResourceSet(IAsyncEnumerable<object> resourceSetInstance, IEdmCollectionTypeReference resourceSetType,
```
habbes, line 504: Not sure if this could share code with the `CreateResourceSet(IEnumerable)` overload.

test/Microsoft.AspNetCore.OData.E2E.Tests/IAsyncEnumerableTests/IAsyncEnumerableController.cs
```diff
@@ -0,0 +1,128 @@
...
+        [EnableQuery]
+        [HttpGet("odata/Customers")]
+        public IAsyncEnumerable<Customer> Get()
+        {
+            return _context.Customers.AsAsyncEnumerable();
+        }
+
+        public async IAsyncEnumerable<Customer> CreateCollectionAsync<T>()
+        {
+            await Task.Delay(100);
```
habbes, line 50: Since the actual delay doesn't matter - we only care that it's asynchronous - I'd suggest using a shorter delay time to avoid slowing down the tests.

habbes (dismissed):

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -403,6 +491,53 @@ private async Task WriteUntypedResourceSetItemAsync(object item, IEdmTypeReferen
...
         }
 
+        internal void WriteEntityTypeOperations(
+            ODataResourceSet resourceSet,
+            ResourceSetContext resourceSetContext,
+            IEdmStructuredTypeReference structuredType,
+            ODataSerializerContext writeContext)
+        {
+            IEdmEntityType entityType = structuredType.AsEntity().EntityDefinition();
+            var operations = writeContext.Model.GetAvailableOperationsBoundToCollection(entityType);
```
KenitoInc, line 501: Use the type

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -440,6 +575,45 @@ private async Task WriteUntypedResourceSetItemAsync(object item, IEdmTypeReferen
             return (obj) => { return null; };
         }
 
+        /// <summary>
+        /// Creates a function that takes in an object and generates nextlink uri.
```
KenitoInc, line 579:
```suggestion
        /// Creates a function that takes in an object and generates a nextlink uri.
```

test/Microsoft.AspNetCore.OData.E2E.Tests/IAsyncEnumerableTests/IAsyncEnumerableTests.cs
```diff
@@ -0,0 +1,88 @@
...
+//      See License.txt in the project root for license information.
+// </copyright>
+//------------------------------------------------------------------------------
+
+using Microsoft.AspNetCore.OData.TestCommon;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.OData.Edm;
+using Newtonsoft.Json.Linq;
+using System.Collections.Generic;
```
KenitoInc, line 13: `using System.*` should be ordered first

test/Microsoft.AspNetCore.OData.E2E.Tests/IAsyncEnumerableTests/IAsyncEnumerableTests.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.Net.Http;
+using System.Net;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.AspNetCore.OData.E2E.Tests.IAsyncEnumerableTests
+{
+    public class IAsyncEnumerableTests : WebODataTestBase<IAsyncEnumerableTests.TestsStartup>
+    {
+       public class TestsStartup : TestStartupBase
```
KenitoInc, line 24: Can we add a test that applies query options?

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -120,6 +131,73 @@ await WriteObjectInlineAsync(graph, resourceSetType, writer, writeContext)
...
+            // the next page link is not set when calling WriteStart(resourceSet) but is instead set later on that resourceSet
+            // object before calling WriteEnd(), the next page link will be written at the end, as required for
+            // odata.streaming=true support.
+
+            resourceSet.NextPageLink = nextLinkGenerator(lastResource);
+
+            await writer.WriteEndAsync().ConfigureAwait(false);
+        }
+
+        private void WriteResourceSetInternal(
```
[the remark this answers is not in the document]
ElizabethOkerio: but it is part of the WriteResourceSet hence the `Internal` suffix

test/Microsoft.AspNetCore.OData.E2E.Tests/IAsyncEnumerableTests/IAsyncEnumerableTests.cs
ElizabethOkerio: I'd like to leave this out yet for next PR since the PR does not support applying query options asynchronously.

src/Microsoft.AspNetCore.OData/Formatter/ResourceSetContext.cs
```diff
@@ -61,5 +62,25 @@ internal static ResourceSetContext Create(ODataSerializerContext writeContext, I
 
             return resourceSetContext;
         }
+
+        /// <summary>
+        /// Create a <see cref="ResourceSetContext"/> from an <see cref="ODataSerializerContext"/> and <see cref="IAsyncEnumerable{T}"/>.
+        /// </summary>
+        /// <param name="resourceSetInstance">The instance representing the resourceSet being written.</param>
+        /// <param name="writeContext">The serializer context.</param>
```
xuzhg, line 70: switch the order

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -137,52 +215,47 @@ await WriteObjectInlineAsync(graph, resourceSetType, writer, writeContext)
...
-                    // for null complex element, it can be serialized as "null" in the collection.
-                    await writer.WriteStartAsync(resource: null).ConfigureAwait(false);
-                    await writer.WriteEndAsync().ConfigureAwait(false);
-                }
-                else if (isUntypedCollection)
-                {
-                    await WriteUntypedResourceSetItemAsync(item, resourceSetType, writer, writeContext).ConfigureAwait(false);
-                }
-                else
+        private async Task WriteResourceEntryAsync(
```
xuzhg, line 231: No 'Entry' or use other

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -403,6 +491,53 @@ private async Task WriteUntypedResourceSetItemAsync(object item, IEdmTypeReferen
             return resourceSet;
         }
 
+        internal void WriteEntityTypeOperations(
```
xuzhg, line 494: why "internal", not a private?

ElizabethOkerio: The existing methods are internal. This is an ovveride of the methods that we already have..

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -120,6 +131,73 @@ await WriteObjectInlineAsync(graph, resourceSetType, writer, writeContext)
...
+            WriteResourceSetInternal(resourceSet, elementType, resourceSetType, writeContext, out bool isUntypedCollection, out IODataEdmTypeSerializer resourceSerializer);
+
+            await writer.WriteStartAsync(resourceSet).ConfigureAwait(false);
+            object lastResource = null;
+
+            foreach (object item in enumerable)
+            {
+                lastResource = item;
+
+                await WriteResourceSetItemAsync(item, elementType, isUntypedCollection, resourceSetType, writer, resourceSerializer, writeContext);
```
habbes, line 143: Why don't we call `ConfigureAwait(false)` here?

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -120,6 +131,73 @@ await WriteObjectInlineAsync(graph, resourceSetType, writer, writeContext)
...
+            WriteResourceSetInternal(resourceSet, elementType, resourceSetType, writeContext, out bool isUntypedCollection, out IODataEdmTypeSerializer resourceSerializer);
+            
+            await writer.WriteStartAsync(resourceSet).ConfigureAwait(false);
+            object lastResource = null;
+
+            await foreach (object item in asyncEnumerable)
+            {
+                lastResource = item;
+
+                await WriteResourceSetItemAsync(item, elementType, isUntypedCollection, resourceSetType, writer, resourceSerializer, writeContext);
```
habbes, line 179: Should we call `ConfigureAwait(false)` here?

```suggestion
                await WriteResourceSetItemAsync(item, elementType, isUntypedCollection, resourceSetType, writer, resourceSerializer, writeContext).ConfigureAwait(false);
```

habbes (approved):

src/Microsoft.AspNetCore.OData/Formatter/ResourceSetContext.cs
```diff
@@ -61,5 +62,23 @@ internal static ResourceSetContext Create(ODataSerializerContext writeContext, I
...
+        /// <param name="resourceSetInstance">The instance representing the resourceSet being written.</param>
+        /// <returns>A new <see cref="ResourceSetContext"/>.</returns>
+        /// <remarks>This signature uses types that are AspNetCore-specific.</remarks>
+        internal static ResourceSetContext Create(ODataSerializerContext writeContext, IAsyncEnumerable<object> resourceSetInstance)
+        {
+            return new ResourceSetContext
+            {
+                Request = writeContext.Request,
+                EntitySetBase = writeContext.NavigationSource as IEdmEntitySetBase,
+                //  Url = writeContext.Url,
```
xuzhg, line 79: >  //  Url = writeContext.Url, [](http://example.com/codeflow?start=14&length=29)

still need this line?

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
```diff
@@ -97,14 +96,26 @@ await WriteObjectInlineAsync(graph, resourceSetType, writer, writeContext)
                 throw new SerializationException(Error.Format(SRResources.CannotSerializerNull, ResourceSet));
             }
 
-            IEnumerable enumerable = graph as IEnumerable; // Data to serialize
-            if (enumerable == null)
+            if (writeContext.Type != null && writeContext.Type.IsGenericType && writeContext.Type?.GetGenericTypeDefinition() == typeof(IAsyncEnumerable<>))
             {
-                throw new SerializationException(
-                    Error.Format(SRResources.CannotWriteType, GetType().Name, graph.GetType().FullName));
+                if (!(graph is IAsyncEnumerable<object> asyncEnumerable))
```
xuzhg, line 101: >object [](http://example.com/codeflow?start=48&length=6)

why the generic type is 'object'?

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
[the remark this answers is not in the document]
xuzhg: I agree with @KenitoInc , the method name doesn't do any writing?

xuzhg (approved):
:shipit:

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
ElizabethOkerio: The `graph` is an object and we do not know the type `T` at this point not unless we use reflection to find it which I think is not of significance at this point.

ElizabethOkerio: This is why we use `IEnumerable` and not `IEnumerable<T>` in the sync enumerations.

test/Microsoft.AspNetCore.OData.E2E.Tests/IAsyncEnumerableTests/IAsyncEnumerableDataModel.cs
```diff
@@ -0,0 +1,54 @@
+﻿//-----------------------------------------------------------------------------
+// <copyright file="IAsyncEnumerableDataModel.cs" company=".NET Foundation">
+//      Copyright (c) .NET Foundation and Contributors. All rights reserved.
+//      See License.txt in the project root for license information.
+// </copyright>
+//------------------------------------------------------------------------------
+
+using Microsoft.EntityFrameworkCore;
+using System.Collections.Generic;
```
KenitoInc, line 9: re-order usings

KenitoInc (approved):

KenitoInc (approved):

aldrashan:
Doesn't appear to work in all cases?
See https://github.com/OData/WebApi/issues/2598#issuecomment-2042677354

PR #993, opened 2023-07-20
fix duplicate join issues

ElizabethOkerio:


KenitoInc (commented):
Add some unit tests [here ](https://github.com/OData/AspNetCoreOData/blob/main/test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/SelectExpandBinderTest.cs)

src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinderContext.cs
```diff
@@ -168,6 +168,11 @@ public QueryBinderContext(QueryBinderContext context, ODataQuerySettings querySe
         /// </summary>
         public Expression Source { get;set; }
 
+        /// <summary>
+        /// Gets or sets the query provider.
+        /// </summary>
+        internal string QueryProvider { get; set; }
```
xuzhg, line 174: do you want to add test cases to cover different query provider?

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -427,6 +435,50 @@ internal Expression ProjectElement(QueryBinderContext context, Expression source
             return Expression.MemberInit(Expression.New(wrapperType), wrapperTypeMemberAssignments);
         }
 
+        // Generates the expression
+        //      { Instance = new Customer() {Id = $it.Id}}
+        private Expression UpdateMemberInitExpression(Expression source, IEdmStructuredType structuredType)
```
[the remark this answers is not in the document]
ElizabethOkerio: Here:                 Expression sourceExpression = UpdateMemberInitExpression(source, structuredType);

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -427,6 +435,51 @@ internal Expression ProjectElement(QueryBinderContext context, Expression source
...
+        //      { Instance = new Customer() {Id = $it.Id, Name= $it.Name}}
+        private Expression UpdateMemberInitExpression(Expression source, IEdmStructuredType structuredType)
+        {
+            Expression sourceExpression = null;
+
+            if (QueryProvider != null &&
+                QueryProvider == HandleNullPropagationOptionHelper.EntityFrameworkQueryProviderNamespace ||
+                QueryProvider == HandleNullPropagationOptionHelper.ObjectContextQueryProviderNamespaceEF5 ||
+                QueryProvider == HandleNullPropagationOptionHelper.ObjectContextQueryProviderNamespaceEF6 ||
+                QueryProvider == HandleNullPropagationOptionHelper.ObjectContextQueryProviderNamespaceEFCore2)
```
habbes, line 448: Not sure how often we use this pattern, but I think a helper method like `IsEntityFrameworkProvider` could be helpful to ensure consistency.

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -58,6 +59,8 @@ internal SelectExpandBinder()
         /// </summary>
         public IOrderByBinder OrderByBinder { get; }
 
+        internal string QueryProvider { get; set; }
```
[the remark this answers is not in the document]
habbes: Is this property used outside of this class? Maybe it should be a private readonly field.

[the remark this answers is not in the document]
habbes: Actually, does it need to be a field/property at all? Since it can be retrieved from the `QueryBinderContext` that's passed to relevant methods, I don't think it should be a property or field. Just a local variable where needed.

In general, it's a bad idea for methods to mutate the object in ways that are not expected. Especially classes like this that could be injected as a singleton (not sure how it's injected in the service container) could suffer from race conditions if they're arbitrarily mutated by instance methods.

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -368,9 +374,11 @@ internal Expression ProjectElement(QueryBinderContext context, Expression source
             bool isSelectedAll = IsSelectAll(selectExpandClause);
             if (isSelectedAll)
             {
+                Expression sourceExpression = UpdateMemberInitExpression(source, structuredType);
```
habbes, line 377: This `sourceExpression` variable name here is very confusing. It's not clear what the different between `source` and `sourceExpression` is.

The name `UpdateMemberInitExpression` is also confusing, it doesn't help understand what it does or why it's needed. Consider a more descriptive name like `RemoveNonStructucalProperties(source, structuredType)`. If you use this name, it might make sense to move the `if (IsEFCore...)` check outside of the method and place it here.

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -427,6 +435,51 @@ internal Expression ProjectElement(QueryBinderContext context, Expression source
             return Expression.MemberInit(Expression.New(wrapperType), wrapperTypeMemberAssignments);
         }
 
+        // Generates the expression
+        //      { Instance = new Customer() {Id = $it.Id, Name= $it.Name}}
+        private Expression UpdateMemberInitExpression(Expression source, IEdmStructuredType structuredType)
+        {
+            Expression sourceExpression = null;
+
+            if (QueryProvider != null &&
```
habbes, line 444: I think it would be helpful to mention in a comment why this transformation is only being applied to EF. You may also reference the issue URL in the explanation.

habbes, line 444: I think the if statement would be easier to read with less indentation:

```c#
if (isNotEfCore) {
    return source;
}

// source transformation logic...
}

File: src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs [hunk 7]
```diff
@@ -427,6 +435,51 @@ internal Expression ProjectElement(QueryBinderContext context, Expression source
...
+
+                foreach (PropertyInfo prop in props)
+                {
+                    if (structuralPropertyNames.Contains(prop.Name))
+                    {
+                        MemberExpression propertyExpression = Expression.Property(source, prop);
+                        bindings.Add(Expression.Bind(prop, propertyExpression));
+                    }
+                }
+
```
habbes, line 470: This appears potentially inefficient. We create two List<T> instances. O(n) loop to populate the `structuralPropertyNames` and O(n^2) loop when adding members to the binding due to the `Contains` call. I don't know if a `HashSet` would help, in practice that would depend on the size of the collection. But I just wonder if we can do with fewer loops and allocations.

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/SelectExpandBinderTest.cs
```diff
@@ -203,12 +203,52 @@ public void Bind_GeneratedExpression_ContainsExpandedObject()
                 .ToDictionary(PropertyMapper)["Orders"] as IEnumerable<SelectExpandWrapper<QueryOrder>>;
             Assert.NotNull(innerOrders);
             SelectExpandWrapper<QueryOrder> partialOrder = innerOrders.Single();
+            var tt = _queryable.First().Orders.First();
+            var xx = _queryable.First();
```
habbes: What are these?

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/SelectExpandBinderTest.cs
```diff
@@ -203,12 +203,52 @@ public void Bind_GeneratedExpression_ContainsExpandedObject()
...
+        [InlineData(HandleNullPropagationOptionHelper.ObjectContextQueryProviderNamespaceEF5)]
+        [InlineData(HandleNullPropagationOptionHelper.ObjectContextQueryProviderNamespaceEF6)]
+        public void Bind_UsingEFQueryProvider_GeneratedExpression_ContainsExpandedObject(string queryProvider)
+        {
+            // Arrange
+            SelectExpandQueryOption selectExpand = new SelectExpandQueryOption("Orders", "Orders,Orders($expand=Customer)", _context);
+
+            // Act
+            SelectExpandBinder binder = new SelectExpandBinder();
+            _queryBinderContext.QueryProvider = queryProvider;
```
habbes, line 224: The query provider should ideally be provided through the context rather than set directly.

habbes (commented):
I think it would be great to add a test that looks like the customer scenario described in the issue, i.e.: applying OData query options to an IQueryable that is the result of a .Select() mapping from one type to another, where the input comes from a DbSet with an Includes call on a navigation property.

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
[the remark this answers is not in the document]
KenitoInc: All properties should be wrapped inside `QueryBinderContext` since we pass `QueryBinderContext` to all methods.
`SelectExpandBinder` is injected as a singleton

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/SelectExpandBinderTest.cs
KenitoInc: +1

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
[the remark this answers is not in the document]
ElizabethOkerio: It should be a field...Since the method in which it's called..get called by other some methods which pass a null `QueryBinderContext`.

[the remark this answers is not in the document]
ElizabethOkerio: Also, it can't be a readonly since we assign the query provider value to it from the context the first time this method gets called.

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
ElizabethOkerio: fixed.

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
ElizabethOkerio: fixed

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
ElizabethOkerio: fixed

ElizabethOkerio: fixed

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
ElizabethOkerio: fixed

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/SelectExpandBinderTest.cs
ElizabethOkerio: This is a test scenario. In normal case, the query provider is gotten from the query,, The QueryBinderContext does not have any references to the query. The reason I've added the `QueryProvider` is to enable it to set to the `QueryBinderContext` and can be accessed from other methods with ease since the `QueryBinderContext` is moved to methods..

KenitoInc (approved):

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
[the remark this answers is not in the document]
ElizabethOkerio: fixed.

habbes (approved):

PR #996, opened 2023-07-20
Support writing nested entities without a navigation source.

mikepizzo:
There are cases where we don't have a URL when writing a collection of nested entities. This may be, for example, because there was no navigation property binding for the collection, because the collection was returned as the child of a complex type returned by a function, or because the entities were "transient".

In any case, lack of a URL should not prevent us from writing the nested data, if available.

Currently, we conflate writing the content with writing the url and, if we can't determine a navigation source, we skip writing the nested content entirely (although there are other cases where, even though we have a navigation source, we can't determine the url, so we already have cases where we write nested content without the url but only if we have a navigation source.)

This PR allows writing the content in any case, and optionally adds the URL if the necessary information is available.

### DIFF
--- a/test/Microsoft.AspNetCore.OData.Tests/Formatter/Serialization/ODataResourceSerializerTests.cs
+++ b/test/Microsoft.AspNetCore.OData.Tests/Formatter/Serialization/ODataResourceSerializerTests.cs
@@ -7,7 +7,10 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
+using System.Reflection.Metadata;
+using System.Runtime.CompilerServices;
 using System.Runtime.Serialization;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.OData.Abstracts;
@@ -20,6 +23,7 @@ using Microsoft.AspNetCore.OData.Tests.Commons;
 using Microsoft.AspNetCore.OData.Tests.Edm;
 using Microsoft.AspNetCore.OData.Tests.Extensions;
 using Microsoft.AspNetCore.OData.Tests.Models;
+using Microsoft.AspNetCore.Rewrite;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Net.Http.Headers;
 using Microsoft.OData;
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.OData.Tests/Formatter/Serialization/ODataResourceSerializerTests.cs
```diff
@@ -2249,6 +2253,71 @@ public async Task WriteObjectInlineAsync_SetsParentContext_ForExpandedNavigation
...
+                    new ExpandedNavigationSelectItem(new ODataExpandPath(new NavigationPropertySegment(productsProperty, null)),null,null)
+                },
+             false);
+
+            var writeContext = new ODataSerializerContext() 
+            { 
+                Model = model,
+                SelectExpandClause = selectExpand
+            };
+            var stream = new MemoryStream();
```
corranrogue9, line 2298:
```suggestion
            using var stream = new MemoryStream();
```

test/Microsoft.AspNetCore.OData.Tests/Formatter/Serialization/ODataResourceSerializerTests.cs
```diff
@@ -2249,6 +2253,71 @@ public async Task WriteObjectInlineAsync_SetsParentContext_ForExpandedNavigation
...
+            IODataResponseMessage responseMessage = new ODataMessageWrapper(stream);
```
corranrogue9, line 2299:
```suggestion
            using ODataMessageWrapper responseMessage = new ODataMessageWrapper(stream);
```

test/Microsoft.AspNetCore.OData.Tests/Formatter/Serialization/ODataResourceSerializerTests.cs
```diff
@@ -2249,6 +2253,71 @@ public async Task WriteObjectInlineAsync_SetsParentContext_ForExpandedNavigation
...
+            ODataUri uri = new ODataUri { ServiceRoot = new Uri("http://myService", UriKind.Absolute) };
+            ODataMessageWriterSettings settings = new ODataMessageWriterSettings
+            {
+                ODataUri = uri                
+            };
+            ODataMessageWriter messageWriter = new ODataMessageWriter(responseMessage, settings);
```
corranrogue9, line 2305:
```suggestion
            using ODataMessageWriter messageWriter = new ODataMessageWriter(responseMessage, settings);
```

test/Microsoft.AspNetCore.OData.Tests/Formatter/Serialization/ODataResourceSerializerTests.cs
```diff
@@ -2249,6 +2253,71 @@ public async Task WriteObjectInlineAsync_SetsParentContext_ForExpandedNavigation
...
+            ODataWriter writer = await messageWriter.CreateODataResourceWriterAsync(null, resultType as IEdmComplexType);
+
+            ODataResourceSerializer serializer = new ODataResourceSerializer(_serializerProvider);
+
+            // Act
+            await serializer.WriteObjectInlineAsync(result, resultTypeReference, writer, writeContext);
+
+            // Assert
+            stream.Position = 0;
+            StreamReader reader = new StreamReader(stream);
```
corranrogue9, line 2315:
```suggestion
            using StreamReader reader = new StreamReader(stream);
```

corranrogue9 (dismissed):

test/Microsoft.AspNetCore.OData.Tests/Formatter/Serialization/ODataResourceSerializerTests.cs
mikepizzo: MessageWrapper isn't IDisposable, but I've fixed the others.

habbes (approved):

ishatyagiit:
Hi @mikepizzo 
What version will this change be available in?

mikepizzo:
Resolved offline.  Controller method needed the [EnableQuery] attribute in order to support the expanded children.

PR #1005, opened 2023-08-01
fixes #986: pass Order value to selector model generated

### LINKED ISSUE #986: Order value should pass to selector model generated
**Assemblies affected**

**Describe the bug**

**Reproduce steps**

Supposed customers set up the order value on the attribute, after OData attribute routing convention, this order value is gone.
```C#
 [Route("/Data(Key1={Key1}, Key2={Key2})", Order = -1)]
 public IActionResult GetOne(long Key1, string Key2)
{
   ..
}
```

We should pass the order value into the new selector model generated in the OData attribute routing convention.

xuzhg:
fixes #986: pass Order value to selector model generated

Customer maybe specify the order value in the attribute, we should pass this value to the new selector model generated.

```C#
 [Route("/Data(Key1={Key1}, Key2={Key2})", Order = -1)]
 public IActionResult GetOne(long Key1, string Key2)
{
   ..
}
```


mikepizzo (approved):
:shipit:

PR #1012, opened 2023-08-08
ODL 7.16.0 makes ASP OData build hang-on

xuzhg:
Let's update to ODL 7.17.0

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.csproj
+++ b/src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.csproj
@@ -29,13 +29,13 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.OData.ModelBuilder" Version="1.0.9" />
-    <PackageReference Include="Microsoft.OData.Core" Version="7.16.0" />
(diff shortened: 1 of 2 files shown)


KenitoInc (approved):

gathogojr (approved):
:shipit:

mikepizzo:
@xuzhg  -- Do we know what about ODL 7.16 makes the ASP OData build hang? Is there a debug assert raised by ODL, or an issue that an app would encounter calling the release ODL 7.16?

mikepizzo (commented):
@xuzhg -- Do we know what about ODL 7.16 makes the ASP OData build hang? Is there a debug assert raised by ODL, or an issue that an app would encounter calling the release ODL 7.16?

xuzhg:
> @xuzhg -- Do we know what about ODL 7.16 makes the ASP OData build hang? Is there a debug assert raised by ODL, or an issue that an app would encounter calling the release ODL 7.16?

@mikepizzo , thanks for reviewing.
I assumed it could be related to ODL commits and Checked with @ElizabethOkerio, it could be related to:

https://github.com/OData/odata.net/pull/2670
https://github.com/OData/odata.net/pull/2671

However, the build is still hang on. I will continue to follow up it.

xuzhg:
It seems it is related to https://github.com/OData/AspNetCoreOData/pull/1013

The root cause is:
In an asysc test case, we can an async method using sync pattern. It's blocking the test running.

Issue #1023, opened 2023-08-18
Response contains only null and default values for all fields when using $expand without $select or (select=) in 8.2.1


orty:
**Assemblies affected**
ASP.NET Core OData 8.2.1

**Describe the bug**
Since the upgrade to v.8.2.1, we experience responses like the following:

```json
{
	"@odata.context": "..........$metadata#apimodel(expandedProperty())",
	"value": [
		{
			"title": null,
			"description": null,
			"id": "00000000-0000-0000-0000-000000000000",
			"expandedProperty": [
				{
					"id": "00000000-0000-0000-0000-000000000000",
					"categories": [],
					"description": null,
					"displayName": null,
					"isPublished": false
				}
			]
		}
	]
}
```
We did not have this behavior with the previous release (8.2.0) and it can be work-arounded by adding some top-level properties to the `$select` query option (hence, the top level `apimodel` object has its selected properties with values, but the expanded ones do not), or some nested properties to the `$expand=expandedProperty(select=..)` (and then, the selected nested properties have values too).

**Reproduce steps**
Simply issue a query with and `$expand` on a navigation property, regardless it is an entity or a collection property.

**Additional context**
I narrowed down my search for a cause of this bug to this PR #993. If I just rollback to the previous behavior in the `Microsoft.AspNetCore.OData.Query.Expressions.SelectExpandBinder.ProjectElement` method, it fixes the issue.

orty:
Found the issue, it is linked to the `EnableLowerCamelCase()` scenario.

In the following loop:
https://github.com/OData/AspNetCoreOData/blob/8fc0cfc3b4caa67b5c26b895920d7895d74fcc77/src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs#L456-L467
the `if (sp.Name == prop.Name)` condition does not take in account the use case with camelcased property names.

PR #1024, opened 2023-08-18
#1023 - Fix all structural properties ignored when using `$expand` without `$select` and camel cased property names

orty:
This fixes the issue #1023 where all structural properties are removed from the generated SQL query when using `$expand` without `$select` in a `builder.EnableLowerCamelCase()` scenario.

Probably fixes the same issue when using aliases in an EdmModel with the `[DataMember(Name=...)]` attribute.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.xml
+++ b/src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.xml
@@ -13961,7 +13961,6 @@
         <member name="P:Microsoft.AspNetCore.OData.Routing.ODataRouteOptions.Order">
             <summary>
             Gets or sets the route order in conventional routing.
-            By default, move the conventional routing later as much as possible.
             </summary>
         </member>
         <member name="P:Microsoft.AspNetCore.OData.Routing.ODataRouteOptions.EnableDollarCountRouting">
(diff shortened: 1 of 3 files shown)


ElizabethOkerio:
@orty Thanks for the contribution. Please add some tests.

src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.xml
```diff
@@ -13961,7 +13961,6 @@
         <member name="P:Microsoft.AspNetCore.OData.Routing.ODataRouteOptions.Order">
             <summary>
             Gets or sets the route order in conventional routing.
-            By default, move the conventional routing later as much as possible.
```
ElizabethOkerio, line 13964: why this change?

orty: I just rebuilt the solution to validate my fix, and the comment must have been removed in a previous commit without committing the regeneration of the documentation XML. I can revert it if, but it might come back the next time someone rebuilds the solution.

orty:
> @orty Thanks for the contribution. Please add some tests.

Done, I added a new lower cased model and all its context, plus a copy of your previous test method in this context, checking in both scenario that the top-level properties are not `null` or `default`

xuzhg (approved):

ElizabethOkerio:
The build is not running. Let me see whether I can manually run it.

PR #1033, opened 2023-08-29
Avoiding return DateTime.Kind=Unspecified when converting from DateTimeOffset to DateTime (issues 378, 384 and others)

### LINKED ISSUE #378: Dates in $filter parsed as DateTime with DateTimeKind.Unspecified
Npgsql 6.0 has a [breaking change](https://www.npgsql.org/efcore/release-notes/6.0.html#timestamp-rationalization-and-improvements) around timestamp mapping which breaks querying via OData. In order to query a Postgres column of type 'timestamp with time zone', the corresponding DateTime needs to have Kind = `DateTimeKind.Utc`.

However, all dates in the $filter clause are parsed with `DateTimeKind.Unspecified`, which causes Npgsql to throw an exception and results in OData returning an invalid and incomplete JSON response.

Is there a way to override the parsing of dates? Would it be possible for setting `ODataOptions.TimeZoneInfo = TimeZoneInfo.Utc` to also set the `Kind` of any parsed dates to UTC?

### LINKED ISSUE #384: Time zone settings not propagated to query
Using Microsoft.AspNetCore.OData V 8.0.4

in my startup I set

```c#
                .AddOData(opt =>
                {

                    opt.Select().Filter().OrderBy().Count().SetMaxTop(1000)
                        .AddRouteComponents("api", EdmModelBuilder.GetEdmModel())
                        .TimeZone = TimeZoneInfo.Utc;
                    opt.EnableAttributeRouting = true;

                })
```

And specify the time zone i want to use

When using 
```c#
[HttpGet("api/my-endpoint")]
        [EnableQuery(PageSize = 44, MaxTop = 100, AllowedQueryOptions = AllowedQueryOptions.All)]
```

The `ExpressionBinderHelper` is not setting the TimeZone

```c#
             right = DateTimeOffsetToDateTime(right, querySettings.TimeZone, querySettings);
```
https://github.com/OData/AspNetCoreOData/blob/master/src/Microsoft.AspNetCore.OData/Query/Expressions/ExpressionBinderHelper.cs#L76

Has a null timezone

I've read https://github.com/OData/AspNetCoreOData/issues/268 But this doesn't seem to have fixed the issue in 8.0.4

darkeagle76:
fixes #378 , fixes #384 
With this modification we would avoid getting a DateTime value (derived from a conversion from a DateTimeOffset value) that has Kind=Unspecified. The problem is mainly evident when passing Dates as parameters of the ODATA $filter option. With this change the conversion will aways return an UTC datetime when it is originated from a DateTimeOffset value.

I already provided additional [details here:](https://github.com/OData/AspNetCoreOData/issues/378#issuecomment-1672782754) but my original comment remained without answer.

Thanks in advance for considering my request.
D

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Edm/EdmPrimitiveHelper.cs
+++ b/src/Microsoft.AspNetCore.OData/Edm/EdmPrimitiveHelper.cs
@@ -106,7 +106,7 @@ namespace Microsoft.AspNetCore.OData.Edm
                         DateTimeOffset dateTimeOffsetValue = (DateTimeOffset)value;
                         TimeZoneInfo timeZone = timeZoneInfo ?? TimeZoneInfo.Local;
                         dateTimeOffsetValue = TimeZoneInfo.ConvertTime(dateTimeOffsetValue, timeZone);
-                        return dateTimeOffsetValue.DateTime;
+                        return dateTimeOffsetValue.UtcDateTime;
                     }
 
                     if (value is Date)


darkeagle76:
> With this modification we would avoid getting a DateTime value (derived from a conversion from a DateTimeOffset value) that has Kind=Unspecified. The problem is mainly evident when passing Dates as parameters of the ODATA $filter option. With this change the conversion will aways return an UTC datetime when it is originated from a DateTimeOffset value.
> 
> I already provided additional [details here:](https://github.com/OData/AspNetCoreOData/issues/378#issuecomment-1672782754) but my original comment remained without answer.
> 
> Thanks in advance for considering my request. D

pgrimmert (approved):

gathogojr:
Thank you @darkeagle76 for your contribution. The build was triggered and tests are failing after your change. Please check that out
In addition, please add tests to validate your change.

jusbuc2k:
Is there a reason this hasn't been addressed? This problem breaks any usage of odata with newer Postgres/EFcore with date filtering.

Perhaps if there's concern about breaking existing usages, we just need another option on query settings like this that can default to the old way:

```csharp
            options.TimeZone = TimeZoneInfo.Utc;
            options.DateTimeOffsetToDateTimeTransform = (DateTimeOffset dto) => dto.UtcDateTime;
```

VitaliyChaban:
What specific tests were not passed, and why hasn’t this been added to the release yet? 
I don’t know what might have broken due to these changes, but I can definitely say that WITHOUT these changes, it is impossible to work with OData on the new frameworks if time filtering is required.

I don't think that's a valid reason to reject this fix.

Expected: 2014-12-11T17:02:03.0000000
Actual: 2014-12-12T01:02:03.0000000Z

If you're expecting local time here, you can pass it like this: return DateTime.SpecifyKind(dateTimeOffsetValue, DateTimeKind.Local).ToUniversalTime();

But the 'Z' should be at the end and sent to the database, because any other format leads to the aforementioned error.

I believe the issue lies in your tests, not in the proposed fixes.

There is a compromise option as suggested above—just pass a function in the settings that will modify the dateTime if needed.

At the moment, the only solution to the problem is to abandon the official NuGet package and use the fork from darkeagle76.

VitaliyChaban:
What specific tests were not passed, and why hasn’t this been added to the release yet? 
I don’t know what might have broken due to these changes, but I can definitely say that WITHOUT these changes, it is impossible to work with OData on the new frameworks if time filtering is required.

I don't think that's a valid reason to reject this fix.

Expected: 2014-12-11T17:02:03.0000000
Actual: 2014-12-12T01:02:03.0000000Z

If you're expecting local time here, you can pass it like this: return DateTime.SpecifyKind(dateTimeOffsetValue, DateTimeKind.Local).ToUniversalTime();

But the 'Z' should be at the end and sent to the database, because any other format leads to the aforementioned error.

I believe the issue lies in your tests, not in the proposed fixes.

There is a compromise option as suggested above—just pass a function in the settings that will modify the dateTime if needed.

At the moment, the only solution to the problem is to abandon the official NuGet package and use the fork from darkeagle76.

VitaliyChaban:
I added a fix so that your tests pass, and I also added a new test that checks for the explicit indication of Kind=UTC.

Can we release this fix? At least as a preview version or something like that. We need this fix.

https://github.com/OData/AspNetCoreOData/pull/1316

PR #1075, opened 2023-10-24
Write Nav links when metadata full

### LINKED ISSUE #1071: A lot of time is spent processing navigation links even when we they're not written in response
Navigation links are usually only written when writing full metadata (`ODataMetadataLevel.Full`). By default, we do not write them in the response. However, we always process the navigation links when serializing the response, even we don't write them.

In the `ODataResourceSerializer.WriteResponseAsync` method below, we have a call to `WriteNavigationLinks`. 

image

This method will create and process navigation links, which will eventually not be written unless we're in full metadata. The actual navigation links values that are written are retrieved from a subclass of `ODataResourceMetadataBuilder`. A different type of `ODataResourceMetadataBuilder` is created depending on the metadata level. In the default case, the `NoOpResourceMetadataBuilder` is used, which returns null for the links. However, there's still a substantial amount of processing and expensive code that's executed before we get to the JSON output.

As you can see from the following graph, `WriteNavigationLinks` takes about 17% of CPU when profiling some sample code. 
image

Can we skip calling this method if the metadata level is not set to full?

ElizabethOkerio:
Fixes #1071

Navigation links only get written when `MetadataLevel` is set to `ODataMetadataLevel.Full` or when `MetadataLevel` is set to `ODataMetadataLevel.Minimal` and the `NavigationLinkBuilder's` FollowsConvention property is set to false. The ODataWriter methods for writing navigation links get called every time even when the navigation links are not being written. Calling these methods leads to a lot of CPU usage and high latency when the navigation links are not being written. 

In this PR, I've added a bool property `WriteNavigationLinks` in the ResourceContext... The default value for this property is `false`. If the metadata level is set to full or minimal and FollowsConvention is set to false, then the `WriteNavigationLinks`
is set to true. Otherwise, it is set to false.

So, before calling the writer methods for writing navigation links, we'll be checking the value of the `WriteNavigationLinks` property. If it is set to true, then the writer methods are called. Otherwise, the writer methods do not get called. 

NOTE: The response has 3000 records. 

Latency before this change: 
Bombarding https://localhost:7034/odata/Products for 10s using 50 connection(s)
[=================================================================================================================] 10s
Done!
```table
Statistics        Avg      Stdev        Max
  Reqs/sec        37.18     181.19    2499.47
  Latency         1.94s      0.91s      8.79s
  Latency Distribution
     50%      1.78s
     75%      1.92s
     90%      2.12s
     95%      4.24s
     99%      4.57s
  HTTP codes:
    1xx - 0, 2xx - 267, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    10.42MB/s
```
Latency after this change:
Bombarding https://localhost:7034/odata/Products for 10s using 50 connection(s)
[=================================================================================================================] 10s
Done!
```table
Statistics        Avg      Stdev        Max
  Reqs/sec        57.94     173.85    2249.86
  Latency         1.15s   244.10ms      1.88s
  Latency Distribution
     50%      1.19s
     75%      1.29s
     90%      1.38s
     95%      1.50s
     99%      1.88s
  HTTP codes:
    1xx - 0, 2xx - 449, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    16.27MB/s
```

This gives a `40.7%` latency improvement. 

CPU usage: 
Before
ssas

After
popo

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
@@ -25,7 +25,6 @@ using NavigationSourceLinkBuilderAnnotation = Microsoft.AspNetCore.OData.Edm.Nav
 using Microsoft.AspNetCore.OData.Common;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.OData.Deltas;
-using System.Xml.Linq;
 
 namespace Microsoft.AspNetCore.OData.Formatter.Serialization
 {
@@ -799,8 +798,12 @@ namespace Microsoft.AspNetCore.OData.Formatter.Serialization
             IEnumerable<ODataNestedResourceInfo> navigationLinks = CreateNavigationLinks(selectExpandNode.SelectedNavigationProperties, resourceContext);
             foreach (ODataNestedResourceInfo navigationLink in navigationLinks)
             {
-                await writer.WriteStartAsync(navigationLink).ConfigureAwait(false);
-                await writer.WriteEndAsync().ConfigureAwait(false);
+                if (resourceContext.WriteNavigationLinks)
+                {
+                    await writer.WriteStartAsync(navigationLink).ConfigureAwait(false);
+                    await writer.WriteEndAsync().ConfigureAwait(false);
+
+                }
             }
         }
 
@@ -1214,6 +1217,15 @@ namespace Microsoft.AspNetCore.OData.Formatter.Serialization
                     if (navigationUrl != null)
                     {
                         navigationLink.Url = navigationUrl;
+                        resourceContext.WriteNavigationLinks = true;
+                    }
+                    else if (writeContext.MetadataLevel == ODataMetadataLevel.Full)
+                    {
+                        resourceContext.WriteNavigationLinks = true;
+                    }
+                    else
+                    {
+                        resourceContext.WriteNavigationLinks = false;
                     }
                 }
             }
--- a/test/Microsoft.AspNetCore.OData.Tests/Formatter/Serialization/ODataResourceSerializerTests.cs
+++ b/test/Microsoft.AspNetCore.OData.Tests/Formatter/Serialization/ODataResourceSerializerTests.cs
@@ -322,8 +322,11 @@ namespace Microsoft.AspNetCore.OData.Tests.Formatter.Serialization
             serializer.Setup(s => s.CreateNavigationLink(selectExpandNode.SelectedNavigationProperties.ElementAt(0), It.IsAny<ResourceContext>())).Verifiable();
             serializer.Setup(s => s.CreateNavigationLink(selectExpandNode.SelectedNavigationProperties.ElementAt(1), It.IsAny<ResourceContext>())).Verifiable();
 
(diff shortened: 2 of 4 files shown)


habbes:
@ElizabethOkerio would be great to also show CPU improvements if that's available

test/Microsoft.AspNetCore.OData.Tests/Formatter/Serialization/ODataResourceSerializerTests.cs
```diff
@@ -46,6 +46,7 @@ public class ODataResourceSerializerTests
         private Order _order;
         private ODataResourceSerializer _serializer;
         private ODataSerializerContext _writeContext;
+        private ODataSerializerContext writeContext;
```
habbes, line 49: I think it's strange to have to private `writerContext` fields, where the only difference in the name is the leading `_`. Maybe keep just one of them, or you'll keep both, you can give one a more descriptive name, like `fullMetadataWriteContext`. But since the writer context is initialized per test and shared across methods, maybe it's better not to introduce a new field.

test/Microsoft.AspNetCore.OData.Tests/Formatter/Serialization/ODataResourceSerializerTests.cs
```diff
@@ -85,6 +86,7 @@ public ODataResourceSerializerTests()
             _serializer = new ODataResourceSerializer(_serializerProvider);
             _path = new ODataPath(new EntitySetSegment(_customerSet));
             _writeContext = new ODataSerializerContext() { NavigationSource = _customerSet, Model = _model, Path = _path };
+            writeContext = new ODataSerializerContext() { NavigationSource = _customerSet, Model = _model, Path = _path, MetadataLevel = ODataMetadataLevel.Full};
```
habbes, line 89: These two are confusing. Maybe rename the latter to `fullMetadataWriteContext`, or remove it altogether.

habbes: Another alternative to creating the field is to instantiate the context in the methods that need it, since only 2 methods need the full metadata.

habbes (approved):

habbes (changes requested):
If the workload is not following conventions, they could decide to write navigation links even on minimal metadata. So we need a way to ensure that we only skip writing navigation links if the service is following conventions.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1214,6 +1217,15 @@ public virtual ODataNestedResourceInfo CreateNavigationLink(IEdmNavigationProper
                     if (navigationUrl != null)
                     {
                         navigationLink.Url = navigationUrl;
+                        resourceContext.WriteNavigationLinks = true;
```
habbes, line 1220: Does this cover the scenario where the customer is not following conventions?

ElizabethOkerio: yes.

ElizabethOkerio: The method that returns the navigationUrl already does the check..

habbes (approved):

PR #1081, opened 2023-11-01
Perf-improvement: CreateEtag method

ElizabethOkerio:
The CreateEtag method creates an ETag for a given entity.  This method is called for every resource and seems to take-up a lot of CPU time even in cases where ETags are not being created. 

c1

This is how the CreateEtag method looks like: 
```c#
 public virtual string CreateETag(ResourceContext resourceContext)
 {
     if (resourceContext == null)
     {
         throw Error.ArgumentNull(nameof(resourceContext));
     }

     if (resourceContext.Request != null)
     {
         IEdmModel model = resourceContext.EdmModel;
         IEdmNavigationSource navigationSource = resourceContext.NavigationSource;

         IEnumerable<IEdmStructuralProperty> concurrencyProperties;
         if (model != null && navigationSource != null)
         {
             concurrencyProperties = model.GetConcurrencyProperties(navigationSource).OrderBy(c => c.Name);
         }
         else
         {
             concurrencyProperties = Enumerable.Empty<IEdmStructuralProperty>();
         }

         IDictionary<string, object> properties = new Dictionary<string, object>();
         foreach (IEdmStructuralProperty etagProperty in concurrencyProperties)
         {
             properties.Add(etagProperty.Name, resourceContext.GetPropertyValue(etagProperty.Name));
         }

         return resourceContext.Request.CreateETag(properties, resourceContext.TimeZone);
     }

     return null;
 }
```

We call `return resourceContext.Request.CreateETag(properties, resourceContext.TimeZone)` every time even when the properties are null. This call involves getting the ETag handler from the DI container which seems to take up some CPU time.. Adding a check to only call this method when the concurrency properties are not null seems to improve CPU usage a bit. 

c2

Comparing the before and after CPU usage, you can see an improvement in CPU usage of about `1.37%`

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Edm/EdmModelAnnotationExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Edm/EdmModelAnnotationExtensions.cs
@@ -24,8 +24,6 @@ namespace Microsoft.AspNetCore.OData.Edm
     /// </summary>
     public static class EdmModelAnnotationExtensions
     {
-        private static ConcurrentDictionary<IEdmNavigationSource, IEnumerable<IEdmStructuralProperty>> _cachedConcurrencyProperties = new ConcurrentDictionary<IEdmNavigationSource, IEnumerable<IEdmStructuralProperty>>();
-        
         /// <summary>
         /// Gets the Org.OData.Core.V1.AcceptableMediaTypes
         /// </summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.OData/Edm/EdmModelAnnotationExtensions.cs
```diff
@@ -79,55 +82,63 @@ public static IEnumerable<IEdmStructuralProperty> GetConcurrencyProperties(this
...
+                {
+                    concurrencyProperties = new ConcurrencyPropertiesAnnotation();
+                    model.SetAnnotationValue(model, concurrencyProperties);
+                }
+                
+                if (concurrencyProperties.TryGetValue(navigationSource, out cachedProperties))
+                {
+                    _cachedConcurrencyProperties[navigationSource] = cachedProperties;
+                    return cachedProperties;
+                }
```
habbes, line 100: Why not make the `concurrencyProperties` the static cache instead of creating a new one?

src/Microsoft.AspNetCore.OData/Edm/EdmModelAnnotationExtensions.cs
```diff
@@ -79,55 +82,63 @@ public static IEnumerable<IEdmStructuralProperty> GetConcurrencyProperties(this
...
-            if (concurrencyProperties == null)
-            {
-                concurrencyProperties = new ConcurrencyPropertiesAnnotation();
-                model.SetAnnotationValue(model, concurrencyProperties);
-            }
-
-            IEnumerable<IEdmStructuralProperty> cachedProperties;
-            if (concurrencyProperties.TryGetValue(navigationSource, out cachedProperties))
+            
+            if (!_cachedConcurrencyProperties.TryGetValue(navigationSource, out IEnumerable<IEdmStructuralProperty> cachedProperties))
```
habbes, line 86: I'm not convinced about this cache, while less CPU time is reported, can we also measure the impact on clock time/latency. Since the versioning dictionary is designed to eliminate clock contention, and ConcurrentDictionary does use locks, I think a fairer comparison would be one that considers latency/throughput in highly concurrent scenario and not just plain CPU time.

Another problem I see with this cache is that it does not take the model into consideration. The `ConcurrencyPropertiesAnnotation` is probably also a concurrent dictionary internally, so essentially, we have a cache per model. But you're replacing that with a single global cache irrespective of the model. So these not are equivalents. If you can guarantee that the same cache can hold entries from different models without conflicts, maybe your approach can work in the general case, otherwise it would not.

If you want to save the cost of accessing the annotations cache, maybe a safer approach would be store the cache on the context instead of making it static.

src/Microsoft.AspNetCore.OData/Edm/EdmModelAnnotationExtensions.cs
```diff
@@ -79,55 +82,63 @@ public static IEnumerable<IEdmStructuralProperty> GetConcurrencyProperties(this
...
-                                IEdmProperty edmProperty = entityType.FindProperty(propertyName);
-                                IEdmStructuralProperty structuralProperty = edmProperty as IEdmStructuralProperty;
-                                if (structuralProperty != null)
+                                IEdmPathExpression pathExpression = property as IEdmPathExpression;
+                                if (pathExpression != null)
                                 {
-                                    results.Add(structuralProperty);
+                                    // So far, we only consider the single path, because only the direct properties from declaring type are used.
+                                    // However we have an issue tracking on: https://github.com/OData/WebApi/issues/472
+                                    string propertyName = pathExpression.PathSegments.First();
```
habbes, line 121: Is `PathSegments` an `IEnumerable` or `IList` or `ICollection`? I wonder wther there's room to index the list instead of call `First()`. Not sure if it's significant though.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -776,7 +776,10 @@ public virtual string CreateETag(ResourceContext resourceContext)
                     properties.Add(etagProperty.Name, resourceContext.GetPropertyValue(etagProperty.Name));
                 }
 
-                return resourceContext.Request.CreateETag(properties, resourceContext.TimeZone);
+                if (properties.Count != 0)
```
habbes, line 779: This one looks good to me.

src/Microsoft.AspNetCore.OData/Edm/EdmModelAnnotationExtensions.cs
ElizabethOkerio: I think storing in the context didn't have much improvements as a global cache. I think I can leave out this cache.

src/Microsoft.AspNetCore.OData/Edm/EdmModelAnnotationExtensions.cs
ElizabethOkerio: concurrencyProperties is already  a cache.. I think I'll leave out this change and let perf improvements be made to the VersioningDictionary in Microsoft.OData.Edm.

src/Microsoft.AspNetCore.OData/Edm/EdmModelAnnotationExtensions.cs
```diff
@@ -6,6 +6,7 @@
 //------------------------------------------------------------------------------
 
 using System;
+using System.Collections.Concurrent;
```
gathogojr, line 9: Is this import being used?

gathogojr:
@ElizabethOkerio In your pull request description you state:
> Having another cache in Microsoft.AspNetCore.OData to avoid calling this method: model.GetAnnotationValue<ConcurrencyPropertiesAnnotation>(model); every time seems to have a slight improvement on CPU usage.

Where is the change for the other cache?

ElizabethOkerio:
@gathogojr  I decided to leave it out since there is on-going work to improve the VersioningDictionary from the Edm lib.

habbes:
@ElizabethOkerio could you share the CPU usage for the new changes.

habbes (approved):

gathogojr (approved):

PR #1082, opened 2023-11-01
support ActionResult<IAsyncEnumerable<T>>

ElizabethOkerio:
We added support for the `IAsyncEnumerable` interface. The current support can only be used like this: 

```c#

[HttpGet] 
 [EnableQuery] 
 public IAsyncEnumerable<Customer> Get() 
 { 
     return this.context.Customers.AsAsyncEnumerable(); 
 } 
```
This PR adds support for this format: 

```c#
[EnableQuery] 
public ActionResult<IAsyncEnumerable<Customer>> Get() 
{ 
    var query = this.context.Customers.AsAsyncEnumerable(); 
    return this.Ok(query); 
} 
```

In the second format, the declared type of the returned object is null `ObjectResult responseContent = actionExecutedContext.Result as ObjectResult`. This is the type that is used to determine how to retrieve items from the database during serialization. That's if the type is `IAsyncEnumerable<T>` then we use `await foreach` otherwise we just use a `foreach`. 

The changes here get the type `T` in `ActionResult<T>` and check if `T` is of type `IAsyncEnumerable<>`. If it is, then we update the response content's DeclaredType to `IAsyncEnumerable<>`

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/EnableQueryAttribute.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/EnableQueryAttribute.cs
@@ -19,7 +19,6 @@ using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Abstractions;
 using Microsoft.AspNetCore.Mvc.Controllers;
-using Microsoft.AspNetCore.Mvc.Diagnostics;
 using Microsoft.AspNetCore.Mvc.Filters;
 using Microsoft.AspNetCore.Mvc.Infrastructure;
 using Microsoft.AspNetCore.OData.Common;
@@ -260,15 +259,7 @@ namespace Microsoft.AspNetCore.OData.Query
                     ControllerActionDescriptor controllerActionDescriptor = actionDescriptor as ControllerActionDescriptor;
                     Type returnType = controllerActionDescriptor.MethodInfo.ReturnType;
 
-                    if (returnType.IsGenericType && returnType.GetGenericTypeDefinition() == typeof(ActionResult<>))
-                    {
-                        returnType = returnType.GetGenericArguments().First();
-
-                        if (returnType.IsGenericType && returnType.GetGenericTypeDefinition() == typeof(IAsyncEnumerable<>))
-                        {
-                            responseContent.DeclaredType = returnType;
-                        }
-                    }
+                    responseContent.DeclaredType = returnType;
 
                     if (responseContent != null)
                     {
--- a/src/Microsoft.AspNetCore.OData/Query/EnableQueryAttribute.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/EnableQueryAttribute.cs
@@ -259,7 +259,15 @@ namespace Microsoft.AspNetCore.OData.Query
                     ControllerActionDescriptor controllerActionDescriptor = actionDescriptor as ControllerActionDescriptor;
                     Type returnType = controllerActionDescriptor.MethodInfo.ReturnType;
 
-                    responseContent.DeclaredType = returnType;
+                    if (returnType.IsGenericType && returnType.GetGenericTypeDefinition() == typeof(ActionResult<>))
+                    {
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.OData/Query/EnableQueryAttribute.cs
```diff
@@ -19,6 +19,7 @@
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Abstractions;
 using Microsoft.AspNetCore.Mvc.Controllers;
+using Microsoft.AspNetCore.Mvc.Diagnostics;
```
gathogojr, line 22: I don't think this import is used

src/Microsoft.AspNetCore.OData/Query/EnableQueryAttribute.cs
```diff
@@ -255,6 +256,20 @@ public override void OnActionExecuted(ActionExecutedContext actionExecutedContex
                 if (statusCodeResult?.StatusCode == null || IsSuccessStatusCode(statusCodeResult.StatusCode.Value))
                 {
                     ObjectResult responseContent = actionExecutedContext.Result as ObjectResult;
+
+                    ControllerActionDescriptor controllerActionDescriptor = actionDescriptor as ControllerActionDescriptor;
```
gathogojr, line 260: Curious... Our library is currently supports the following without the code you added:
```csharp
public ActionResult<IEnumerable<Customer>> Get()
{
   // ...
}
```
How is it that we need the code you added to support a return type of `ActionResult<IAsyncEnumerable<T>>` but we didn't/don't need it to support a return type of `ActionResult<IEnumerable<T>>`?
Is there a feature gap in the ASP.NET Core framework such that a return type of `ActionResult<IEnumerable<T>>` is supported out of the box while a return type of `ActionResult<IAsyncEnumerable<T>>` isn't?
Or do we need to identify what tweak is in place to make `ActionResult<IEnumerable<T>>` work and apply a similar tweak for `ActionResult<IAsyncEnumerable<T>>`?

ElizabethOkerio: The reason why this works is because, before adding support for `IAsyncEnumerable`, we converted the object to be serialized to an `IEnumerable` and used a `foreach` to enumerate the `IEnumerable`. 

When we added support for IAsyncEnumerable, then we added a check during serialization that will check the type of the object being serialized.. If the type is an `IAsyncEnumerable` then an `await foreach` will be used,,, otherwise a normal `foreach` is used..

Without this fix, when we check the type of the object being serialized, since the declared type of the object is null and we can't tell whether it is an `IAsyncEnumerable` or not we default to `IEnumerable` and use `foreach` to enumerate the  collection which is not suppossed to be.

gathogojr: @ElizabethOkerio Would it make better sense to set the declared type even in the case of `ActionResult<IEnumerable<T>>` so we don't have to rely on `null` to default to `IEnumerable`?

ElizabethOkerio: This means I can remove the check that I had and just set the declared type in all scenarios..

ElizabethOkerio: I think that since all the other return types can be represented as `IEnumerable` and are enumerated using `foreach` it's ok not to worry about the type here.. This is because then we'll have to start checking for the return types for all the other scenarios like `ActionResult<Customer>` for example.

gathogojr: If we are in control of all the callers, then that's okay. If there's any chance we can come across a null (e.g. through overriding/inheritance) then it'd be better to have `null` and `ActionResult<IEnumerable<T>>` mapped to the default/current (`foreach`) behaviour.

gathogojr (approved):
:shipit:

xuzhg (approved):

PR #1088, opened 2023-11-09
Generate valid id/read/edit links for expanded non-contained resource with a contained navigation source

gathogojr:
This pull request fixes https://github.com/OData/odata.net/issues/2273.

### Description
In the described scenario, a `Site` entity type has a **contained** `Plants` navigation property. The `Plant` entity type in turn has a **contained** `Pipelines` navigation property. The `Pipeline` entity type in turn has a **non-contained** single-valued `Plant` navigation property.

The **non-contained** `Plant` navigation property is bound to the `Plants` **contained** entity set via the Edm path `Sites/Plants`:

```csharp
namespace Core
{
    public abstract class EntityBase
    {
        [Key]
        public int Id { get; set; }
        public Dictionary<string, object>? Attributes { get; set; }
    }

    public class Site : EntityBase
    {
        [Contained]
        public IEnumerable<Plant>? Plants { get; set; }
    }

    public class Plant : EntityBase
    {
        public Site Site { get; set; }
        [Contained]
        public IEnumerable<PipelineBase> Pipelines { get; set; }
    }

    public abstract class PipelineBase : EntityBase
    {
        public Plant Plant { get; set; }
    }
}

namespace Plant1
{
    public class Pipeline : PipelineBase
    {
        public int? Length { get; set; }
    }
}
```
The `Sites` entity set and the navigation property bindings are as follows:
```xml
<EntitySet Name=""Sites"" EntityType=""Core.Site"">
    <NavigationPropertyBinding Path=""Plants/Pipelines/Plant"" Target=""Sites/Plants""/>
    <NavigationPropertyBinding Path=""Plants/Site"" Target=""Sites""/>
</EntitySet>
```
If the `$format` or `Accept` request headers specify `odata.metadata=full`, and we attempt to expand the `Plant` navigation property, i.e., `/Sites(1)/Plants(1)/Pipelines(1)?$expand=Plant` - where `Plant` is non-contained, failure to use the link builder to build the id/read/edit links results into invalid links since we generate the links by appending `/Plants(1)` to the parent `Pipeline` resource id - `/Sites(1)/Plants(1)/Pipelines(1)` => `/Sites(1)/Plants(1)/Pipelines(1)/Plants(1)`. This link is invalid since there's no `Plants` navigation property on `Pipeline` entity type.

The following line causes us not to use the link builder to generate the id/read/edit links if the `resourceContext.NavigationSource` is a contained entity set:

https://github.com/OData/AspNetCoreOData/blob/d94daadf7efdc55d9e0533f21a9d333322cfc900/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs#L584

This PR fixes the issue by causing the link builder to be used if `resourceContext.NavigationSource` is a contained entity set and an expanded non-contained resource is being written. A navigation property binding should exist and a navigation link builder should be on hand to build the links.

Tests verify that expected behaviour is observed even when the non-contained navigation property is in a nested expand.

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
@@ -581,7 +581,18 @@ namespace Microsoft.AspNetCore.OData.Formatter.Serialization
 
             if (resourceContext.StructuredType.TypeKind == EdmTypeKind.Entity && resourceContext.NavigationSource != null)
             {
-                if (!(resourceContext.NavigationSource is IEdmContainedEntitySet))
+                // Condition 1. If resourceContext.NavigationSource is a contained entity set
+                //    and a contained resource is being written, the id/read/edit links can be derived
+                //    from the entity set or parent resource, i.e., no need to use link builder to build the links.
+                // Condition 2. If resourceContext.NavigationSource is a contained entity set
+                //    but an expanded non-contained resource is being written,
+                //    deriving the id/read/edit links from the entity set or parent resource will
+                //    most likely result into invalid links.
+                //    A navigation property binding should exist and we should try
+                //    to use the navigation link builder to build the links.
+                // NOTE: resourceContext.SerializerContext.NavigationProperty will not be null when writing an expanded resource
+                if (!(resourceContext.NavigationSource is IEdmContainedEntitySet)
+                    || resourceContext.SerializerContext.NavigationProperty?.ContainsTarget == false)
                 {
                     IEdmModel model = resourceContext.SerializerContext.Model;
                     NavigationSourceLinkBuilderAnnotation linkBuilder = EdmModelLinkBuilderExtensions.GetNavigationSourceLinkBuilder(model, resourceContext.NavigationSource);
--- a/test/Microsoft.AspNetCore.OData.E2E.Tests/MetadataProperties/MetadataPropertiesController.cs
+++ b/test/Microsoft.AspNetCore.OData.E2E.Tests/MetadataProperties/MetadataPropertiesController.cs
@@ -1,0 +1,98 @@
+//-----------------------------------------------------------------------------
+// <copyright file="MetadataPropertiesController.cs" company=".NET Foundation">
+//      Copyright (c) .NET Foundation and Contributors. All rights reserved.
+//      See License.txt in the project root for license information.
+// </copyright>
+//------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.OData.E2E.Tests.MetadataProperties.Core;
+using Microsoft.AspNetCore.OData.Query;
+using Microsoft.AspNetCore.OData.Routing.Controllers;
+
+namespace Microsoft.AspNetCore.OData.E2E.Tests.MetadataProperties
+{
+    [Route("NonContainedNavPropInContainedNavSource")]
+    [Route("ContainedNavPropInContainedNavSource")]
+    public class SitesController : ODataController
+    {
+        [EnableQuery]
+        [HttpGet("Sites")]
+        public ActionResult<IEnumerable<Site>> Get()
+        {
+            return MetadataPropertiesDataSource.Sites;
+        }
+
+        [EnableQuery]
+        [HttpGet("Sites({key})")]
+        public ActionResult<Site> Get(int key)
+        {
+            var site = MetadataPropertiesDataSource.Sites.SingleOrDefault(d => d.Id == key);
+
+            if (site == null)
+            {
+                return NotFound();
+            }
+
+            return site;
+        }
+
+        [EnableQuery]
+        [HttpGet("Sites({key})/Plants")]
+        public ActionResult<IEnumerable<Plant>> GetPlants(int key)
+        {
+            var site = MetadataPropertiesDataSource.Sites.SingleOrDefault(d => d.Id == key);
+
+            if (site == null || site.Plants == null)
+            {
+                return Enumerable.Empty<Plant>().ToList();
+            }
+
+            return site.Plants.ToList();
+        }
+
+        [EnableQuery]
+        [HttpGet("Sites({siteKey})/Plants({plantKey})")]
+        public ActionResult<Plant> GetPlant(int siteKey, int plantKey)
+        {
+            var plant = MetadataPropertiesDataSource.Sites.SingleOrDefault(d => d.Id == siteKey)?.Plants?.SingleOrDefault(d => d.Id == plantKey);
+
+            if (plant == null)
+            {
+                return NotFound();
+            }
+
+            return plant;
+        }
+
+        [EnableQuery]
+        [HttpGet("Sites({siteKey})/Plants({plantKey})/Pipelines")]
+        public ActionResult<IEnumerable<PipelineBase>> GetPipelines(int siteKey, int plantKey)
+        {
+            var plant = MetadataPropertiesDataSource.Sites.SingleOrDefault(d => d.Id == siteKey)?.Plants?.SingleOrDefault(d => d.Id == plantKey);
+
+            if (plant == null || plant.Pipelines == null)
+            {
+                return Enumerable.Empty<PipelineBase>().ToList();
+            }
+
+            return plant.Pipelines.ToList();
+        }
+
+        [EnableQuery]
+        [HttpGet("Sites({siteKey})/Plants({plantKey})/Pipelines({pipelineKey})")]
+        public ActionResult<PipelineBase> GetPlantPipeline(int siteKey, int plantKey, int pipelineKey)
+        {
+            var pipeline = MetadataPropertiesDataSource.Sites.SingleOrDefault(d => d.Id == siteKey)?.Plants?.SingleOrDefault(d => d.Id == plantKey)?.Pipelines?.SingleOrDefault(d => d.Id == pipelineKey);
+
+            if (pipeline == null)
+            {
+                return NotFound();
+            }
+
+            return pipeline;
+        }
+    }
+}
--- a/test/Microsoft.AspNetCore.OData.E2E.Tests/MetadataProperties/MetadataPropertiesDataModel.cs
+++ b/test/Microsoft.AspNetCore.OData.E2E.Tests/MetadataProperties/MetadataPropertiesDataModel.cs
@@ -1,0 +1,50 @@
+//-----------------------------------------------------------------------------
+// <copyright file="MetadataPropertiesDataModel.cs" company=".NET Foundation">
+//      Copyright (c) .NET Foundation and Contributors. All rights reserved.
+//      See License.txt in the project root for license information.
+// </copyright>
+//------------------------------------------------------------------------------
+
+namespace Microsoft.AspNetCore.OData.E2E.Tests.MetadataProperties.Core
+{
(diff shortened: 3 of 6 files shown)


test/Microsoft.AspNetCore.OData.E2E.Tests/DollarExpand/DollarExpandDataModel.cs
```diff
@@ -0,0 +1,49 @@
...
+        public IEnumerable<PipelineBase> Pipelines { get; set; }
+    }
+
+    public abstract class PipelineBase : EntityBase
+    {
+        public Plant Plant { get; set; }
+    }
+}
+
+namespace Microsoft.AspNetCore.OData.E2E.Tests.DollarExpand.Plant1
```
xuzhg, line 41: what's this part?

test/Microsoft.AspNetCore.OData.E2E.Tests/DollarExpand/DollarExpandEdmModel.cs
```diff
@@ -0,0 +1,75 @@
...
+            var pipelinesNavigationProperty = plantsContainedEntitySet.EntityType().DeclaredNavigationProperties().Single(d => d.Name.Equals("Pipelines"));
+            var pipelineContainedEntitySet = plantsContainedEntitySet.FindNavigationTarget(pipelinesNavigationProperty, new EdmPathExpression("Plants", "Pipelines"));
+            var plantNavigationProperty = pipelineContainedEntitySet.EntityType().DeclaredNavigationProperties().Single(d => d.Name.Equals("Plant"));
+
+            sitesEntitySet.AddNavigationTarget(siteNavigationProperty, sitesEntitySet, new EdmPathExpression("Plants", "Site"));
+            sitesEntitySet.AddNavigationTarget(plantNavigationProperty, plantsContainedEntitySet, new EdmPathExpression("Plants", "Pipelines", "Plant"));
+            
+            // Expected entity set definition
+
+            //<EntitySet Name=""Sites"" EntityType=""Microsoft.AspNetCore.OData.E2E.Tests.DollarExpand.Core.Site"">
```
xuzhg, line 47: clean

xuzhg (dismissed):

test/Microsoft.AspNetCore.OData.E2E.Tests/DollarExpand/DollarExpandDataModel.cs
gathogojr: @xuzhg I wanted to repro the customer's scenario. In their case they had one type defined in a different namespace. Doing this helps verify the type segment on the different links that are generated.

test/Microsoft.AspNetCore.OData.E2E.Tests/DollarExpand/DollarExpandController.cs
```diff
@@ -0,0 +1,98 @@
+//-----------------------------------------------------------------------------
+// <copyright file="DollarExpandController.cs" company=".NET Foundation">
```
ElizabethOkerio, line 2: This name doesn't really give an overview of what these tests are for.. Use a more descriptive name for this scenario for all the files.

test/Microsoft.AspNetCore.OData.E2E.Tests/DollarExpand/DollarExpandDataModel.cs
ElizabethOkerio: I think we should also have a test with the type in the same namespace

test/Microsoft.AspNetCore.OData.E2E.Tests/DollarExpand/DollarExpandEdmModel.cs
```diff
@@ -0,0 +1,68 @@
...
+
+namespace Microsoft.AspNetCore.OData.E2E.Tests.DollarExpand
+{
+    public class DollarExpandEdmModel
+    {
+        /// <summary>
+        /// Returns scenario 1 model where Site and Plant navigation properties are non-contained.
+        /// </summary>
+        /// <returns>Returns Edm model.</returns>
+        public static IEdmModel GetScenario1EdmModel()
```
ElizabethOkerio, line 22: nit: Can this name be more descriptive of what is going on for someone who doesn't understand what scenario 1 is.

test/Microsoft.AspNetCore.OData.E2E.Tests/DollarExpand/DollarExpandEdmModel.cs
```diff
@@ -0,0 +1,68 @@
...
+            sitesEntitySet.AddNavigationTarget(plantNavigationProperty, plantsContainedEntitySet, new EdmPathExpression("Plants", "Pipelines", "Plant"));
+
+            return model;
+        }
+
+        /// <summary>
+        /// Returns scenario 1 model where Site and Plant navigation properties are contained.
+        /// </summary>
+        /// <returns>Returns Edm model.</returns>
+        public static IEdmModel GetScenario2EdmModel()
```
ElizabethOkerio, line 52: This too.

test/Microsoft.AspNetCore.OData.E2E.Tests/DollarExpand/DollarExpandEdmModel.cs
```diff
@@ -0,0 +1,68 @@
...
+            var plantNavigationProperty = pipelineContainedEntitySet.EntityType().DeclaredNavigationProperties().Single(d => d.Name.Equals("Plant"));
+
+            sitesEntitySet.AddNavigationTarget(siteNavigationProperty, sitesEntitySet, new EdmPathExpression("Plants", "Site"));
+            sitesEntitySet.AddNavigationTarget(plantNavigationProperty, plantsContainedEntitySet, new EdmPathExpression("Plants", "Pipelines", "Plant"));
+
+            return model;
+        }
+
+        /// <summary>
+        /// Returns scenario 1 model where Site and Plant navigation properties are contained.
```
ElizabethOkerio, line 49:
```suggestion
        /// Returns scenario 2 model where Site and Plant navigation properties are contained.
```

test/Microsoft.AspNetCore.OData.E2E.Tests/DollarExpand/DollarExpandEdmModel.cs
gathogojr: @ElizabethOkerio I couldn't find a name or route prefix that could summarize the scenario so I added the docstring to the method to describe the scenario. What name would you propose?

test/Microsoft.AspNetCore.OData.E2E.Tests/DollarExpand/DollarExpandController.cs
gathogojr: @ElizabethOkerio What would you propose?

test/Microsoft.AspNetCore.OData.E2E.Tests/DollarExpand/DollarExpandEdmModel.cs
gathogojr: I thought I could use "contained" and "noncontained" but it so happens that applies only to the two navigation properties. All other navigation properties in both scenarios are contained. It's that difficulty that I faced so I settled for "scenario1" and "scenario2" and the described each in the doc for the methods.

test/Microsoft.AspNetCore.OData.E2E.Tests/DollarExpand/DollarExpandController.cs
ElizabethOkerio: LinksForExpandedResourcesWithContainedNavSources .. this is too long but something in those lines.

test/Microsoft.AspNetCore.OData.E2E.Tests/DollarExpand/DollarExpandDataModel.cs
habbes: If the linked issue uses different namespaces, then it may be a good idea to add a link to the issue in this comment so that future readers can be able to trace this "reported scenario" that's being repro'ed by the test.

test/Microsoft.AspNetCore.OData.E2E.Tests/DollarExpand/DollarExpandTests.cs
```diff
@@ -0,0 +1,264 @@
...
+            Assert.EndsWith($"/Sites(1)/Plants(1)/Pipelines(1)/{typeofPipeline}/Plant", result.Value<string>("Plant@odata.navigationLink"));
+
+            var plant = result.GetValue("Plant") as JObject;
+            Assert.NotNull(plant);
+            Assert.Equal($"#{typeofPlant}", plant.Value<string>("@odata.type"));
+            Assert.EndsWith($"{plantResourceBase}", plant.Value<string>("@odata.id"));
+            Assert.EndsWith($"{plantResourceBase}", plant.Value<string>("@odata.editLink"));
+            Assert.Equal(1, plant.Value<int>("Id"));
+            Assert.Equal("Plant 1", plant.Value<string>("Name"));
+            Assert.EndsWith($"{scenario}/{plantResourceBase}/Site/$ref", plant.Value<string>("Site@odata.associationLink"));
```
habbes, line 88: Here there's no type segment?

gathogojr: @habbes There shouldn't be a type segment there. The `Site` navigation property is not initialized to a derived type as is the case for the `Pipelines` navigation property.

test/Microsoft.AspNetCore.OData.E2E.Tests/DollarExpand/DollarExpandDataModel.cs
gathogojr: @habbes The issue is referenced in the pull request description but I don't mind adding the link to my comment

test/Microsoft.AspNetCore.OData.E2E.Tests/DollarExpand/DollarExpandController.cs
gathogojr: @ElizabethOkerio None of the existing directories that categorize the tests under it has this level of detail - `LinksForExpandedResourcesWithContainedNavSources`. The directory name is a pointer for the kind of tests one should expect under it. I renamed the directory to `MetadataProperties` which is in essence what the fix is all about. The tests are a combination of many scenarios (contained, non-contained, expanded, nested expand, etc.) and tests for more scenario could be added with time.

test/Microsoft.AspNetCore.OData.E2E.Tests/DollarExpand/DollarExpandEdmModel.cs
gathogojr: @ElizabethOkerio Changed the names to `GetEdmModelWithNonContainedNavPropInContainedNavSource` and `GetEdmModelWithContainedNavPropInContainedNavSource`

test/Microsoft.AspNetCore.OData.E2E.Tests/DollarExpand/DollarExpandEdmModel.cs
gathogojr: @ElizabethOkerio Addressed. See https://github.com/OData/AspNetCoreOData/pull/1088#discussion_r1395174669

xuzhg (approved):

ElizabethOkerio (approved):

Issue #1084, opened 2023-11-03
Custom Method with nullable properties


SirTipzy:
I experiencing problems when creating Custom Methods that should handle nullable properties.

My dbSet contain nullable properties. e.g.

`public DateTime? SomeDateTime { get; set; }`

My custom method also takes a nullable DateTime as argument. e.g.

```
public static Date? DateOnly(DateTime? date)
{
    if (date == null)
    {
        return null;
    }
    return date.Value.ToUniversalTime().Date;
}
```
An OData query with the following compute statement throws an exception.

`$apply=compute(DateOnly(SomeDateTime) as date)`

The Exception is:

"'Expression of type 'System.DateTime' cannot be used for parameter of type 'System.Nullable\`1[System.DateTime]' of method 'System.Nullable\`1[Microsoft.OData.Edm.Date] DateOnly(System.Nullable`1[System.DateTime])' (Parameter 'arg0')"

The problem lies in the method ExpressionBinderHelper.MakeFunctionCall

In this method line 228 the following method is called
```
// if the argument is of type Nullable<T>, then translate the argument to Nullable<T>.Value as none
// of the canonical functions have overloads for Nullable<> arguments.
functionCallArguments = ExtractValueFromNullableArguments(functionCallArguments);
```

Then when calling Expression.Call will fail because 'System.DateTime' is not System.Nullable`1[System.DateTime]

xuzhg:
@SirTipzy  Since you have a method call "DateOnly(...)", Can you share me your configuration about this method into the model?
Otherwise, ODL doesn't know this method?

And, if you can share a repro using a github repository, it can help us to dig more for you. Thanks.

SirTipzy:
@xuzhg I unfortunately do not have a github repo I can share.

I do not have problems for your OData Library to recognize and use my methods. I would not even get the exception I am getting if it did not know of the method. 

Here is my class I use to add my methods. I call the UseCustomODataFunctions at startup


    public static class CustomODataFunctionService
    {
        public static void UseCustomODataFunctions(this IApplicationBuilder app)
        {
            var methodInfos = typeof(CustomODataFunctions).GetMethods(BindingFlags.Static | BindingFlags.Public);
            foreach (var methodInfo in methodInfos)
            {
                RegisterCustomFunction(methodInfo);
            }          
        }

        private static void RegisterCustomFunction(MethodInfo methodInfo)
        {
            var functionName = methodInfo.Name;
[...]

PR #1089, opened 2023-11-11
Custom method can now have nullable parameters and nullable arguments

bjelbo:
See issue https://github.com/OData/AspNetCoreOData/issues/1084


bjelbo:

PR #1100, opened 2023-11-23
Refactor write links perf changes

ElizabethOkerio:
This PR refactors changes made in this PR: https://github.com/OData/AspNetCoreOData/pull/1075. I removed the `WriteNavigationLinks` property introduced in the `ResourceContext`. 

Before the change in PR 1075, we were writing navigation links when the navigationlink is not null but with these changes, we write navigation link when navigationlink url is not null (this means that either the follows convention is set to false or the metadata level is full)  and also when the navigation link url is null but metadata level is full. 

The perf improvements are the same as from the attached PR.


habbes (dismissed):

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -812,12 +812,8 @@ private async Task WriteNavigationLinksAsync(SelectExpandNode selectExpandNode,
             IEnumerable<ODataNestedResourceInfo> navigationLinks = CreateNavigationLinks(selectExpandNode.SelectedNavigationProperties, resourceContext);
             foreach (ODataNestedResourceInfo navigationLink in navigationLinks)
             {
-                if (resourceContext.WriteNavigationLinks)
-                {
-                    await writer.WriteStartAsync(navigationLink).ConfigureAwait(false);
-                    await writer.WriteEndAsync().ConfigureAwait(false);
-
-                }
+                await writer.WriteStartAsync(navigationLink).ConfigureAwait(false);
```
xuzhg, line 815: if (navigationLink != null)
{
    // ....
}

ElizabethOkerio: The check is already happening in the CreateNavigationLink method.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -1131,7 +1127,7 @@ await serializer.WriteObjectInlineAsync(propertyValue, edmProperty.Type, writer,
             foreach (IEdmNavigationProperty navProperty in navigationProperties)
             {
                 ODataNestedResourceInfo navigationLink = CreateNavigationLink(navProperty, resourceContext);
-                if (navigationLink != null)
+                if (navigationLink?.Url != null || resourceContext.SerializerContext.MetadataLevel == ODataMetadataLevel.Full)
```
xuzhg, line 1130: >resourceContext.SerializerContext.MetadataLevel == ODataMetadataLevel.Full [](http://example.com/codeflow?start=51&length=74)

I'd suggest move this check to 'CreateNavigationLink' method

ElizabethOkerio: The CreateNavigationLink can return a `NavigationLink` with a null `URL`. This navigation link can be written in cases where we have  the metadata set to full.

xuzhg: Customer can do anything without taking MetadataLevel  or isFollowsConvention into consideration. 
So, we should allow him to do anything he wants to do.

ElizabethOkerio: These changes do not change how customers are currently using this feature.. It only ensures that we do not call ODL's write methods for writing navigation links when metadata level is minimal and the service follows convention since even if we call those methods nothing gets written and it takes up a lot of CPU time for nothing.

ElizabethOkerio: Then they set followsConvention to false.. This is currently how customers write navigation links when metadata level is minimal

xuzhg (approved):

PR #1115, opened 2023-11-29
Fixes #1103: Enable any/all in $compute

### LINKED ISSUE #1103: Compute with a lambda throw exception
**Assemblies affected**
tried on 8.0.6 and 8.2.3

**Describe the bug**
Here is a repo with the problem described below:  https://github.com/vabic/ODataComputeLambdaErrorExample.git
This example is based on Sam Xu blog `$compute and $search in ASP.NET Core OData 8` [here](https://devblogs.microsoft.com/odata/compute-and-search-in-asp-net-core-odata-8/) with the starter repo [NewQueryOptionIn8](https://github.com/xuzhg/MyAspNetCore/tree/master/src/NewQueryOptionIn8) also tagged in your README. So I took it as official example.
I added a new prop on the Product model IList<string> Candys like this:
`       public IList<string> Candys { get; set; } `

```
                new Product
                {
                    Id = 3, Name = "Noodle", Color = Color.Brown, Price = 2.99, Qty = 50, Category = Category.Food, TaxRate = 0.10,
                    Location = addresses[1],
                    Sales = new List<Sale>
                    {
                        new Sale { SaleDate = new Date(1999, 8, 12), ProductId = 3, Amount = 11 },
                        new Sale { SaleDate = new Date(2018, 3, 19), ProductId = 3, Amount = 8 },
                    },
                     **Candys = new List<string>(){ "kit kat"}**
                },
                new Product
                {
                    Id = 4, Name = "Flute", Color = Color.Yellow, Price = 12.0, Qty = 75, Category = Category.Music, TaxRate = 0.14 ,
                    Location = addresses[0],
                    Sales = new List<Sale>
                    {
                        new Sale { SaleDate = new Date(2022, 12, 9), ProductId = 4, Amount = 15 },
                        new Sale { SaleDate = new Date(2022, 5, 29), ProductId = 4, Amount = 18 },
                    },
                     **Candys = new List<string>(){ "candy"}**
                },
                new Product
                {
                    Id = 5, Name = "Paper", Color = Color.White, Price = 6.99, Qty = 14 , Category = Category.Office, TaxRate = 0.24,
                    Location = addresses[2],
                    Sales = new List<Sale>
                    {
                        new Sale { SaleDate = new Date(2013, 1, 9), ProductId = 5, Amount = 2 },
                        new Sale { SaleDate = new Date(2022, 1, 1), ProductId = 5, Amount = 3 },
                    },
                     **Candys = new List<string>(){ "Snikers"}**
                },
```
I try to do a $compute on a Candy prop with a lambda expression to compute who has Snikers and I get an error : 
OData Url: `http://localhost:5102/odata/products?$select=HasSnikers&$compute=Candys/any(r:r eq 'Snikers') as HasSnikers`
Error: `"The query specified in the URI is not valid. The binary operator Equal is not defined for the types 'NewQueryOptionIn8.Models.Product' and 'System.String'.",`

**Reproduce steps**
clone repo, start the solution, use Postman or equivalent.

**Screenshots**
image

xuzhg:
Fixes #1103: Compute with lambda expression

When we $select a computed property from a $compute containing a lambda expression, we should be careful to process the "source" for rangeVariable.

So, for example: `
http://localhost:5102/odata/products?$select=HasSnikers&$compute=Candys/any(r:r eq 'Snikers') as HasSnikers
`

The compute.expression is an `AnyNode` and the body is the binary operator node. so, the rangeVariable ($it) of source for `AnyNode` is the toplevel source, but the rangeVariable (r) is the lambda variable.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
@@ -373,7 +373,7 @@ namespace Microsoft.AspNetCore.OData.Query.Expressions
             }
 
             // Be noted: it's used in $compute for $select and $expand.
-            if (context.Source != null)
+            if (context.Source != null && rangeVariable.Name == "$it")
             {
                 return context.Source;
             }
--- a/src/Microsoft.AspNetCore.OData/Formatter/ResourceContext.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/ResourceContext.cs
@@ -201,10 +201,10 @@ namespace Microsoft.AspNetCore.OData.Formatter
                 return null;
             }
 
-            TypedEdmStructuredObject edmStructruredObject = EdmObject as TypedEdmStructuredObject;
-            if (edmStructruredObject != null)
+            TypedEdmStructuredObject edmStructuredObject = EdmObject as TypedEdmStructuredObject;
+            if (edmStructuredObject != null)
             {
-                return edmStructruredObject.Instance;
+                return edmStructuredObject.Instance;
             }
 
             SelectExpandWrapper selectExpandWrapper = EdmObject as SelectExpandWrapper;
--- a/src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
@@ -792,6 +792,7 @@ namespace Microsoft.AspNetCore.OData.Query.Expressions
             {
                 foreach (var computedProperty in computedProperties)
                 {
+                    // $select=computed&$compute=.... as computed
                     BindComputedProperty(source, context, computedProperty, includedProperties);
                 }
             }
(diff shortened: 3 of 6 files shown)


src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
```diff
@@ -373,7 +373,7 @@ public virtual Expression BindRangeVariable(RangeVariable rangeVariable, QueryBi
             }
 
             // Be noted: it's used in $compute for $select and $expand.
-            if (context.Source != null)
+            if (context.Source != null && rangeVariable.Name == "$it")
```
habbes, line 376: Could you explain why this fixes the issue?

xuzhg: in any/all lambda expression, it contains the RangeVariable also, it could be like "r", "x" (nested lambda). 
However, for the whole $compute, it also contains the RangeVariable, it is "$it" (top-level).

When we Bind the top-level, we should use the "cached" source if apply,
When we Bind the nested lambda, we should use the lambda parameter.

So far, the "context.Source" solution looks "not a good" design, we'd like to continue to improve it when we refactor the applyBinder.

Issue #1109, opened 2023-11-28
Perf: Creating OData properties is expensive


habbes:
During serialization, a relatively high amount of CPU is used to create `ODataProperty` instances that are used to populate `ODataResource` objects. This is performed in the [`ODataSerializerPropertyHelper.CreateProperty`](https://github.com/OData/AspNetCoreOData/blob/main/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataSerializerPropertyHelper.cs#L27) method.

The most of the cost come's from the property's EDM serializer's call to `CreateODataValue`.

![image](https://github.com/OData/AspNetCoreOData/assets/8460169/be98e4fe-a55d-4090-ba78-a93caab34d1e)

The `ODataPrimitiveSerializer.CreateODataValue`:
- makes expensive calls to ODL's `EdmTypeSemantics`, see: https://github.com/OData/odata.net/issues/2807
- `ODataPrimitiveSerializer.ConverUnsupportedPrimitives` seems to be expensive due to its reliance on `Type.GetTypeCode()`. We should assess whether there are more efficient ways to handle CLR primitive types that are not part of ODL's core primitive types.

The `ODataEnumSerializer.CreateODataValue`:
- retrieves the `ClrEnumMemberAnnotation` from the the `IEdmModel` annotations cache. This model's annotation cache is constant performance hog that does not seem easy to optimize.
- does an expensive dictionary lookup to retrieve the EDM enum value corresponding to the CLR enum value
- makes an expensive call to `Enum.ToString()`

julealgon:
@habbes what framework version are you targeting on those tests?

habbes:
@julealgon the tests were running on net7.0, should try them on net8.0 as well.

PR #1116, opened 2023-11-30
CreateProperty perf optimizations

ElizabethOkerio:
This PR makes some improvements to issues raised in this issue: https://github.com/OData/AspNetCoreOData/issues/1109

-  ODL's EdmTypeSemantics extension methods like `IsPrimitive`, `IsDate`, `IsEnum` etc seem to be expensive. Maybe we need to create an issue to improve them from ODL side but doing the TypeKind checks directly from `AspNetCore OData` seems to be more performant in terms of CPU usage. 
- This (PR) additionally introduces a cache mechanism for storing `IEdmTypeSerializers`. This helps avoid the need for repeated look-ups of these serializers from the Dependency Injection (DI) container, a process that can be resource-intensive. The caching strategy aims to improve the loop-ups for the same `IEdmTypeReferences`.  
- In this pull request, there are modifications to optimize the determination of the size of the list holding `ODataProperties`. The aim is to eliminate the need for calling `List.AddResize`, a potentially expensive operation. This adjustment is made to enhance the performance of the code by efficiently managing the size of the list without incurring unnecessary overhead from resizing operations.

When I send this request using bombardier, I get the following profiles for before these changes and after: 
bombardier -c 50 -n 500 -l https://localhost:7034/odata/Products
Bombarding https://localhost:7034/odata/Products with 500 request(s) using 50 connection(s)

Before the changes: 
micor1

After the changes: 
micro

 As you can see there is some improvements on the `CreateProperty` method which propagates up to the `CreateResource` method.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataEnumSerializer.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataEnumSerializer.cs
@@ -56,7 +56,7 @@ namespace Microsoft.AspNetCore.OData.Formatter.Serialization
         /// <inheritdoc/>
         public sealed override ODataValue CreateODataValue(object graph, IEdmTypeReference expectedType, ODataSerializerContext writeContext)
         {
-            if (!expectedType.IsEnum())
+            if (expectedType.Definition?.TypeKind != EdmTypeKind.Enum)
             {
                 throw Error.InvalidOperation(SRResources.CannotWriteType, typeof(ODataEnumSerializer).Name, expectedType.FullName());
             }
--- a/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataSerializerContext.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataSerializerContext.cs
@@ -27,7 +27,7 @@ namespace Microsoft.AspNetCore.OData.Formatter.Serialization
     public class ODataSerializerContext
     {
         private IDictionary<object, object> _items;
-        private IDictionary<IEdmTypeReference, IODataEdmTypeSerializer> _edmTypeSerializers;
+        private Dictionary<IEdmTypeReference, IODataEdmTypeSerializer> _edmTypeSerializers;
         private ODataQueryContext _queryContext;
         private SelectExpandClause _selectExpandClause;
         private bool _isSelectExpandClauseSet;
@@ -255,7 +255,7 @@ namespace Microsoft.AspNetCore.OData.Formatter.Serialization
         /// <summary>
         /// A cache of the various edm type serializers.
         /// </summary>
-        public IDictionary<IEdmTypeReference, IODataEdmTypeSerializer> EdmTypeSerializers
+        internal Dictionary<IEdmTypeReference, IODataEdmTypeSerializer> EdmTypeSerializers
         {
             get
             {
--- a/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataSerializerContext.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataSerializerContext.cs
@@ -262,10 +262,6 @@ namespace Microsoft.AspNetCore.OData.Formatter.Serialization
                 _edmTypeSerializers = _edmTypeSerializers ?? new Dictionary<IEdmTypeReference, IODataEdmTypeSerializer>();
                 return _edmTypeSerializers;
             }
-            private set
-            {
-                _edmTypeSerializers = value;
-            }
         }
 
         /// <summary>
--- a/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataSerializerContext.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataSerializerContext.cs
@@ -27,6 +27,7 @@ namespace Microsoft.AspNetCore.OData.Formatter.Serialization
     public class ODataSerializerContext
     {
         private IDictionary<object, object> _items;
+        private IDictionary<IEdmTypeReference, IODataEdmTypeSerializer> _edmTypeSerializers;
         private ODataQueryContext _queryContext;
         private SelectExpandClause _selectExpandClause;
         private bool _isSelectExpandClauseSet;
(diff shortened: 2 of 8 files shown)


src/Microsoft.AspNetCore.OData/PublicAPI.Unshipped.txt
```diff
@@ -452,6 +452,7 @@ Microsoft.AspNetCore.OData.Formatter.Serialization.ODataSerializerContext
 Microsoft.AspNetCore.OData.Formatter.Serialization.ODataSerializerContext.ComputedProperties.get -> System.Collections.Generic.ISet<string>
 Microsoft.AspNetCore.OData.Formatter.Serialization.ODataSerializerContext.EdmProperty.get -> Microsoft.OData.Edm.IEdmProperty
 Microsoft.AspNetCore.OData.Formatter.Serialization.ODataSerializerContext.EdmProperty.set -> void
+Microsoft.AspNetCore.OData.Formatter.Serialization.ODataSerializerContext.EdmTypeSerializers.get -> System.Collections.Generic.IDictionary<Microsoft.OData.Edm.IEdmTypeReference, Microsoft.AspNetCore.OData.Formatter.Serialization.IODataEdmTypeSerializer>
```
habbes, line 455: Does this need to be public?

test/Microsoft.AspNetCore.OData.Tests/PublicApi/Microsoft.AspNetCore.OData.PublicApi.Net6.bsl
```diff
@@ -2327,6 +2327,7 @@ public class Microsoft.AspNetCore.OData.Formatter.Serialization.ODataSerializerC
 
 	System.Collections.Generic.ISet`1[[System.String]] ComputedProperties  { public get; }
 	Microsoft.OData.Edm.IEdmProperty EdmProperty  { public get; public set; }
+	System.Collections.Generic.IDictionary`2[[Microsoft.OData.Edm.IEdmTypeReference],[Microsoft.AspNetCore.OData.Formatter.Serialization.IODataEdmTypeSerializer]] EdmTypeSerializers  { public get; }
```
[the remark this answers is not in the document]
ElizabethOkerio: made it internal

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataEnumSerializer.cs
```diff
@@ -56,7 +56,7 @@ public override async Task WriteObjectAsync(object graph, Type type, ODataMessag
         /// <inheritdoc/>
         public sealed override ODataValue CreateODataValue(object graph, IEdmTypeReference expectedType, ODataSerializerContext writeContext)
         {
-            if (!expectedType.IsEnum())
+            if (expectedType.Definition?.TypeKind != EdmTypeKind.Enum)
```
xuzhg, line 59: `?` is necessary?

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataSerializerContext.cs
```diff
@@ -251,6 +252,22 @@ private set
...
+        /// A cache of the various edm type serializers.
+        /// </summary>
+        internal Dictionary<IEdmTypeReference, IODataEdmTypeSerializer> EdmTypeSerializers
+        {
+            get
+            {
+                _edmTypeSerializers = _edmTypeSerializers ?? new Dictionary<IEdmTypeReference, IODataEdmTypeSerializer>();
+                return _edmTypeSerializers;
+            }
+            private set
```
xuzhg, line 265: private set is use-less

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataSerializerContext.cs
```diff
@@ -251,6 +252,22 @@ private set
             }
         }
 
+        /// <summary>
+        /// A cache of the various edm type serializers.
+        /// </summary>
+        internal Dictionary<IEdmTypeReference, IODataEdmTypeSerializer> EdmTypeSerializers
```
xuzhg, line 258: ODataSerializerContext is transient object. Are you sure to cache it?
Since the IServiceProvider caches the Serializer once it's required (since it's singleton"), I don't think 
`SerializerProvider.GetEdmTypeSerializer(structuralProperty.Type);` is perf improved.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataEnumSerializer.cs
ElizabethOkerio: The definition can be null.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataSerializerContext.cs
ElizabethOkerio: We are not cachin gthe `ODataSerializerContext` rather the `IEdmTypeSerializer`. And the cache is per request. So for every property, we get it's type serializer from the DI container using `return _serviceProvider.GetRequiredService<ODataPrimitiveSerializer>();` depending on the `EdmTypeKind` we are dealing with.  If we are serializing several properties with the same type then having to do these look-ups from the DI container is a bit expensive.. And the cache improves CPU usage.. From the  CPU report above,, this call: `SerializerProvider.GetEdmTypeSerializer(structuralProperty.Type);` takes up `1.47%` of CPU units. Caching the serializers reduces the percentage to `0.02%`

habbes: I don't think it's accurate to say that using a local cache reduce the CPU units used to fetch the `IEdmTypeSerializer` from `1.47%` to `0.02%`.

In the "before" graph in the description, looking up the `IEdmTypeSerializer` in the DI container does take up `1.47%`. However, the `0.02%` seems to refer only to the retrieving the `EdmTypeSerializers` cache from the property context. But we see and additional `Dictionary.TryGetValue()` lookup that costs `0.75%` CPU in the "after" screenshot that was not there in the "before" screenshot. This is likely due to fetching the `IEdmTypeSerializer` from the local cache. Granted `0.77%` is still almost 2x improvement from `1.47`, but it's good to ensure there are no other hidden costs that came up as a result of the change.

ElizabethOkerio: @habbes yes I realized this. That's why I created different PRs for some of the improvements in this PR.

PR #1121, opened 2023-12-04
Update ConvertPrimitiveType method

ElizabethOkerio:
The `ConvertPrimitive` method checks if a primitive is a `Date`, `TimeOfDay`, `Char`, `UInt16`, `UInt32`, `UInt64`. `DateTime`, `char[]`, `XElement` and handles them differently.  

If a primitive type is not any of these types, then the value is returned. 

A lot of checks are done to check for the above-mentioned types which can be avoided if we return the value early if we know the type, we are dealing with is not any of the above-mentioned types that need special processing. 

Making this change leads to the following CPU improvements: 
Before:
before1
After:
after1

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataPrimitiveSerializer.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataPrimitiveSerializer.cs
@@ -134,6 +134,17 @@ namespace Microsoft.AspNetCore.OData.Formatter.Serialization
             }
 
             Type type = value.GetType();
+
+            // Return values for supported primitive values. 
+            if (type == typeof(string)
+                || type == typeof(int)
+                || type == typeof(bool)
+                || type == typeof(double)
+                || type == typeof(Guid))
+            {
+                return value;
+            }
+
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataPrimitiveSerializer.cs
```diff
@@ -134,6 +134,17 @@ internal static object ConvertPrimitiveValue(object value, IEdmPrimitiveTypeRefe
             }
 
             Type type = value.GetType();
+
+            // Return values for supported primitive values. 
+            if (type == typeof(string)
+                || type == typeof(int)
+                || type == typeof(bool)
+                || type == typeof(double)
+                || type == typeof(Guid))
```
xuzhg, line 143: why thoese 5 types?

xuzhg (approved):
:shipit:

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataPrimitiveSerializer.cs
ElizabethOkerio: They are the most commonly used types..

KenitoInc (approved):
LGTM

PR #1137, opened 2023-12-08
Perf Fix : GetPropertyValue

ElizabethOkerio:
The `GetPropertyValue` method retrieves the value of a property from an instance object through reflection.

Currently, we generate property getter delegates for each property type and store them in a cache. Subsequently, we invoke these cached delegates to obtain property values. This approach with delegates is more efficient compared to direct reflection. However, the utilization of a cache introduces some overhead. We employ a `ConcurrentDictionary` for caching the property getter delegates. Checking for the existence of a key in the dictionary using TryGetValue appears to be relatively resource-intensive.

I attempted to enhance the performance by replacing the existing logic, which involves creating delegates through reflection, with expressions and dynamic methods, including the use of an IL code generator. Unfortunately, this modification did not result in a significant improvement; the performance remained nearly identical. So I maintained the changes that were there and tried to improve the Concurrent Dictionary's TryGetValue by providing a custom equality comparer:  

CPU usage before any changes: 
This is the logic that uses reflection to create property getter delegates
before

CPU usage - using expressions to create property getter delegates
before1

CPU usage - using IL code 
before2

Adding a custom EqualityComparer for the `ConcurrentDictionaryCache` seems to improve the `TryGetValue` method: 
CPU usage after using a custom equality comparer: 
after

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Formatter/Value/TypedEdmStructuredObject.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Value/TypedEdmStructuredObject.cs
@@ -7,6 +7,7 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Diagnostics.Contracts;
 using System.Reflection;
 using Microsoft.AspNetCore.OData.Common;
@@ -20,8 +21,8 @@ namespace Microsoft.AspNetCore.OData.Formatter.Value
     /// </summary>
     internal abstract class TypedEdmStructuredObject : IEdmStructuredObject
     {
-        private static readonly ConcurrentDictionary<Tuple<string, Type>, Func<object, object>> _propertyGetterCache =
-            new ConcurrentDictionary<Tuple<string, Type>, Func<object, object>>();
+        private static readonly ConcurrentDictionary<(string, Type), Func<object, object>> _propertyGetterCache =
+            new ConcurrentDictionary<(string, Type), Func<object, object>>(new PropertyGetterCacheEqualityComparer());
 
         private IEdmStructuredTypeReference _edmType;
         private Type _type;
@@ -88,7 +89,7 @@ namespace Microsoft.AspNetCore.OData.Formatter.Value
             IEdmStructuredTypeReference edmType,
             IEdmModel model)
         {
-            Tuple<string, Type> key = Tuple.Create(propertyName, type);
+            (string, Type) key = (propertyName, type);
             Func<object, object> getter;
 
             if (!_propertyGetterCache.TryGetValue(key, out getter))
@@ -118,6 +119,39 @@ namespace Microsoft.AspNetCore.OData.Formatter.Value
             var helper = new PropertyHelper(property);
 
             return helper.GetValue;
+        }
+    }
+
+    /// <summary>
+    /// A custom equality comparer for the property getter cache. 
+    /// </summary>
+    internal class PropertyGetterCacheEqualityComparer : IEqualityComparer<(string, Type)>
+    {
+        public bool Equals((string, Type) x, (string, Type) y)
+        {
+            return x.Item1 == y.Item1 && x.Item2 == y.Item2;
+        }
+
+        /// <summary>
+        /// This method overrides the default GetHashCode() implementation
+        /// for a tuple of (string, Type) to provide a more effective hash code.
+        /// </summary>
+        /// <param name="obj">The tuple object to calculate a hash code for</param>
+        /// <returns>The calculated hash code. </returns>
+        public int GetHashCode((string, Type) obj)
+        {
+            unchecked
+            {
+                // The choice of 19 as the initial prime number is arbitrary but common in most hash code implementations.
+                // Multyplying by a prime number helps to reduce the chance of collisions.
+                // The hashcode of each tuple element is combined with the calculated hash to create
+                // a more unique hash code for the tuple.
+                int hash = 19;
+                hash = hash * 23 + obj.Item1.GetHashCode();
+                hash = hash * 23 + obj.Item2.GetHashCode();
+
+                return hash;
+            }
         }
     }
 }


src/Microsoft.AspNetCore.OData/Formatter/Value/TypedEdmStructuredObject.cs
```diff
@@ -120,4 +121,26 @@ public bool TryGetPropertyValue(string propertyName, out object value)
...
+            return x.Item1 == y.Item1 && x.Item2 == y.Item2;
+        }
+
+        public int GetHashCode(Tuple<string, Type> obj)
+        {
+            unchecked
+            {
+                int hash = 19;
+                hash = hash * 23 + obj.Item1.GetHashCode();
+                hash = hash * 23 + obj.Item2.GetHashCode();
```
gathogojr, line 141: Why 19 and 23? Have you also considered the possibility of the multiplication yield a result greater than `int.MaxValue`?

ElizabethOkerio: 19 and 23 because they are prime numbers and are likely to generate good hash codes that  will helo reduce the likely hood of collisions.

ElizabethOkerio: for the `int.MaxValue` I could use long instead.

ElizabethOkerio: updated please check.

src/Microsoft.AspNetCore.OData/Formatter/Value/TypedEdmStructuredObject.cs
```diff
@@ -120,4 +121,28 @@ public bool TryGetPropertyValue(string propertyName, out object value)
...
+                hash %= int.MaxValue;
```
gathogojr, line 142: By the time we get here it's already too late. The `OverflowException` will have been thrown.

ElizabethOkerio: The use of unchecked here takes care of overflow exception. Right?

gathogojr: @ElizabethOkerio Yes indeed: _In an unchecked context, the operation result is truncated by discarding any high-order bits that don't fit in the destination type. For example, in the case of addition it wraps from the maximum value to the minimum value._

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/checked-and-unchecked

Which also means you don't need this statement, no?

gathogojr: Confirm the behaviour in the case of multiplication just to be sure.

ElizabethOkerio: Yes.

ElizabethOkerio: Tried out something like this: 

```c#

int a = int.MaxValue;

unchecked
{
    Console.WriteLine(a*3); 
}

```

It gives `2147483645`. No exception gets thown..

ElizabethOkerio: Also, from this link [2147483645](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/checked-and-unchecked#operations-affected-by-the-overflow-checking-context) multiplication operation is included in the overflow-context checking

gathogojr: > Tried out something like this:
> 
> ```cs
> int a = int.MaxValue;
> 
> unchecked
> {
>     Console.WriteLine(a*3); 
> }
> ```
> 
> It gives `2147483645`. No exception gets thrown..

I meant confirm the overflow behaviour in the case of multiplication:
Here is what I tried:
```csharp
int input = int.MaxValue - 3;
int result1;
int result2;

unchecked
{
    result1 = input * 2;
    result2 = input + input;
}

Console.WriteLine(result1); // Result: -8
Console.WriteLine(result2); // Result: -8
```
Are we okay with the above outcomes?

ElizabethOkerio: Is the question whether a hash code can be a -ve?

gathogojr: Negative values for hash codes are okay. I just wanted to confirm you're aware such outcomes are possible.

src/Microsoft.AspNetCore.OData/Formatter/Value/TypedEdmStructuredObject.cs
```diff
@@ -120,4 +121,27 @@ public bool TryGetPropertyValue(string propertyName, out object value)
...
     }
+
+    /// <summary>
+    /// A custom equality comparer for the property getter cache. 
+    /// </summary>
+    internal class PropertyGetterCacheEqualityComparer : IEqualityComparer<Tuple<string, Type>>
+    {
+        public bool Equals(Tuple<string, Type> x, Tuple<string, Type> y)
+        {
+            return x.Item1 == y.Item1 && x.Item2 == y.Item2;
```
gathogojr, line 132:
```suggestion
            return x.Equals(y);
```
`Tuple` supports Equals such that:
```csharp
using System;

var tuple1 = Tuple.Create("typeofInt", typeof(int));
var tuple2 = Tuple.Create("typeofInt", typeof(int));

Console.WriteLine(tuple1.Equals(tuple2));

// Result: True
```

src/Microsoft.AspNetCore.OData/Formatter/Value/TypedEdmStructuredObject.cs
```diff
@@ -120,4 +121,27 @@ public bool TryGetPropertyValue(string propertyName, out object value)
...
+        }
+
+        public int GetHashCode(Tuple<string, Type> obj)
+        {
+            unchecked
```
habbes, line 137: Is this custom implementation better than using [`HashCode.Combine`](https://learn.microsoft.com/en-us/dotnet/api/system.hashcode.combine)?

src/Microsoft.AspNetCore.OData/Formatter/Value/TypedEdmStructuredObject.cs
```diff
@@ -120,4 +121,27 @@ public bool TryGetPropertyValue(string propertyName, out object value)
...
+            {
+                int hash = 19;
+                hash = hash * 23 + obj.Item1.GetHashCode();
+                hash = hash * 23 + obj.Item2.GetHashCode();
+
+                return hash;
+            }
```
gathogojr, line 144:
```suggestion
            return obj.GetHashCode();
```
`Tuple` supports `GetHashCode` method such that:
```csharp
using System;

var tuple1 = Tuple.Create("typeofInt", typeof(int));
var tuple2 = Tuple.Create("typeofInt", typeof(int));
var tuple3 = Tuple.Create("typeofFlot", typeof(float));

Console.WriteLine(tuple1.GetHashCode().Equals(tuple2.GetHashCode()));
// Result: True

Console.WriteLine(tuple1.GetHashCode().Equals(tuple3.GetHashCode()));
// Result: False
```

src/Microsoft.AspNetCore.OData/Formatter/Value/TypedEdmStructuredObject.cs
```diff
@@ -21,7 +22,7 @@ namespace Microsoft.AspNetCore.OData.Formatter.Value
     internal abstract class TypedEdmStructuredObject : IEdmStructuredObject
     {
         private static readonly ConcurrentDictionary<Tuple<string, Type>, Func<object, object>> _propertyGetterCache =
-            new ConcurrentDictionary<Tuple<string, Type>, Func<object, object>>();
+            new ConcurrentDictionary<Tuple<string, Type>, Func<object, object>>(new PropertyGetterCacheEqualityComparer());
```
gathogojr, line 25: Do we need to apply a custom comparer given that the `Tuple` type implements `Equals` and `GetHashCode`?

habbes, line 25: I would suggest also trying `ValueTuple<string, Type>` instead of the `Tuple` class. There are many benefits:
- `ValueTuple` is a struct, so we have a chance to reduce allocations
- Syntax sugar: `(string, Type)` instead of `Tuple<string, Type>`
- Custom optimized built-in equality comparer (but you can still compare the perf with your custom one).

src/Microsoft.AspNetCore.OData/Formatter/Value/TypedEdmStructuredObject.cs
ElizabethOkerio: You mean having something like this: 

```c#
public int GetHashCode(Tuple<string, Type> obj)
{
     return HashCode.Combine(obj.Item1, obj.Item2);
}

```

ElizabethOkerio: There is a slight difference..The custom implementation seems to perform better slightly

src/Microsoft.AspNetCore.OData/Formatter/Value/TypedEdmStructuredObject.cs
ElizabethOkerio: I have updated the Tuple to the ValueTuple and ran the benchmarks project.. 
Using the default equality comparer. This is what I get: 

be4

When I use the ValueTuple with the CustomeEquality Comparer,, this is what I get:

ElizabethOkerio: @gathogojr The reason for using the custom comparer is to optimize the hashing algorithm for the tuple elements. This will potentially reduce collisions and improve the overall performance of the concurrent dictionary especially during the key look-ups.

habbes (approved):

gathogojr (approved):
LGTM

PR #1139, opened 2023-12-14
Add collection property, action, function and dynamic properties to benchmark server project. 

ElizabethOkerio:
I've added the following to the server benchmark project: 

1. A collection property
2. An action and a function
3. A dynamic property
4. A concurrency property.

### DIFF
--- a/sample/BenchmarkServer/EdmModelBuilder.cs
+++ b/sample/BenchmarkServer/EdmModelBuilder.cs
@@ -20,7 +20,18 @@ namespace ODataPerformanceProfile
             builder.EntitySet<Supplier>("Suppliers");
             builder.EntitySet<Order>("Orders");
 
-            return builder.GetEdmModel();
+            builder.EntityType<Product>().Collection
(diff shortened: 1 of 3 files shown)


sample/BenchmarkServer/EdmModelBuilder.cs
```diff
@@ -20,6 +20,14 @@ public static IEdmModel GetEdmModel()
...
 
+            builder.EntityType<Product>().Collection
+                .Function("mostRecent")
+                .Returns<string>();
+
+            builder.EntityType<Product>()
+                .Action("rate")
+                .Parameter<int>("rating");
+
             return builder.GetEdmModel();
```
habbes, line 31: I think you should also make the model immutable since that's what we want to encourage by default.

habbes:
```c#
var model = builder.GetEdmModel();
model.MarkAsImmutable();
return model;
```

xuzhg (approved):

PR #1140, opened 2023-12-15
Support Minimal API

### LINKED ISSUE #578: ASP.NET Core OData 8.x does not work with Minimal API
**Assemblies affected**
ASP.NET Core OData 8.x

**Describe the bug**
OData does not work with Minimal API, ie without Controllers

**Reproduce steps**
```
app.MapGet("/WeatherForecast", [EnableQuery] () => Enumerable.Range(1, 5).Select((int index) => new WeatherForecast
    {
        Date = DateTime.Now.AddDays(index),
        TemperatureC = Random.Shared.Next(-20, 55),
        Summary =  new[]  {
            "Freezing", "Bracing", "Chilly", "Cool", "Mild", "Warm", "Balmy", "Hot", "Sweltering", "Scorching"
        }
    })
    .ToArray());
app.Run();
```

**Expected behavior**
I would like to use OData without Controllers at all

xuzhg:
Fixes #578

Minimal APIs are a simplified approach for building fast HTTP APIs with ASP.NET Core. You can build fully functioning REST endpoints with minimal code and configuration. Skip traditional scaffolding and avoid unnecessary controllers by fluently declaring API routes and actions. For example, the following code creates an API at the root of the web app that returns the text, "Hello World!".

```C#
var app = WebApplication.Create(args);
app.MapGet("/", () => "Hello World!");
app.Run();
```

Customer want to get the OData query options functionalities.  We can provide couple solutions

1) Provide a HttpRequest-less ODataQueryOptions< T > (Without using IEdmModel, using CLR type directly) and customer can use it in the `Delegate` to generate the Linq Expression and apply to the data. 

2) Use the existing ODataQueryOptions< T >. It convers two scenarios: 1) EdmModel based  2) Non-EdmModel (but will create EdmModel on the fly)

This PR is to support the second. 

```C#
app.MapGet("/schools", (AppDb db, ODataQueryOptions<School> options) => {
     ....
    return options.ApplyTo(db.Schools);
});
```

Then, client can send request the "/schools" using OData query options, for example, `$select`, etc.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Edm/IODataEndpointModelMapper.cs
+++ b/src/Microsoft.AspNetCore.OData/Edm/IODataEndpointModelMapper.cs
@@ -1,0 +1,25 @@
+//-----------------------------------------------------------------------------
+// <copyright file="IODataEndpointModelMapper.cs" company=".NET Foundation">
+//      Copyright (c) .NET Foundation and Contributors. All rights reserved.
+//      See License.txt in the project root for license information.
+// </copyright>
+//------------------------------------------------------------------------------
+
+using System.Collections.Concurrent;
+using Microsoft.AspNetCore.Http;
+using Microsoft.OData.Edm;
+
+namespace Microsoft.AspNetCore.OData.Edm
+{
+    /// <summary>
+    /// A cache for <see cref="Endpoint"/> and <see cref="IEdmModel"/> mapping.
+    /// It's typically used for minimal API scenario.
+    /// </summary>
+    public interface IODataEndpointModelMapper
+    {
+        /// <summary>
+        /// Gets the map between <see cref="Endpoint"/> and <see cref="IEdmModel"/>
+        /// </summary>
+        ConcurrentDictionary<Endpoint, IEdmModel> Maps { get; }
+    }
+}
--- a/src/Microsoft.AspNetCore.OData/Edm/IODataModelConfiguration.cs
+++ b/src/Microsoft.AspNetCore.OData/Edm/IODataModelConfiguration.cs
@@ -1,0 +1,25 @@
+//-----------------------------------------------------------------------------
+// <copyright file="IODataModelConfiguration.cs" company=".NET Foundation">
+//      Copyright (c) .NET Foundation and Contributors. All rights reserved.
+//      See License.txt in the project root for license information.
+// </copyright>
+//------------------------------------------------------------------------------
+
+using Microsoft.AspNetCore.Http;
+using Microsoft.OData.ModelBuilder;
+
+namespace Microsoft.AspNetCore.OData.Edm
+{
+    /// <summary>
+    /// Defines a contract used to apply extra logic on the model builder.
+    /// </summary>
+    public interface IODataModelConfiguration
+    {
+        /// <summary>
+        /// Applies model configurations using the provided builder.
+        /// </summary>
+        /// <param name="context">The HttpContext.</param>
+        /// <param name="builder">The <see cref="ODataModelBuilder">builder</see> used to apply configurations.</param>
+        void Apply(HttpContext context, ODataModelBuilder builder);
+    }
+}
--- a/src/Microsoft.AspNetCore.OData/Query/Wrapper/SelectExpandWrapperConverter.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Wrapper/SelectExpandWrapperConverter.cs
@@ -16,7 +16,7 @@ namespace Microsoft.AspNetCore.OData.Query.Wrapper
     /// <summary>
     /// Supports converting <see cref="SelectExpandWrapper{T}"/> types by using a factory pattern.
     /// </summary>
-    internal class SelectExpandWrapperConverter : JsonConverterFactory
+    public class SelectExpandWrapperConverter : JsonConverterFactory
     {
         public static readonly Func<IEdmModel, IEdmStructuredType, IPropertyMapper> MapperProvider =
             (IEdmModel model, IEdmStructuredType type) => new JsonPropertyNameMapper(model, type);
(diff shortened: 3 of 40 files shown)


habbes:
Happy to see some progress on minimal API support @xuzhg 
Would you consider adding more context or a more detailed breakdown of the PR implementation?

xuzhg:
> Happy to see some progress on minimal API support @xuzhg Would you consider adding more context or a more detailed breakdown of the PR implementation?
> 
> Also can you clarify the extent of OData features we support in this PR and what do not support (and do not plan to support). I assume OData routing is not supported at all. Does this support model binding? Delta? Or is support restricted to only `ODataQueryOptions<T>`?



> Happy to see some progress on minimal API support @xuzhg Would you consider adding more context or a more detailed breakdown of the PR implementation?
> 
> Also can you clarify the extent of OData features we support in this PR and what do not support (and do not plan to support). I assume OData routing is not supported at all. Does this support model binding? Delta? Or is support restricted to only `ODataQueryOptions<T>`?

Thanks for your feedback. Actually, the product codes in the PR changed less. 

### OData routing
In my understanding, most of customers only want to get the 'OData Query functionalities" when building the minimal API.  They don't want to know "Entity Type, Complex Type", or "Entity set, Singleton, Navigation Source, Navigation Property", concepts like these. Even thought they don't want to know about "Key Segement" and It should come after "Collection Segement". So, there's no "OData routing" at all. However, we can add "OData routing" mechanise, for example:
```
app.MapOData<ODataServiceContext>(EdmModel);
```
Where, `ODataServiceContext` is a class containing OData routings. 

### Does this support model binding?

In my understanding, ASP.NET Core doesn't add `model binding` for minimal API. So we can't add it. In the minimal API, it only contains the parameter binding, it's same as "model binding" and most of logics are same. For example, You can put [FromBody] on the parameter to bind the parameter from request body. There's new solution to customize the parameter binding. That's what I used in my PR to binding the ODataQueryOptiont< T >.

### Delta< T > ? DeltaSet< T >
I think we should support to bind those types. I also added the comments in my PR to support them. It could be a little bit same as ODataQueryOption< T > implementation. It's not complicated but I think let's do it step by step. Before implementing it, let's finish our first BrainStorm on the initial design. 

### To Be Discussion

OData payload is JSON payload. I think it's time to consider to use `Default Json serializer/deserializer`. The only thing is to implement the JsonConverter on the OData object. We can sync it more later.

cilerler (approved):

julealgon:
> OData payload is JSON payload. I think it's time to consider to use `Default Json serializer/deserializer`. The only thing is to implement the JsonConverter on the OData object. We can sync it more later.

AntonioBanderasOhGIF

Honestly... this would be the biggest improvement to OData in years and would avoid _so many problems_.

At the same time... seems ambitious to consider that alongside the Minimal API changes 😅. As much as I want that, I'd avoid doing them together, or potentially even in the same release.

robertmclaws:
It would be ABSOLUTELY AMAZING if OData switched to out-of-the-box System.Text.Json serialization with converters. I think it would improve perf by at least 50%, and I wouldn't be surprised if you saw a 200-400% improvement.

Probably needs a separate work branch tho. Hopefully the unit tests are designed to be able to handle this.

Happy to help wherever possible.

habbes:
I agree we stand to gain a lot from leveraging System.Text.Json. But it's not a trivial change given all the scenarios that OData's current serialization supports as well as adherence to the OData spec. However, this is something we're currently looking into. I'll share results from the progress soon. (see: https://github.com/OData/odata.net/issues/2798)

src/Microsoft.AspNetCore.OData/Edm/IODataModelConfiguration.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// Defines a contract used to apply extra logic on the model builder.
+    /// </summary>
+    public interface IODataModelConfiguration
+    {
+        /// <summary>
+        /// Applies model configurations using the provided builder.
+        /// </summary>
+        /// <param name="context">The HttpContext.</param>
+        /// <param name="builder">The <see cref="ODataModelBuilder">builder</see> used to apply configurations.</param>
+        void Apply(HttpContext context, ODataModelBuilder builder);
```
habbes, line 23: Do we envision that more arguments could be needed in the future, such that we could use an `ODataModelConfiguratonOptions` class to encapsulate the options, or is an `ODataModelBuilder` sufficient?

src/Microsoft.AspNetCore.OData/Query/Wrapper/SelectExpandWrapperConverter.cs
```diff
@@ -16,7 +16,7 @@ namespace Microsoft.AspNetCore.OData.Query.Wrapper
     /// <summary>
     /// Supports converting <see cref="SelectExpandWrapper{T}"/> types by using a factory pattern.
     /// </summary>
-    internal class SelectExpandWrapperConverter : JsonConverterFactory
+    public class SelectExpandWrapperConverter : JsonConverterFactory
```
habbes, line 19: Why is this made public?

src/Microsoft.AspNetCore.OData/Edm/IODataEndpointModelMapper.cs
```diff
@@ -0,0 +1,25 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.OData.Edm;
+
+namespace Microsoft.AspNetCore.OData.Edm
+{
+    /// <summary>
+    /// A cache for <see cref="Endpoint"/> and <see cref="IEdmModel"/> mapping.
+    /// It's typically used for minimal API scenario.
+    /// </summary>
+    public interface IODataEndpointModelMapper
```
habbes, line 18: What is the likelihood that the customer will want to customize this? Maybe we should make it internal until there's explicit/clear demand from the customer to make it extensible, that way it can give us room to change the design without breaking changes.

habbes (commented):
Will you add tests?

danielpastoor:
Is there any update on this pull-request?

cbrianball:
Can this please be merged?! I would really like to use minimal API with OData!

cbrianball:
One thing I have noticed is that the response from the minimal API endpoints just have the "bare" data. Meaning it doesn't have "@odata.context" field and then the result in a "value" field. If there is a way to have this, please include an example.

mike-bagheri:
What does it take for this PR to be merged?

mguinness:
@xuzhg It's been a year since you last worked on this PR, are there any blockers?  @mikepizzo The original issue has 28 upvotes, is this something that is considered?

PR #1143, opened 2023-12-19
Fixes #1127: generate skip token using computed property

### LINKED ISSUE #1127: $skipToken is replaced with $skip when $compute is used
**Assemblies affected**

**Describe the bug**
`NextLink` does not have `$skipToken` even when configured, but instead has `$skip` when a `$compute` expression is used in the request.

**Reproduce steps**
1. Enable SkipToken for your controllers.
2. Hit your API service with a `$compute` expression:
```
http://localhost:2000/beta/students?$orderby=lowername&$compute=tolower(name) as lowername
```

**Data Model**
```
public class Student
{
    [Key]
    public string Id {get; set;}
    public string Name {get; set;}
}
```

**EDM (CSDL) Model**
```
<EntityType Name="Student">
    <Key>
        <PropertyRef Name="id" />
    </Key>
    <Property Name="id" Type="Edm.String" Nullable="false" />
    <Property Name="name" Type="Edm.String" Nullable="false" />
</EntityType>
```

**Request/Response**
```
http://localhost:2000/beta/students?$orderby=lowername&$compute=tolower(name) as lowername
```

**Expected behavior**
`NextLink` in the response should have a `$skipToken`, but it instead has a `$skip`.

**Screenshots**
NA

**Additional context**
1. The default SkipToken handler is used.
2. The same code returns a `$skipToken` when used without the `$compute` expression and even with just the `$orderby` expression.

xuzhg:
Fixes #1127

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.xml
+++ b/src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.xml
@@ -11089,14 +11089,14 @@
             <param name="context">Serializer context</param>
             <returns>Returns the URI for NextPageLink. If a null object is passed for the instance, resorts to the default paging mechanism of using $skip and $top.</returns>
         </member>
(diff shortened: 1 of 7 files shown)


xuzhg:
Should be fixed at #1164

PR #1155, opened 2024-01-10
Log the invalid key or function parameter 

### LINKED ISSUE #1152: OData Query parameters are not working in 8.2.3
We have migrated our API from 2.2 to 6.0 and we are upgraded OData also to 8.2.3. 
Before update custom methods with query parameters are working well but after update no luck.
Basic OData methods and functionalities are working well and good after update too.

Below is my controller code.

```
public partial class DepartmentController : ODataController
 {
	[HttpGet, Route("api/v1/department/getDepartmentsByCompany")]
	[Microsoft.AspNetCore.OData.Query.EnableQuery]
	public Task<BaseResponse<IEnumerable<Department>>> GetDepartmentByCompany([FromQuery] int id, [FromQuery] FunctionalType? functionalType)
	{
		try
		{
			return "success data";
		}
		catch (Exception ex)
		{
			LogHelper.LogException(ex, _hostingEnvironment);
			return "Failed";
		}
	}
	
	[Microsoft.AspNetCore.Mvc.HttpGet] 
	[Microsoft.AspNetCore.OData.Query.EnableQuery]
	public System.Threading.Tasks.Task<System.Collections.Generic.IEnumerable<Department>> Get()
	{
		return _implementation.GetAllAsync();
	}
	
    [Microsoft.AspNetCore.Mvc.HttpGet]
	[Microsoft.AspNetCore.OData.Query.EnableQuery] 
	public Microsoft.AspNetCore.OData.Results.SingleResult<Department> Get(int key)
	{
		return _implementation.GetAsync(key);
	}
 }
```
And Program.cs looks like 

```
builder.Services.AddControllers(g =>
{
    g.EnableEndpointRouting = false;
}).AddOData(opt =>
    {
        var batchHandler = new DefaultODataBatchHandler();
        opt.AddRouteComponents("api/v1", DIRegister.GetEdmModels(), batchHandler);
        opt.Count().Select().Filter().OrderBy().Expand().SetMaxTop(int.MaxValue);
        opt.EnableQueryFeatures(int.MaxValue);
        opt.EnableAttributeRouting = false;
    }).AddJsonOptions(g =>
    {
        g.AllowInputFormatterExceptionMessages = true;
    });
```
I am getting below exception 

    The key value (getComapnyByDepartment) from request is not valid. The key value should be format of type 'Edm.Int32'

and inner exception as
 `Expected literal type token but found token getComapnyByDepartment`

Am i missing anything here or configuration issues.
I have gone through the MS [article](https://devblogs.microsoft.com/odata/attribute-routing-in-asp-net-core-odata-8-0-rc/) no luck here too

Sample code is [uploaded here][1]

Swagger error 
image

  [1]: https://github.com/itsmegopi/OdataTrial

xuzhg:
Fixes #1152,

if we have the logger enabled in the TryMatch. Otherwise, keep to throw exception.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Routing/Template/SegmentTemplateHelpers.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/Template/SegmentTemplateHelpers.cs
@@ -13,6 +13,8 @@ using Microsoft.AspNetCore.OData.Common;
 using Microsoft.AspNetCore.OData.Edm;
 using Microsoft.AspNetCore.OData.Formatter;
 using Microsoft.AspNetCore.Routing;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 using Microsoft.OData;
 using Microsoft.OData.Edm;
 using Microsoft.OData.ModelBuilder;
@@ -95,7 +97,16 @@ namespace Microsoft.AspNetCore.OData.Routing.Template
                         catch (ODataException ex)
                         {
                             string message = Error.Format(SRResources.InvalidParameterValueInUriFound, originalStrValue, edmParameter.Type.FullName());
-                            throw new ODataException(message, ex);
+                            ILoggerFactory loggerFactory = context.HttpContext?.RequestServices?.GetService<ILoggerFactory>();
+                            if (loggerFactory != null)
+                            {
+                                loggerFactory.CreateLogger("ODataFunctionParameterMatcher").LogError(message, ex);
+                                return null;
+                            }
+                            else
+                            {
+                                throw new ODataException(message, ex);
+                            }
                         }
 
                         // for without FromODataUri, so update it, for example, remove the single quote for string value.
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.OData/Routing/Template/SegmentTemplateHelpers.cs
```diff
@@ -93,7 +95,16 @@ internal static class SegmentTemplateHelpers
                         catch (ODataException ex)
                         {
                             string message = Error.Format(SRResources.InvalidParameterValueInUriFound, originalStrValue, edmParameter.Type.FullName());
-                            throw new ODataException(message, ex);
+                            ILoggerFactory loggerFactory = context.HttpContext?.RequestServices?.GetService<ILoggerFactory>();
+                            if (loggerFactory != null)
+                            {
+                                loggerFactory.CreateLogger("ODataFunctionParameterMatcher").LogError(message, ex);
```
gathogojr, line 101: Curious is there's a way we can verify that a logger is created successfully and the error message logged...

xuzhg: I tried to add the test, but, it seems we have to set up more things. I will try more later.

PR #1161, opened 2024-01-23
Optimize `ODataPathExtensions.GetNavigationSource`

### LINKED ISSUE #1118: Perf: Optimize `GetNavigationSource(ODataPath)`
The `GetNavigationSource(ODataPath)` is relatively expensive. Part of the cost is due to the fact that we rely on a custom path segment handler to determine the path.

Once we ship an ODL version that support indexing ODataPath, we can rewrite this method using a simple loop that reads the segments from the end until it find the appropriate one to determine the navigation source.

Related:
- https://github.com/OData/odata.net/issues/2814
- https://github.com/OData/odata.net/issues/2804
- https://github.com/OData/odata.net/pull/2811

habbes:
Fix #1118 

Replaces the use of `ODataPathNavigationSourceHandler` in `ODataPath.GetNavigationSource` extension method with a simpler reverse loop of the path.

With `ODataPathNavigationSourceHandler`
- a new instance of the handler is allocated
- all the segments of the path are visited
- a `List<string>` is created and populated with the string representation of each path segment. This list is used to generate a string representation of the path via a `Path` property. But this property is not needed to determine the navigation source (it's also not used by existing code).

The new implementation scans the path segments from the end and stops as soon as it finds the navigation source because only the last navigation source in the path matters.

The `ODataPathNavigationSourceHandler` is public so I kept it to avoid breaking changes.

**Before**

image

**After**

image

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Routing/ODataPathExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/ODataPathExtensions.cs
@@ -77,7 +77,7 @@ namespace Microsoft.AspNetCore.OData.Routing
                 throw Error.ArgumentNull(nameof(path));
             }
 
-            for (int i = path.Count - 1; i > -1; --i)
+            for (int i = path.Count - 1; i >= 0; --i)
             {
                 ODataPathSegment segment = path[i];
                 if (segment is EntitySetSegment entitySetSegment)
@@ -126,7 +126,6 @@ namespace Microsoft.AspNetCore.OData.Routing
                 }
 
                 return null;
-
             }
 
             return null;
--- a/src/Microsoft.AspNetCore.OData/Routing/ODataPathExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/ODataPathExtensions.cs
@@ -77,9 +77,59 @@ namespace Microsoft.AspNetCore.OData.Routing
                 throw Error.ArgumentNull(nameof(path));
             }
 
-            ODataPathNavigationSourceHandler handler = new ODataPathNavigationSourceHandler();
-            path.WalkWith(handler);
-            return handler.NavigationSource;
+            for (int i = path.Count - 1; i > -1; --i)
+            {
+                ODataPathSegment segment = path[i];
+                if (segment is EntitySetSegment entitySetSegment)
+                {
+                    return entitySetSegment.EntitySet;
+                }
+
+                if (segment is KeySegment keySegment)
+                {
+                    return keySegment.NavigationSource;
+                }
+
+                if (segment is NavigationPropertyLinkSegment navigationPropertyLinkSegment)
+                {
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.OData/Routing/ODataPathExtensions.cs
```diff
@@ -77,9 +77,59 @@ public static IEdmNavigationSource GetNavigationSource(this ODataPath path)
...
+                }
+
+                if (segment is PropertySegment)
+                {
+                    continue;
+                }
+
+                return null;
+
+            }
```
gathogojr, line 130: nit:
```suggestion
                return null;
            }
```

src/Microsoft.AspNetCore.OData/Routing/ODataPathExtensions.cs
```diff
@@ -77,9 +77,59 @@ public static IEdmNavigationSource GetNavigationSource(this ODataPath path)
                 throw Error.ArgumentNull(nameof(path));
             }
 
-            ODataPathNavigationSourceHandler handler = new ODataPathNavigationSourceHandler();
-            path.WalkWith(handler);
-            return handler.NavigationSource;
+            for (int i = path.Count - 1; i > -1; --i)
```
gathogojr, line 80: Is `i > -1;` better or more performant than `i >= 0`?

gathogojr (approved):

src/Microsoft.AspNetCore.OData/Routing/ODataPathExtensions.cs
habbes: Good question. I'd assume it depends on the architecture. I may have naively assumed `>=` does "more" work than plan `>`, however, I think in this case architectures usually have special treatment for 0 and that may be more important here.

I have compared the[ assembly output](https://sharplab.io/#v2:EYLgtghglgdgNAFxBAzmOATEBqAPgAQCYBGAWACh8AGAAn2IDoAlAVxgSjAFMGBhAezAAHKABsuAJwDKkgG5QAxlxQBuCvgDMdQjV40A3hRrG6W/ABYaAWQAUsBDQAeASgNGTHqADMaNxzQA+AF4aKldDcg8okwARfj9nNUjomgBfd2N05OMM4006SytCO3YncNzPHz9AmgBaYnLslOM4hKTmrI9Ok1yAbSsuBAALfgwASWFRGwHh0YmhUQB5IQ5+GBQGADl+MZhRWFgAc2cAXVz8ixpW+zLciK6KVKA) of `>=0` and `> -1` using sharplab.io and it appears that both of them get translated to the same assembly code which uses a greater than 0 comparison, rather than -1.

```asm
cmp dword ptr [ebp-8], 0
setge cl
```

With that in mind, I'll change to `>= 0` since it reads more naturally and it appears to be the compiler's preference.

PR #1164, opened 2024-01-30
Fixes #1147: Enable skiptoken handler to handler advanced orderby clause

### LINKED ISSUE #1147: @odata.nextLink doesn't work when query contains $orderby=parentEntity/childEntity
OData 8.2.3

My Odata query:
`https://localhost:44325/odata/Mas?$expand=product&$select=product&$orderby=product/name`

It returns me such responce:
image

with current @odata.nextLink:
`"@odata.nextLink": "https://localhost:44325/odata/Mas?$expand=product&$select=product&$orderby=product%2Fname&$skiptoken=name-null,source_id-%2711514%27"`

When I follow this link I get the following error:
image

For some reason, the functionality is trying to find the "name" field in the parentEntity.
If it's not a bug, then tell me how to make my query work.
Thank you for your advice.

[Project to reproduce](https://github.com/AndriiLesiuk/TestODataProject)

xuzhg:
Fixes #1147

### 1. How to generate the skip token when we have the advanced $orderby?
    
    Based on the discussion and back-compatibility, the PR follows up:
    a) If it's top level property orderby (declared or dynamic or computed), for example: $orderby=name
        the skiptoken will follow up the existing pattern:   `name-'abc'` 

    b) If it's other orderby, for example: $orderby=tolower(substring(name,1,2))
        the skiptoken will only contains the value:   `tx`

### 2. How to consume the skip token when apply the $skiptoken?
   Basically, the skiptoken is coverted to a 'Where(...)' clause (Linq Where function call). Behind the scene, it's a $filter clause.
   The existing implementation cannot handle all and it contains errors.
   This PR is to change it and based on the above truth, we should use our `IFilterBinder` to handle the $skiptoken.

    The logic is that:
    a) Convert the skiptoken and orderby to filter expression
    b) Call IFilterBinder to bind the filter expression to 'Where' Linq expression
    c) Apply the 'Where' to IQueryable.


src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
```diff
@@ -107,52 +97,44 @@ public override Uri GenerateNextPageLink(Uri baseUri, int pageSize, Object insta
...
-        internal static string GenerateSkipTokenValue(Object lastMember, IEdmModel model, IList<OrderByNode> orderByNodes, TimeZoneInfo timeZoneInfo = null)
+        internal static string GenerateSkipTokenValue(object lastMember, IEdmModel model, OrderByClause clause, ODataSerializerContext context = null)
         {
             if (lastMember == null)
             {
                 return string.Empty;
             }
 
-            IEnumerable<IEdmProperty> propertiesForSkipToken = GetPropertiesForSkipToken(lastMember, model, orderByNodes);
-            StringBuilder skipTokenBuilder = new StringBuilder(String.Empty);
```
xuzhg, line 121: Here's the part 1 refactor.

src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
```diff
@@ -228,234 +300,154 @@ internal static string GenerateSkipTokenValue(Object lastMember, IEdmModel model
...
-        /// <param name="orderByNodes">OrderBy information required to correctly apply the query option for default implementation.</param>
-        /// <param name="context">The <see cref="ODataQueryContext"/> which contains the <see cref="IEdmModel"/> and some type information</param>
-        /// <param name="skipTokenRawValue">The raw string value of the skiptoken query parameter.</param>
-        /// <returns></returns>
-        [SuppressMessage("Microsoft.Maintainability", "CA1506:AvoidExcessiveClassCoupling", Justification = "Class coupling acceptable.")]
-        private static IQueryable ApplyToCore(IQueryable query, ODataQuerySettings querySettings, IList<OrderByNode> orderByNodes, ODataQueryContext context, string skipTokenRawValue)
+        private static IQueryable ApplyToCore(IQueryable query, ODataQuerySettings querySettings,
+             ODataQueryOptions queryOptions, string skipTokenRawValue)
         {
-            Contract.Assert(query != null);
```
xuzhg, line 257: Here's the part 2 refactor

habbes (commented):
Partial review. Will share more feedback later.

src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
```diff
@@ -77,25 +67,25 @@ public override Uri GenerateNextPageLink(Uri baseUri, int pageSize, Object insta
 
                     if (expandedItem.OrderByOption != null)
                     {
-                        orderByNodes = OrderByNode.CreateCollection(expandedItem.OrderByOption);
+                        orderByClause = expandedItem.OrderByOption;
                     }
 
                     skipTokenGenerator = (obj) =>
```
habbes, line 73: Where is this used?

habbes: We should setup analyzers in our codebase to provide hints for such cases.

src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
```diff
@@ -464,28 +456,35 @@ internal static IEnumerable<IEdmProperty> GetPropertiesForSkipToken(object lastM
                 return null;
             }
 
-            IEnumerable<IEdmProperty> key = entity.Key();
-            if (orderByNodes != null)
+            IList<OrderByClause> orderByClauses = new List<OrderByClause>();
```
habbes, line 459:
```suggestion
            List<OrderByClause> orderByClauses = new List<OrderByClause>();
```

src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
```diff
@@ -464,28 +456,35 @@ internal static IEnumerable<IEdmProperty> GetPropertiesForSkipToken(object lastM
...
+            ISet<IEdmProperty> properties = new HashSet<IEdmProperty>();
```
habbes, line 460:
```suggestion
            HashSet<IEdmProperty> properties = new HashSet<IEdmProperty>();
```

test/Microsoft.AspNetCore.OData.E2E.Tests/SkipToken/SkipTokenControllers.cs
```diff
@@ -0,0 +1,199 @@
...
+{
+    public class SkipTokenControllers : ODataController
+    {
+        private static IList<StCustomer> _stCustomers;
+        private static IList<StOrder> _stOrders;
+
+        #region Data
+        static SkipTokenControllers()
+        {
+            IList<StAddress> stAddresses = new List<StAddress>()
```
habbes, line 24: Do you have test cases with dates and negative numbers?

src/Microsoft.AspNetCore.OData/Edm/OrderByClauseHelpers.cs
```diff
@@ -0,0 +1,102 @@
...
+    {
+        public static readonly string OrderByGlobalNameKey = "__orderby_EDC7A1AC-97F7-463F-BFF9-DE1FD7FCE27E";
+        public static readonly string OrderByPropertyNamePrefix = "__orderby_";
+
+        /// <summary>
+        /// Convert the OrderByClause to list.
+        /// </summary>
+        /// <param name="clause">The input orderby clause, the 'ThenBy' in each node does matter.</param>
+        /// <returns>The output orderby clauses, the 'ThenBy' in each node does NOT matter.</returns>
+        public static IList<OrderByClause> ToList(this OrderByClause clause)
```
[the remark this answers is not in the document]
xuzhg: Keep eyes on List<T> is more efficent than IList<T>. Since it's internal, I am ok to replace it using List<T>.

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -406,10 +411,10 @@ internal Expression ProjectElement(QueryBinderContext context, Expression source
 
                 bool isSelectingOpenTypeSegments = isContainDynamicPropertySelection || IsSelectAllOnOpenType(selectExpandClause, structuredType);
 
-                if (propertiesToExpand != null || propertiesToInclude != null || computedProperties != null || autoSelectedProperties != null || isSelectingOpenTypeSegments)
+                if (propertiesToExpand != null || propertiesToInclude != null || computedProperties != null || autoSelectedProperties != null || isSelectingOpenTypeSegments || context.OrderByClauses != null)
                 {
                     Expression propertyContainerCreation =
-                        BuildPropertyContainer(context, source, structuredType, propertiesToExpand, propertiesToInclude, computedProperties, autoSelectedProperties, isSelectingOpenTypeSegments);
+                        BuildPropertyContainer(context, source, structuredType, propertiesToExpand, propertiesToInclude, computedProperties, autoSelectedProperties, isSelectingOpenTypeSegments, isSelectedAll);
```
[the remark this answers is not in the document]
xuzhg: Line 377 tells us if 'isSelectedAll == true', we use the 'instance' to save the property mapping. So, when doing binding, we don't need this flag. But, when doing extra "orderby" property binding, we need this flag to tell us whether we should bind property or not.

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -1067,6 +1075,95 @@ public virtual Expression CreateTotalCountExpression(QueryBinderContext context,
...
+        protected virtual void BindOrderByProperties(QueryBinderContext context, Expression source, IEdmStructuredType structuredType,
+            IList<NamedPropertyExpression> includedProperties, bool isSelectedAll)
+        {
+            if (context == null || context.OrderByClauses == null || source == null || structuredType == null || includedProperties == null)
+            {
+                return;
+            }
+
+            // Avoid duplicated binding.
+            ISet<string> usedPropertyNames = new HashSet<string>();
```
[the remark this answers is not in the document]
xuzhg: Maybe you can give us a 'session' about this.

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -1067,6 +1075,95 @@ public virtual Expression CreateTotalCountExpression(QueryBinderContext context,
...
+                    // $orderby=computedProp&$select=computedProp
+                    if ((edmProperty != null && isSelectedAll) || usedPropertyNames.Contains(name))
+                    {
+                        continue;
+                    }
+                }
+
+                name = GetOrderByName(usedPropertyNames, ref start);
+                names.Add(name);
+                usedPropertyNames.Add(name);
```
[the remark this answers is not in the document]
xuzhg: Good Suggestion. Let me comment it out and leave a comment for it.

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -1067,6 +1075,95 @@ public virtual Expression CreateTotalCountExpression(QueryBinderContext context,
...
+                {
+                    AutoSelected = true
+                });
+            }
+
+            if (names.Count > 0)
+            {
+                // We use this to keep the order and naming mapping.
+                propertyName = Expression.Constant(OrderByClauseHelpers.OrderByGlobalNameKey);
+                propertyValue = Expression.Constant(string.Join(",", names));
```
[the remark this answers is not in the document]
xuzhg: That's good suggest. But, I'd say:
"We generate the 'LINQ Expression' as simple as possible so that it can be translated to run on most of LINQ providers." So, Keep it unchanged.

src/Microsoft.AspNetCore.OData/Query/Expressions/SelectExpandBinder.cs
```diff
@@ -1067,6 +1075,95 @@ public virtual Expression CreateTotalCountExpression(QueryBinderContext context,
...
+        private static string GetOrderByName(ISet<string> usedPropertyNames, ref int start)
+        {
+            do
+            {
+                // For advanced orderby, for example: $orderby=tolower(substring(location/street,1,2))
+                // we want to create a 'propertyName: propertyValue' binding.
+                // Since we cannot use the raw orderby literal to naming the property name, since it could be complicated and invalid
+                // We create a random name for each orderby clause.
+                string name = $"{OrderByClauseHelpers.OrderByPropertyNamePrefix}{start++}";
+                if (!usedPropertyNames.Contains(name))
```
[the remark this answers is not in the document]
xuzhg: Customer can create a property as "__ordreby__{index}".  It's uncommon but it's valid. Use this code to avoid any potential risk.

src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
```diff
@@ -679,6 +680,9 @@ private static IEnumerable<IEdmStructuralProperty> GetAvailableOrderByProperties
                     : entityType
                         .StructuralProperties()
                         .Where(property => property.Type.IsPrimitive() && !property.Type.IsStream())
+
+                        // @discuss: Orderby the keys using the key name alphabet order doesn't make sense for me.
+                        // Since we don't have the 'Order' value, we should keep the order same as definition order in the schema?
```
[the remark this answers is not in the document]
xuzhg: I'd like to here input from @mikepizzo also.
It's just for discussion. I will leave this code unchanged since change it is a behavior breaking changes.

src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
xuzhg: Thanks for suggestion. I'd like to keep it out of this PR scope. I'd like to leave it if we need to refactor for perf.

src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
```diff
@@ -165,17 +147,107 @@ internal static string GenerateSkipTokenValue(Object lastMember, IEdmModel model
...
+                if (index != clauses.Count)
+                {
+                    skipTokenBuilder.Append(',');
+                }
             }
 
             return skipTokenBuilder.ToString();
         }
 
+        private static IList<KeyValuePair<string, object>> GetPropertyValues(object source, IEdmModel model, IList<OrderByClause> clauses,
```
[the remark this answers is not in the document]
xuzhg: No, It's for 'Part-1' to generate the Skip Token string.

src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
```diff
@@ -228,234 +300,154 @@ internal static string GenerateSkipTokenValue(Object lastMember, IEdmModel model
...
-                directionMap = new Dictionary<string, OrderByDirection>();
+                return query;
             }
 
-            IDictionary<string, (object PropertyValue, Type PropertyType)> propertyValuePairs = PopulatePropertyValuePairs(skipTokenRawValue, context);
+            // It's better to visit the nodes of OrderByClause reclusively to get the orderby raw value.
+            // Since we don't have such methods, let's simply split the request raw value.
+            IList<string> orderBys = orderByOption.RawValue.Split(',');
+            IList<OrderByClause> orderByClauses = orderByOption.OrderByClause.ToList();
+            IList<string> tokenValueParis = PopulatePropertyValuePairs(skipTokenRawValue);
```
habbes, line 319:
```suggestion
            IList<string> tokenValuePairs = PopulatePropertyValuePairs(skipTokenRawValue);
```

src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
```diff
@@ -228,234 +300,154 @@ internal static string GenerateSkipTokenValue(Object lastMember, IEdmModel model
...
                         // We are aiming for the following expression
                         // when value is NOT null in the ascending order scenario:
                         // (Prop1 > Value1) OR ((Prop1 == Value1) AND (Prop2 > Value2)) ...
-                        compare = ExpressionBinderHelper.CreateBinaryExpression(
-                            binaryOperator: BinaryOperatorKind.GreaterThan,
-                            left: property,
-                            right: constant,
-                            liftToNull: !propertyIsNullable,
-                            querySettings: querySettings);
+                        compare = $"({orderByRaw} gt {orderByValue})";
```
habbes, line 367: If the user specified `asc` in the `$orderby` clause, (e.g. `$orderby=Category asc`) this will result in an error because `orderByRaw` will be `Category asc`, which leads to an invalid where clause being generated:

image

When `desc` is specified, you remove the `desc` keyword using the `RemoveDesc()` helper method. You should do the same thing for `asc`. And probably also add a test case that specifies `asc` explicitly.

habbes (changes requested):

src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
habbes: Actually, since this list only contains the values (without keys), why not just `tokenValues`?

```suggestion
            List<string> tokenValues = PopulatePropertyValuePairs(skipTokenRawValue);
```

src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
```diff
@@ -228,234 +300,154 @@ internal static string GenerateSkipTokenValue(Object lastMember, IEdmModel model
...
-             * OR (Prop1=Value1 AND Prop2=Value2 AND Prop3>Value3)
-             * and so on...
-             * Adding the first true to simplify implementation.
-             */
-            Expression lastEquality = null;
-            bool firstProperty = true;
-
-            foreach (KeyValuePair<string, (object PropertyValue, Type PropertyType)> item in propertyValuePairs)
+            string where = string.Empty;
+            string lastEquality = null;
```
habbes, line 327: nit: The name `lastEquality` might give the impression that this refers to the last or final equality in the loop, but actually it refers to the equality generated in the previous iteration.
```suggestion
            string previousEquality = null;
```

src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
```diff
@@ -228,234 +300,154 @@ internal static string GenerateSkipTokenValue(Object lastMember, IEdmModel model
...
-            else
-            {
-                directionMap = new Dictionary<string, OrderByDirection>();
+                return query;
             }
 
-            IDictionary<string, (object PropertyValue, Type PropertyType)> propertyValuePairs = PopulatePropertyValuePairs(skipTokenRawValue, context);
+            // It's better to visit the nodes of OrderByClause reclusively to get the orderby raw value.
+            // Since we don't have such methods, let's simply split the request raw value.
+            IList<string> orderBys = orderByOption.RawValue.Split(',');
```
habbes, line 317: This will fail if for non-trivial orderby expressions. For example, if you have a function call with multiple parameters like `$orderby=substring(Field, 1),Id`, this will splitinto: `["substring(Field", "1)", "Id"]` which is invalid and will lead to an argument out of range exception because `orderBys` list will have more items than `orderByClauses` list.

Did you intend to support these scenarios in this PR or in a future one? If you want to add support for that in the future then we should detect such cases and throw a meaningful exception that lets the user know that operation is not supported, and then we should create an issue to support that. Otherwise we should handle it better in this PR. Do we have an method that can generate a string representation of an `orderByClause.Expression`  that is a valid OData query expression?

image

src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
```diff
@@ -228,234 +300,154 @@ internal static string GenerateSkipTokenValue(Object lastMember, IEdmModel model
...
-             * Adding the first true to simplify implementation.
-             */
-            Expression lastEquality = null;
-            bool firstProperty = true;
-
-            foreach (KeyValuePair<string, (object PropertyValue, Type PropertyType)> item in propertyValuePairs)
+            string where = string.Empty;
+            string lastEquality = null;
+            bool isFirst = true;
+            Contract.Assert(orderBys.Count == orderByClauses.Count);
```
habbes, line 329: Since we knoe of valid user requests that will result in `orderBys.Count` being different from `orderByClauses.Count`, I don't think `Conract.Assert` is sufficient. I think we should throw an exception that tells that user that this particular orderby expression is not supported. Otherwise, it will throw a misleading argument of range exception like the following:

image

But ideally we should fix the issue such that this doesn't happen because it's not desirable for the server to generate a skiptoken value that it cannot process.

xuzhg: Since it's tested above and throws exception if it's not matched. It's safe to remove this assert.

src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
xuzhg: Changed the codes and add the test case.

src/Microsoft.AspNetCore.OData/Query/Query/DefaultSkipTokenHandler.cs
xuzhg: Fixed. Can you help to test it again?

habbes (approved):

PR #1183, opened 2024-03-02
Set capacity for TruncatedCollection(IQueryable, int, boolean) overload

### LINKED ISSUE #1185: Truncated Collection causes list resizing when endpoint has a huge page size
**Assemblies affected**
AspNetCore.OData 8.2.4
**Describe the bug**
TruncatedCollection inherits from list. IQueryable constructor of Truncated Collection does not set the list capacity and this causes list resizing

**Reproduce steps**
Updated sample benchmark server see my pr
https://github.com/OData/AspNetCoreOData/pull/1183

**Data Model**
n/a

**EDM (CSDL) Model**
n/a

**Request/Response**
n/a

**Expected behavior**
Capacity should be set to avoid list re-sizing

**Screenshots**
na/

**Additional context**
n/a

senioroman4uk:
After profiling our service that uses AspNetCore.OData with a huge page size of 3000. I've noticed a lot of time is spent inside TruncatedCollection overload for IQueryable
image

Since it uses List as a base class current initialization will resize quite a few times before reaching the page size. This PR attempts to fix this for some of the overloads were size of the page is known

fixes #1185

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
@@ -29,8 +29,10 @@ namespace Microsoft.AspNetCore.OData.Query.Container
         /// <param name="source">The collection to be truncated.</param>
         /// <param name="pageSize">The page size.</param>
         public TruncatedCollection(IEnumerable<T> source, int pageSize)
-            : base(source.Take(checked(pageSize + 1)))
+            : base(checked(pageSize + 1))
         {
+            var items = source.Take(checked(pageSize + 1));
+            AddRange(items);
             Initialize(pageSize);
         }
 
@@ -54,8 +56,10 @@ namespace Microsoft.AspNetCore.OData.Query.Container
         // NOTE: The queryable version calls Queryable.Take which actually gets translated to the backend query where as 
         // the enumerable version just enumerates and is inefficient.
         public TruncatedCollection(IQueryable<T> source, int pageSize, bool parameterize)
-            : base(Take(source, pageSize, parameterize))
+            : base(checked(pageSize + 1))
         {
+            var items = Take(source, pageSize, parameterize);
+            AddRange(items);
             Initialize(pageSize);
         }
 
--- a/src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
@@ -85,6 +85,7 @@ namespace Microsoft.AspNetCore.OData.Query.Container
             {
                 AddRange(source);
             }
+
             if (pageSize > 0)
             {
                 Initialize(pageSize);
--- a/src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
@@ -17,6 +17,9 @@ namespace Microsoft.AspNetCore.OData.Query.Container
     /// <typeparam name="T">The collection element type.</typeparam>
     public class TruncatedCollection<T> : List<T>, ITruncatedCollection, IEnumerable<T>, ICountOptionCollection
     {
+        // The default capacity of the list.
+        // https://github.com/microsoft/referencesource/blob/master/mscorlib/system/collections/generic/list.cs#L38
+        private const int defaultCapacity = 4;
         private const int MinPageSize = 1;
 
         private bool _isTruncated;
@@ -31,7 +34,7 @@ namespace Microsoft.AspNetCore.OData.Query.Container
         public TruncatedCollection(IEnumerable<T> source, int pageSize)
             : base(checked(pageSize + 1))
         {
-            var items = source.Take(checked(pageSize + 1));
+            var items = source.Take(Capacity);
             AddRange(items);
             Initialize(pageSize);
         }
@@ -70,8 +73,18 @@ namespace Microsoft.AspNetCore.OData.Query.Container
         /// <param name="pageSize">The page size.</param>
         /// <param name="totalCount">The total count.</param>
         public TruncatedCollection(IEnumerable<T> source, int pageSize, long? totalCount)
-            : base(pageSize > 0 ? source.Take(checked(pageSize + 1)) : source)
+            : base(pageSize > 0
+                ? checked(pageSize + 1)
+                : (totalCount > 0 ? (totalCount < int.MaxValue ? (int)totalCount : int.MaxValue) : defaultCapacity))
         {
+            if (pageSize > 0)
+            {
+                AddRange(source.Take(Capacity));
+            }
+            else
+            {
+                AddRange(source);
+            }
             if (pageSize > 0)
             {
                 Initialize(pageSize);
--- a/src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
@@ -19,7 +19,7 @@ namespace Microsoft.AspNetCore.OData.Query.Container
     {
         // The default capacity of the list.
         // https://github.com/microsoft/referencesource/blob/master/mscorlib/system/collections/generic/list.cs#L38
-        private const int defaultCapacity = 4;
+        private const int DefaultCapacity = 4;
         private const int MinPageSize = 1;
 
         private bool _isTruncated;
@@ -75,7 +75,7 @@ namespace Microsoft.AspNetCore.OData.Query.Container
         public TruncatedCollection(IEnumerable<T> source, int pageSize, long? totalCount)
             : base(pageSize > 0
                 ? checked(pageSize + 1)
-                : (totalCount > 0 ? (totalCount < int.MaxValue ? (int)totalCount : int.MaxValue) : defaultCapacity))
+                : (totalCount > 0 ? (totalCount < int.MaxValue ? (int)totalCount : int.MaxValue) : DefaultCapacity))
         {
             if (pageSize > 0)
             {
--- a/src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
@@ -101,7 +101,7 @@ namespace Microsoft.AspNetCore.OData.Query.Container
         /// <param name="totalCount">The total count.</param>
         // NOTE: The queryable version calls Queryable.Take which actually gets translated to the backend query where as 
         // the enumerable version just enumerates and is inefficient.
-        [Obsolete("should not used should be removed in the next major version")]
+        [Obsolete("should not be used, will be marked internal in the next major version")]
         public TruncatedCollection(IQueryable<T> source, int pageSize, long? totalCount) : this(source, pageSize,
             totalCount, false)
         {
@@ -116,7 +116,7 @@ namespace Microsoft.AspNetCore.OData.Query.Container
         /// <param name="parameterize">Flag indicating whether constants should be parameterized</param>
         // NOTE: The queryable version calls Queryable.Take which actually gets translated to the backend query where as 
         // the enumerable version just enumerates and is inefficient.
-        [Obsolete("should not be used should be removed in the next major version")]
+        [Obsolete("should not be used, will be marked internal in the next major version")]
         public TruncatedCollection(IQueryable<T> source, int pageSize, long? totalCount, bool parameterize)
             : base(pageSize > 0 ? Take(source, pageSize, parameterize) : source)
         {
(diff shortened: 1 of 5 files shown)


senioroman4uk:
@microsoft-github-policy-service agree company="Microsoft"

senioroman4uk:
here are the results of the test I did with a sample server and bombardier

Bombarding https://localhost:7034/odata/Products for 1m0s using 10 connection(s)
Before
image

After
image

Note that for smaller lists (less than 16 pre-allocating will be slower or showing no effect)

senioroman4uk:
fixes #1185

src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
```diff
@@ -29,8 +29,10 @@ public class TruncatedCollection<T> : List<T>, ITruncatedCollection, IEnumerable
         /// <param name="source">The collection to be truncated.</param>
         /// <param name="pageSize">The page size.</param>
         public TruncatedCollection(IEnumerable<T> source, int pageSize)
-            : base(source.Take(checked(pageSize + 1)))
+            : base(checked(pageSize + 1))
         {
+            var items = source.Take(checked(pageSize + 1));
```
julealgon, line 34: Would it make sense to do something like this to avoid the same calculation again (and opening it up for potential inconsistencies in the future)?
```suggestion
            var items = source.Take(Capacity);
```

src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
julealgon, line 1: What about other overloads? I'm pretty sure there is a couple more that would also need the same optimization.

senioroman4uk: there are some overloads that actually only used in tests. I would like to propose marking them as obsolete, since the class itself is public and leaving them as is

julealgon: Ah I see.

It is usually considered a bad practice having code that is only ever used from tests. I'd advocate for removing them as well if that's the case.

senioroman4uk: Removing them will be breaking change, since class itself is public. Should we just remove them? That would break some users who might be using those overloads even though they are not used by ODataAspNetCore

julealgon: Oh if they are `public` the best approach is definitely marking them as `[Obsolete]` first as you suggested earlier and removing them in the release after the next or the next major version change.

In other words, I was basically agreeing with you 😛

src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
senioroman4uk: You are right, applied

src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
senioroman4uk: Marked two other overloads as obsolete. Modified the remaining overload
public TruncatedCollection(IEnumerable<T> source, int pageSize, long? totalCount)

it is actually being used in one place ( CollectionExpandedProperty<T>.GetValue)

senioroman4uk: please take a look

src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
```diff
@@ -54,8 +59,10 @@ public TruncatedCollection(IQueryable<T> source, int pageSize) : this(source, pa
         // NOTE: The queryable version calls Queryable.Take which actually gets translated to the backend query where as 
         // the enumerable version just enumerates and is inefficient.
         public TruncatedCollection(IQueryable<T> source, int pageSize, bool parameterize)
-            : base(Take(source, pageSize, parameterize))
+            : base(checked(pageSize + 1))
```
habbes, line 62: Why `pageSize + 1`?

src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
```diff
@@ -84,7 +101,9 @@ public TruncatedCollection(IEnumerable<T> source, int pageSize, long? totalCount
         /// <param name="totalCount">The total count.</param>
         // NOTE: The queryable version calls Queryable.Take which actually gets translated to the backend query where as 
         // the enumerable version just enumerates and is inefficient.
-        public TruncatedCollection(IQueryable<T> source, int pageSize, long? totalCount) : this(source, pageSize, totalCount, false)
+        [Obsolete("should not used should be removed in the next major version")]
```
habbes, line 104: Why is this obsolete? Does it have no practical use case?

src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
```diff
@@ -97,6 +116,7 @@ public TruncatedCollection(IQueryable<T> source, int pageSize, long? totalCount)
         /// <param name="parameterize">Flag indicating whether constants should be parameterized</param>
         // NOTE: The queryable version calls Queryable.Take which actually gets translated to the backend query where as 
         // the enumerable version just enumerates and is inefficient.
+        [Obsolete("should not be used should be removed in the next major version")]
         public TruncatedCollection(IQueryable<T> source, int pageSize, long? totalCount, bool parameterize)
```
habbes, line 120: @xuzhg do you know why we have this overload? Why was it made public in the first case? Does it have practical use cases? Is it okay to obsolete?

src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
habbes: @xuzhg do you know why we have this overload? Why was it made public in the first case? Does it have practical use cases? Is it okay to obsolete?

src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
senioroman4uk: from implementation of Take. It will build an expression to get pagesize + 1 elements, in order to set truncated flag

  private static IQueryable<T> Take(IQueryable<T> source, int pageSize, bool parameterize)
        {
            if (source == null)
            {
                throw Error.ArgumentNull("source");
            }

            return ExpressionHelpers.Take(source, checked(pageSize + 1), typeof(T), parameterize) as IQueryable<T>;
        }

src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
senioroman4uk: based on my search in the code base those two overloads were only used in unit tests. I've tested by removing them and then re-building AspNetCoreOData and AspNetCoreOData.Newtonsoft solutions

gathogojr:
Thank you @senioroman4uk for your pull request contribution. It looks like a great optimization. However, I don't know if applying the `Obsolete` attribute to some of the constructors without a good understanding of the use cases they were intended for was a good idea since that action introduced extra things/changes to debate about. I hope it won't cause unnecessary delays.

src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
habbes: I'm not sure why this class was made public in the first place instead of internal. But since it's public, I would be reluctant to mark it as obsolete without being certain that it has no practical use for 3rd party code that consumes the library.

src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
xuzhg: I am ok to obsolete them, but maybe update the message that it will be marked as internal later?

src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
xuzhg: I am ok to obsolete them, but maybe update the message that it will be marked as internal later?

xuzhg (dismissed):

habbes (approved):

habbes:
@senioroman4uk please fix the PublicApiOut tests:

image

src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
senioroman4uk: Updated deprecation message to be "should not be used, will be marked internal in the next major version"

senioroman4uk:
Updated public api files

habbes (dismissed):

src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
```diff
@@ -17,6 +17,9 @@ namespace Microsoft.AspNetCore.OData.Query.Container
     /// <typeparam name="T">The collection element type.</typeparam>
     public class TruncatedCollection<T> : List<T>, ITruncatedCollection, IEnumerable<T>, ICountOptionCollection
     {
+        // The default capacity of the list.
+        // https://github.com/microsoft/referencesource/blob/master/mscorlib/system/collections/generic/list.cs#L38
```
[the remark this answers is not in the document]
senioroman4uk: updated

habbes (dismissed):

src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
```diff
@@ -66,8 +73,18 @@ public TruncatedCollection(IQueryable<T> source, int pageSize, bool parameterize
...
         /// <param name="totalCount">The total count.</param>
         public TruncatedCollection(IEnumerable<T> source, int pageSize, long? totalCount)
-            : base(pageSize > 0 ? source.Take(checked(pageSize + 1)) : source)
+            : base(pageSize > 0
+                ? checked(pageSize + 1)
+                : (totalCount > 0 ? (totalCount < int.MaxValue ? (int)totalCount : int.MaxValue) : DefaultCapacity))
         {
+            if (pageSize > 0)
+            {
+                AddRange(source.Take(Capacity));
```
gathogojr, line 82: @senioroman4uk I think there could be a bug here...
If say `pageSize` was 5, by calling `base(checked(pageSize + 1))` we initialize `Capacity` to 6.
If we do `AddRange(source.Take(Capacity))`, that will be equivalent to `source.Take(6)`...

gathogojr: We should have a test for this where we verify the items that are returned in each scenario that the logic in this constructor anticipates

senioroman4uk: the previous version of the code was taking page size + 1 already. +1 element will be used to set the truncated flag in initialize and will be removed from collection

senioroman4uk: I will look into the existing tests in the morning

gathogojr: If this is a bug, I think it might have existed even before based on the previous logic:
```csharp
base(pageSize > 0 ? source.Take(checked(pageSize + 1)) : source)
```
A test should help us verify that...

src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
```diff
@@ -66,8 +73,18 @@ public TruncatedCollection(IQueryable<T> source, int pageSize, bool parameterize
...
+            }
+            else
+            {
+                AddRange(source);
+            }
```
gathogojr, line 87:
```suggestion
            }

```

src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
gathogojr: > +1 element will be used to set the truncated flag in initialize and will be removed from collection

If the extra item is subsequently removed from the collection then there's no cause for alarm

senioroman4uk: From implementation of Initialize
```c#
 private void Initialize(int pageSize)
        {
            if (pageSize < MinPageSize)
            {
                throw Error.ArgumentMustBeGreaterThanOrEqualTo("pageSize", pageSize, MinPageSize);
            }

            _pageSize = pageSize;

            if (Count > pageSize)
            {
                _isTruncated = true;
                RemoveAt(Count - 1);
            }
        }
```

This is covered in the unit test already

senioroman4uk:
@gathogojr applied your suggestions, please give it another review when you have a chance

gathogojr (approved):

habbes (approved):

PR #1213, opened 2024-04-16
Migrate pipelines to 1ES

habbes:
Migrate pipelines to use 1ES approved agents and stages. It adds some required compliance and security checks to our usual pipeline steps. I also made other adjustments to get the pipeline to work and port some of the tasks from classic to YAML pipeline, e.g. renaming artifacts, update the feed publishing configuration, etc.

For the nightly pipelines, the output artifacts are stored in `drop` (the NuGet packages) and `dropOutput` (the built DLLs)


habbes:
> Question: Would it make sense to start putting CI builds on NuGet just like every other build? That way, if people want to take on more frequent builds, they can do so without changing their configuration?

@robertmclaws by that do you mean pushing the nightly builds to NuGet instead of MyGet? @xuzhg  could you share your thoughts on this?

xuzhg:
> > Question: Would it make sense to start putting CI builds on NuGet just like every other build? That way, if people want to take on more frequent builds, they can do so without changing their configuration?
> 
> @robertmclaws by that do you mean pushing the nightly builds to NuGet instead of MyGet? @xuzhg could you share your thoughts on this?

Or @robertmclaws do you mean putting CI builds on Github? using Github Action?

robertmclaws:
@xuzhg @habbes GitHub Actions is terrible and I would not recomend it under any circumstances. There is no good way to separate out build and release functionality the right way... One .yaml file cannot access the artifacts of another.yaml file without extensive non-trivial work.

I mean getting rid of MyGet and just posting CI builds on NuGet with the standard CI-DateTime versioning.

Your process means that nothing gets into Main without a PR, and there is no need to generate "nightly" builds when there are no changes.

Have PR validation builds that don't trigger releases, then use Restier's build tricks to set the build version on the Pipeline and deploy the right versions from the right branches.

It will have the added benefit of allowing you to catch issues between CI builds but before RTM builds when things break.

habbes:
@xuzhg @robertmclaws We do occasionally use the nightly pipeline to create temporarily releases of a pull request that has not be merged to the main branch to allow customers to try out and validate certain features. Do you think it would be appropriate to have such releases on NuGet as well? I guess it doesn't matter if all the CI releases are there too.

pipelines/azure_pipelines.yml
```diff
@@ -6,84 +6,113 @@ resources:
...
+    pool:
+      name: MSSecurity-1ES-Build-Agents-Pool
+      image: MSSecurity-1ES-Windows-2022
+      os: windows
+    customBuildTags:
+    - ES365AIMigrationTooling
+    stages:
+    - stage: stage
+      jobs:
+      - job: Job_1
```
gathogojr, line 38: Job was named "Main" before. Any reason for the change?

pipelines/azure_pipelines.yml
```diff
@@ -6,84 +6,113 @@ resources:
...
+        displayName: Agent job 1
```
gathogojr, line 39: Display name for this job was "Main Build" before. Any reason for the change? Or is it to avoid conflict with the non-1ES pipeline?

pipelines/azure_pipelines_nightly.yml
```diff
@@ -1,3 +1,4 @@
+trigger: none
```
gathogojr, line 1: Why `trigger: none`?

pipelines/azure_pipelines_nightly.yml
```diff
@@ -7,160 +8,244 @@ resources:
...
-...
+      name: MSSecurity-1ES-Build-Agents-Pool
+      image: MSSecurity-1ES-Windows-2022
+      os: windows
+    customBuildTags:
+    - ES365AIMigrationTooling
+    stages:
+    - stage: stage
+      jobs:
+      - job: Job_1
```
gathogojr, line 38: See my comment here https://github.com/OData/AspNetCoreOData/pull/1213/files#r1590554501

pipelines/azure_pipelines_nightly.yml
```diff
@@ -7,160 +8,244 @@ resources:
...
+        displayName: Agent job 1
```
gathogojr, line 39: See my comments https://github.com/OData/AspNetCoreOData/pull/1213/files#r1590554869

pipelines/azure_pipelines.yml
habbes: No reason. The name was auto-generated by a tool and I forgot to rename it. Thanks.

pipelines/azure_pipelines_nightly.yml
habbes: Hmm. `trigger: none` means CI triggers are disabled. CI triggers are used to automatically run the pipeline when we push to a branch. For the nightly build, we usually run it only the cron schedule or when initiated manually.

habbes: See:
- https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/trigger?view=azure-pipelines#trigger-none
- https://learn.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#ci-triggers

pipelines/azure_pipelines.yml
```diff
@@ -6,84 +6,113 @@ resources:
...
+      name: MSSecurity-1ES-Build-Agents-Pool
+      image: MSSecurity-1ES-Windows-2022
+      os: windows
+    customBuildTags:
+    - ES365AIMigrationTooling
+    stages:
+    - stage: stage
+      jobs:
+      - job: Main
+        displayName: Main build
```
gathogojr, line 39:
```suggestion
        displayName: Main Build
```

gathogojr (approved):

ElizabethOkerio (approved):

PR #1232, opened 2024-05-09
Fixes #1212: Link generation with string key and back slash in the key value

### LINKED ISSUE #1212: Faulty link generation when creating entities with string key and slashes
**Assemblies affected**
Microsoft.AspNetCore.OData Version 7.7.4 and also 8.2.5

**Describe the bug**
A change made a while ago in the Web API Framework has led to an issue where the link generation for entities with string keys that contain slashes is faulty. The problematic change (affects version 7.7.4) can be found in this [pull request](https://github.com/OData/WebApi/pull/2332/files#diff-d94a23f5e2bc6c75b9350b2c42cf8ded0b3cb84812e2ed8cdbcaaf7244822680R77).

As a temporary workaround, we have replaced the `ODataRoute` class with a patched version of the class. In this patched class, the problematic code has been commented out (line 77-101 of the mentioned pull request).

For version 8.2.5, the `linkBuilder` used within the `RequestHelper.cs` class is responsible for generating the location header URL. This URL unfortunately includes an unencoded slash within the ID, leading to the creation of an `idLink` that contains this problematic slash.

**Reproduce steps**
- Create an entity with a string key that contains a slash
- Get request to the location header URL
- results in a 404 error because it interprets the unencoded slash as a path separator

I also created a small project that demonstrates the issue. Just follow the steps mentioned in the README.md -> How to replicate.
https://github.com/dankaiser1808/linkGen-bug-replication 

**Expected behavior**
 For the individual segments of a link or a URI, targeted escaping should be performed, especially in the case when a primary key or an entity ID contains a slash (or similarly problematic special characters).

Thank you for taking the time to review this issue. I look forward to your feedback and am ready to assist further if needed.

xuzhg:
fixes #1212 

When a key value contains back slah '/', the link generated should escape this character.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Common/StringExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Common/StringExtensions.cs
@@ -28,6 +28,21 @@ namespace Microsoft.AspNetCore.OData.Common
         }
 
         /// <summary>
+        /// Escape Uri string for '/'
+        /// </summary>
+        /// <param name="uriString">The Uri string.</param>
+        /// <returns>Escaped back slash Uri string.</returns>
+        public static string EscapeBackSlashUriString(this string uriString)
+        {
+            if (uriString == null)
+            {
+                return null;
+            }
+
+            return uriString.Replace("/", "%2F");
+        }
+
+        /// <summary>
         /// Normalize the http method.
         /// </summary>
         /// <param name="method">The http method.</param>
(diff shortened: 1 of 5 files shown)


gathogojr:
@xuzhg Curious, is the slash the only character that would give us problems if not escaped?

xuzhg:
> @xuzhg Curious, is the slash the only character that would give us problems if not escaped?

Currently, yes. ODL use "/" to split the OData path directly.

In my opinion, we should use 'lexer' to split the OData path, then the 'escaped' or 'unescaped' slash is not a problem again.

gathogojr (approved):

PR #1237, opened 2024-05-16
Trigger pipeline on PRs against main

habbes:
Once merged, PRs against `main` should automatically trigger the rolling build pipeline.

I've also renamed the stages from `stage` to `Stage_1`.

### DIFF
--- a/pipelines/azure_pipelines.yml
+++ b/pipelines/azure_pipelines.yml
@@ -37,7 +37,7 @@ extends:
     customBuildTags:
     - ES365AIMigrationTooling
     stages:
-    - stage: Main
+    - stage: build
       jobs:
       - job: Main
         displayName: Main Build
--- a/pipelines/azure_pipelines.yml
+++ b/pipelines/azure_pipelines.yml
@@ -37,7 +37,7 @@ extends:
     customBuildTags:
     - ES365AIMigrationTooling
     stages:
-    - stage: stage
+    - stage: Stage 1
       jobs:
       - job: Main
         displayName: Main Build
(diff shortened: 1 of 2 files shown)


ElizabethOkerio (approved):

pipelines/azure_pipelines.yml
```diff
@@ -33,7 +37,7 @@ extends:
     customBuildTags:
     - ES365AIMigrationTooling
     stages:
-    - stage: stage
+    - stage: Stage_1
```
gathogojr, line 40: Why don't we call this stage "build" like it appears in other internal pipelines:
image

pipelines/azure_pipelines.yml
```diff
@@ -2,6 +2,10 @@ trigger:
   branches:
     include:
     - main
+
+pr:
+  - main
```
gathogojr, line 7: Is this the change that will cause build to be triggered when a pull request against the `main` branch is created?

habbes: Yes

pipelines/azure_pipelines.yml
habbes: I am okay with any name suggestion. But note that in the previous version of the pipeline, we did not explicitly name the stage. `Main` was the name of the job. Not sure if that name was also automatically applied to the stage.

habbes: I've renamed it to `build`.

gathogojr (approved):

ElizabethOkerio (approved):

PR #1247, opened 2024-05-28
Update ODL version

ElizabethOkerio:
Updated the ODL version to `8.0.0-preview.3`. Since `8.0.0-preview.3` had some breaking changes, I've also made some updates like: 

1. Getting rid of the `IContainerBuilder` and using `Microsoft.Extensions.DependencyInjection` instead. 
2. Removed references to `ODataSimplifiedOptions` and used `ODataMessageReaderSettings` instead since we got rid of this class in `ODL 8.0.0-preview.3`
3. I've also made changes to ODataError's `ErrorCode` to `Code` and the way the ODataInnerError is initialized from `new ODataInnerError { Message = ConvertModelStateErrors(modelStateError) };` to `new Dictionary<string, ODataValue>
                            {
                                {
                                    "Message", new ODataPrimitiveValue(ConvertModelStateErrors(modelStateError))
                                }
                            });`


sample/BenchmarkServer/BenchmarkServer.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFramework>net7.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
```
xuzhg, line 4: If we update to ODL 8, we need to update our version to 9 for ASP.NET Core OData. right?

ElizabethOkerio: yes

habbes (changes requested):
We shouldn't target this PR against the main branch due to potential breaking/disruptive changes.

ElizabethOkerio:
> We shouldn't target this PR against the main branch due to potential breaking/disruptive changes.

Yes. Creating a branch for it.

habbes (commented):
We should set the version to 9.0.0-preview.1

PR #1259, opened 2024-06-10
Support Instance annotations using annotation container

xuzhg:
Support instance annotation

OData model builder has the interface 

image

If a customer can modify his model (C# class), he can use the model-based instance annotation container provided in OData model builder.

```C#

public class EntityTypeWithAnnotation
{
   public int Id { get; set; }
   public IODataInstanceAnnotationContainer InstanceAnnotations { get; set; }
 }
```

This PR is to:
1) Enable Instance annotation deserialization
    When a request payload contains instance annotation for resource and property, we can read it and convert to the key/value pairs and save them into instance annotation container.

2) Enable Instance annotation serialization
 When a serialize an object which contains instance annotation in the container, we can convert to ODataInstanceAnnotation and save them into InstanceAnnotation collection of ODataResourceBase or ODataPropertyInfo.

3) We support to read property without value but with instance annotation
 
4) We support to query the entity set which contains instance annotation. 
   for example:  ~/customers/1/name
   the payload could be:

```json
{
    "@context":"$metadata#Customers",
     "@my.instanceAnnotation":"Smithe" 
    "value": [
    {
        ....
    },
    ....
  ]
}    
```

5) We support to query the property directly which contains instance annotation. 
   for example:  ~/customers/1/name
   the payload could be:

```json
{
    "@context":"$metadata#Customers(1)/Name",
     "@my.instanceAnnotation":"Smithe" 
    "value":"Smith",
}    
```


src/Microsoft.AspNetCore.OData/Deltas/DeltaOfT.cs
```diff
@@ -150,6 +154,20 @@ public override bool TrySetPropertyValue(string name, object value)
                 throw Error.ArgumentNull(nameof(name));
             }
 
+            if (IsInstanceAnnotation(name, out PropertyInfo annotationContainerPropertyInfo))
```
habbes, line 157: I am concerned about the performance impact of this check. It would be good to run some profiling to verify whether this call adds significant overhead. Here's why I'm concerned:

- The method `IsInstanceAnnotation` is called each time we set a property value. Potentially this means we call it for each property at least once for each `PUT`/`PATCH` request that uses `Delta<T>`.
- The `IsInstanceAnnotation` method scans all the properties.  So if `TrySetPropertyValue` is called for each property, then we have an `O(n^2)` operation which could be costly if the entity has a lot of properties
- We call `IsInstanceAnnotation` even if the `_instanceAnnotationsContainer` has already been found. I think we only expect one instance annotations container property in the entity class. If there are multiple, I think we should only select the first one. So we could add a condition `if (_instanceAnnotationContainer == null && IsInstanceAnnotation(...))`.
- This approach will be most expensive when the class doesn't have an annotations container. Meaning the people who don't use the feature pay biggest price (should measure the cost to see whether this would actually be an issue in practice)

When adding new opt-in features, I think we should consider as much as possible ways to minimize or eliminate the cost for people who don't use the feature. Maybe we can consider computing only once whether this type has an instance annotations container and cache it somewhere where it's fast to lookup. Since `Delta<T>` is used during deserialization, is there a flag we can set during deserialization when we detect instance annotations? And use that flag to determine whether the contain exists? Not sure if that's a feasible approach.

src/Microsoft.AspNetCore.OData/Deltas/DeltaOfT.cs
```diff
@@ -685,6 +705,44 @@ private static bool IsIgnoredProperty(bool isTypeDataContract, PropertyInfo prop
...
+            {
+                return;
+            }
+
+            IODataInstanceAnnotationContainer desContainer =
+                _instanceAnnotationContainerPropertyInfo.GetValue(targetEntity) as IODataInstanceAnnotationContainer;
+            if (desContainer == null)
+            {
+                _instanceAnnotationContainerPropertyInfo.SetValue(targetEntity, sourceContainer);
+                return;
```
habbes, line 727: If the source container gets modified after this, (e.g. new items added to it), the changes will also affect the destination container since they'll be references to the same instance. Is this expected behaviour? Is it something we should worry about? Or is it a non-issue?

src/Microsoft.AspNetCore.OData/Edm/EdmModelExtensions.cs
```diff
@@ -102,6 +110,51 @@ public static IEdmStructuredTypeReference ResolveResourceType(this IEdmModel mod
...
+        /// followed by the name of the term, optionally followed by a hash (#) and a qualifier.</param>
+        /// <returns>The resolved term or null if not found.</returns>
+        public static IEdmTerm ResolveTerm(this IEdmModel model, string annotationIdentifier)
+        {
+            if (model == null)
+            {
+                throw Error.ArgumentNull(nameof(model));
+            }
+
+            string[] identifier = annotationIdentifier.Split('#');
```
habbes, line 127: Since we only use `identifier[0]`, maybe we should use `Substring` instead of `Split` so that we allocate at most one new string.

It could also be faster:
image

See benchmark:
https://sharpbench.dev/#s:gqF2AaFj2gJpLy8gdmlzaXQgaHR0cHM6Ly9iZW5jaG1hcmtkb3RuZXQub3JnLyBmb3IgbW9yZSBpbmZvIG9uIEJlbmNobWFya0RvdE5ldAoKCnVzaW5nIEJlbmNobWFya0RvdE5ldC5BdHRyaWJ1dGVzOwp1c2luZyBCZW5jaG1hcmtEb3ROZXQuRGlhZ25vc2VyczsKCltNZW1vcnlEaWFnbm9zZXJdCnB1YmxpYyBjbGFzcyBCZW5jaG1hcmtzCnsKICAgIHByaXZhdGUgc3RyaW5nIGFubm90YXRpb24gPSAic29tZS5uYW1lLnNwYWNlLmlkZW50aWZpZXIjcXVhbGlmaWVyIjsKCiAgICBbQmVuY2htYXJrXQogICAgcHVibGljIHN0cmluZyBHZXRJZGVudGlmaWVyV2l0aFNwbGl0KCkKICAgIHsKICAgICAgICB2YXIgaWRlbnRpZmllciA9IGFubm90YXRpb24uU3BsaXQoJyMnKTsKICAgICAgICByZXR1cm4gaWRlbnRpZmllclswXTsKICAgIH0KCiAgICBbQmVuY2htYXJrXQogICAgcHVibGljIHN0cmluZyBHZXRJZGVudGlmaWVyV2l0aFN1YnN0cmluZygpCiAgICB7CiAgICAgICAgdmFyIGluZGV4ID0gYW5ub3RhdGlvbi5JbmRleE9mKCcjJyk7CiAgICAgICAgdmFyIGlkZW50aWZpZXIgPSBhbm5vdGF0aW9uLlN1YnN0cmluZygwLCBpbmRleCk7CiAgICAgICAgcmV0dXJuIGlkZW50aWZpZXI7CiAgICB9Cn0KCgo=

src/Microsoft.AspNetCore.OData/Edm/EdmModelExtensions.cs
```diff
@@ -102,6 +110,51 @@ public static IEdmStructuredTypeReference ResolveResourceType(this IEdmModel mod
...
+
+            IEdmTerm term = model.FindTerm(identifier[0]);
```
habbes, line 129: I think it sucks that we have to perform a new string allocation just so we can lookup a term using a substring. It would be great if `model.FindTerm` would support `ReadOnlySpan<char>` so we can avoid string allocation in such cases. We have an open issue for that: https://github.com/OData/odata.net/issues/2801

src/Microsoft.AspNetCore.OData/Edm/EdmModelExtensions.cs
```diff
@@ -102,6 +110,51 @@ public static IEdmStructuredTypeReference ResolveResourceType(this IEdmModel mod
...
+
+            foreach (var refModels in model.ReferencedModels)
+            {
+                var refedTerms = refModels.SchemaElements.OfType<IEdmTerm>()
+                    .Where(e => string.Equals(termName, e.FullName(), StringComparison.OrdinalIgnoreCase));
+
+                terms = terms.Concat(refedTerms);
+            }
+
+            if (terms.Count() > 1)
```
habbes, line 150: `terms.Count()` will scan the whole collection. If we want to find duplicates, maybe we should use `terms.Take(2).Count()`.

src/Microsoft.AspNetCore.OData/Edm/EdmModelExtensions.cs
```diff
@@ -102,6 +110,51 @@ public static IEdmStructuredTypeReference ResolveResourceType(this IEdmModel mod
...
+            {
+                throw new ODataException(Error.Format(SRResources.AmbiguousTypeNameFound, termName));
+            }
+
+            return terms.SingleOrDefault();
```
habbes, line 155: `Single` ensures that the returned item is unique, but we have already checked for duplicates in the block above. I think it's redundant to have both checks.

src/Microsoft.AspNetCore.OData/Edm/EdmModelExtensions.cs
```diff
@@ -102,6 +110,51 @@ public static IEdmStructuredTypeReference ResolveResourceType(this IEdmModel mod
...
+            var terms = model.SchemaElements.OfType<IEdmTerm>()
+                .Where(e => string.Equals(termName, e.FullName(), StringComparison.OrdinalIgnoreCase));
+
+            foreach (var refModels in model.ReferencedModels)
+            {
+                var refedTerms = refModels.SchemaElements.OfType<IEdmTerm>()
+                    .Where(e => string.Equals(termName, e.FullName(), StringComparison.OrdinalIgnoreCase));
+
+                terms = terms.Concat(refedTerms);
+            }
```
habbes, line 148: This looks like the kind of thing that should be implemented in EdmLib and provided as single method call. It is also expensive to scan all the referenced models if we have no term. Is this meant to support case-insensitive lookups? Should we also do case-insensitive lookups or is that something that the customer enables via configuration?

In ODL we created a cache for case-insensitive lookups of model elements (including `IEdmTerm`) for use with the case-insensitive `ODataUriResolver`. We could consider refactoring the cache and using it in more scenario. It was designed to avoid the cost of scanning the schemas frequently, which was causing a lot of CPU usage in workloads: 
https://github.com/OData/odata.net/pull/2610

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/DeserializationHelper.cs
```diff
@@ -276,28 +277,30 @@ internal static void SetProperty(object resource, string propertyName, object va
         internal static object ConvertValue(object oDataValue, ref IEdmTypeReference propertyType, IODataDeserializerProvider deserializerProvider,
             ODataDeserializerContext readContext, out EdmTypeKind typeKind)
         {
-            if (oDataValue == null)
+            if (oDataValue == null || oDataValue is ODataNullValue)
```
habbes, line 280: Why didn't we check `ODataNullValue` before? Was this a bug?

src/Microsoft.AspNetCore.OData/Edm/EdmModelAnnotationExtensions.cs
```diff
@@ -217,6 +217,34 @@ public static PropertyInfo GetDynamicPropertyDictionary(this IEdmModel edmModel,
...
+                throw Error.ArgumentNull(nameof(edmModel));
+            }
+
+            if (edmType == null)
+            {
+                throw Error.ArgumentNull(nameof(edmType));
+            }
+
+            InstanceAnnotationContainerAnnotation annotation =
+                edmModel.GetAnnotationValue<InstanceAnnotationContainerAnnotation>(edmType);
```
habbes, line 239: I hope we can figure out way to avoid this pattern because this has become a constant perf problem in OData. It's also a usability issue because customers wouldn't know which annotations need to be registered for which feature and we don't have good docs for it. I will investigate possible alternatives and share findings with the team.

src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs
```diff
@@ -26,6 +26,8 @@ namespace Microsoft.AspNetCore.OData.Extensions
     /// </summary>
     public static class HttpRequestExtensions
     {
+        private static readonly string ODataInstanceAnnotationContainerKey = "odataInstanceAnnotation_14802D58-69EF-4B28-9BDC-963D3648F06A";
```
habbes, line 29: Would `const` be better here since the value is known at compile-time? (I assume we'll never change this)
```suggestion
        private const string ODataInstanceAnnotationContainerKey = "odataInstanceAnnotation_14802D58-69EF-4B28-9BDC-963D3648F06A";
```

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/DeserializationHelper.cs
```diff
@@ -365,40 +374,106 @@ private static object GetProperty(object resource, string propertyName)
...
         {
             // Be noted: If a declared property (propertyType != null) is untyped (or collection),
             // It should be never come here. Because for collection untyped, it goes to nested resource set.
             // ODL reads the value as ODataResourceSet in a ODataNestedResourceInfo.
             // So, if it comes here, the untyped value is odata.type annotated. for example, create a ODataProperty using ODataCollectionValue
+
+            // 05/01/2024: new scenario, if we specify an instance annotation using the collection value as:
+            // ""Magics@NS.StringCollection"":[""Skyline"",7,""Beaver""],
+            // ODL generates 'ODataCollectionValue' without providing the 'TypeName' on ODataCollectionValue.
+            // So the above assumption could not be correct again.
```
habbes, line 388: If the above assumption is no longer correct due to the changes introduced, then we should also update the above comment to reflect the new behaviour.

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataDeserializerContext.cs
```diff
@@ -128,5 +132,78 @@ internal ODataDeserializerContext CloneWithoutType()
                 TimeZone = this.TimeZone
             };
         }
+
+        private CachedItem _cached = new CachedItem();
```
habbes, line 136: Is this cache only intended for use with the instance annotations container? If so, then maybe it would be better to give it a more specific name. Also, maybe it should moved up where the other private fields are located.

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataDeserializerContext.cs
```diff
@@ -128,5 +132,78 @@ internal ODataDeserializerContext CloneWithoutType()
...
+            {
+                value = propertyInfo.GetValue(resource);
+            }
+
+            IODataInstanceAnnotationContainer instanceAnnotationContainer = value as IODataInstanceAnnotationContainer;
+            if (instanceAnnotationContainer == null)
+            {
+                try
+                {
+                    if (propertyInfo.PropertyType == typeof(ODataInstanceAnnotationContainer) || propertyInfo.PropertyType == typeof(IODataInstanceAnnotationContainer))
```
habbes, line 175: This means we do support scenarios where the declared property is a different type that inherits from `IODataInstanceAnnotatioContainer`? Is that intentional?

For example, the following would not match:

```c#
public class Customer
{
    CustomInstanceAnnotationContainer InstanceAnnotations { get; set; }
}

public class CustomInstanceAnnotationContainer : IODataInstanceAnnotationContainer
{
    // ...
}
```

The following would also not be supported by this condition:

```c#
public class Customer
{
    IExtendedInstanceAnnotationContainer InstanceAnnotations { get; set; }
}

public interface IExtendedInstanceAnnotationContainer : IODataInstanceAnnotation
{
  // ...
}
```

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -152,9 +159,11 @@ public sealed override object ReadInline(object item, IEdmTypeReference edmType,
                 throw new ArgumentNullException(nameof(readContext));
             }
 
-            if (!String.IsNullOrEmpty(resourceWrapper.Resource.TypeName) &&
-                structuredType.FullName() != resourceWrapper.Resource.TypeName &&
-                resourceWrapper.Resource.TypeName != "Edm.Untyped")
+            string typeName = resourceWrapper.IsResourceValue ?
+                resourceWrapper.ResourceValue.TypeName :
+                resourceWrapper.Resource.TypeName;
```
habbes, line 164: I think have to check the `resourceWrapper.IsResourceValue` so you can call the right property is error-prone and unintuitive. I think @corranrogue9 also had some concerns around this approach and proposed using subclasses. If we don't take that approach, I'd suggest we at least expose a `resourceWrapper.GetTypeName()` method that hides the details and make it less likely that the user will make a mistake.

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -474,6 +484,67 @@ public virtual void ApplyDeletedResource(object resource, ODataResourceWrapper r
...
+            IEdmStructuredTypeReference structuredType, ODataDeserializerContext readContext)
+        {
+            if (resourceWrapper == null)
+            {
+                throw new ArgumentNullException(nameof(resourceWrapper));
+            }
+
+            ICollection<ODataInstanceAnnotation> annotations = resourceWrapper.IsResourceValue ?
+                resourceWrapper.ResourceValue.InstanceAnnotations :
+                resourceWrapper.Resource.InstanceAnnotations;
```
habbes, line 504: I think we should also expose a `resourceWrapper.GetInstanceAnnotations()` method. It's easy to make a mistake because it's not obvious that `ResourceValue` and `Resource` do different things. This could be confusing and lead to errors.

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -489,7 +560,11 @@ public virtual void ApplyDeletedResource(object resource, ODataResourceWrapper r
                 throw new ArgumentNullException(nameof(resourceWrapper));
             }
 
-            foreach (ODataProperty property in resourceWrapper.Resource.Properties)
+            IEnumerable<ODataProperty> properties = resourceWrapper.IsResourceValue ?
+                resourceWrapper.ResourceValue.Properties :
+                resourceWrapper.Resource.Properties;
```
habbes, line 565: Same concern as https://github.com/OData/AspNetCoreOData/pull/1259/files#r1684303079. Consider `resourceWrapper.GetProperties` or a different design for resource value and resource.

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -566,15 +685,19 @@ public virtual void ApplyDeletedResource(object resource, ODataResourceWrapper r
             object value = null;
             if (resourceWrapper != null)
             {
+                string typeName = resourceWrapper.IsResourceValue ?
+                    resourceWrapper.ResourceValue.TypeName :
+                    resourceWrapper.Resource.TypeName;
```
habbes, line 690: See: https://github.com/OData/AspNetCoreOData/pull/1259/files#r1684303079

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
habbes, line 164: See: https://github.com/OData/AspNetCoreOData/pull/1259/files#r1684303079

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
habbes, line 504: See: https://github.com/OData/AspNetCoreOData/pull/1259/files#r1684303079

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -566,15 +685,19 @@ public virtual void ApplyDeletedResource(object resource, ODataResourceWrapper r
             object value = null;
             if (resourceWrapper != null)
             {
+                string typeName = resourceWrapper.IsResourceValue ?
```
habbes, line 688: See https://github.com/OData/AspNetCoreOData/pull/1259/files#r1684303079

src/Microsoft.AspNetCore.OData/Formatter/ResourceContext.cs
```diff
@@ -18,6 +19,8 @@
 using Microsoft.AspNetCore.OData.Formatter.Value;
 using Microsoft.AspNetCore.OData.Query.Wrapper;
 using Microsoft.OData.Edm;
+using Microsoft.OData.ModelBuilder;
+using Microsoft.VisualBasic;
```
habbes, line 23: What is the `VisualBasic` namespace used for?

PR #1272, opened 2024-07-02
update MaxProtocolVersion

ElizabethOkerio:


gathogojr:
> This PR fixes: [OData/odata.net#3004](https://github.com/OData/odata.net/issues/3004) and [OData/odata.net#3009](https://github.com/OData/odata.net/issues/3009).
> 
> It sets the `MaxProtocolVersion` to have the same version as the `OData-Version`. Currently, V4 is set as the default Protocol Version. If someone uses the `V401 OData-Version`, the `MaxProtocolVersion` is not updated which causes exceptions. Even setting `OData-MaxVersion` does not fix this problem.

@ElizabethOkerio If `ODataMessageReaderSettings.Version` is set to V4, should `ODataMessageReaderSettings.MaxProtocolVersion` be set to `V401` or `V4`?

Issue #1269, opened 2024-06-25
Annotation to handle sanitization of field names in open types


marcus905:
**Assemblies affected**
ASP.NET Core OData 8.x

**Describe the bug**
There is no way to handle special characters if present in dynamic properties in OData, causing:

"The property name '01.01' is invalid; property names must not contain any of the reserved characters ':', '.', '@'."

**Reproduce steps**
in an IDictionary<string, object>, if the key contains a special character, on read it throws this error

**Data Model**
class Test {
        public int Id{ get; set; }

        //extradata
        public IDictionary<string, object> ExtraData { get; set; }
}

**EDM (CSDL) Model**
<edmx:Edmx xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx" Version="4.0">
<edmx:DataServices>
<Schema xmlns="http://docs.oasis-open.org/odata/ns/edm" Namespace="ODataExtractorsServiceLuxRY.Models">
<EntityType Name="Test" OpenType="true">
<Key>
<PropertyRef Name="Id"/>
</Key>
<Property Name="Id" Type="Edm.String" Nullable="false"/>
</EntityType>
</Schema>
</edmx:DataServices>
</edmx:Edmx>

**Request/Response**
N/A

**Expected behavior**
There needs to be, for the open type annotation, a way to specify that non-compliant field names must be sanitized before sending them out.

xuzhg:
Can you create a class implementing the 'IDictionary<string, object>' interface and add the verification codes within the interface method?

marcus905:
I probably can, but I shouldn't have to. It's a specification of a format and in my opinion it's expected for the library to have a facility to do that in open types where there is no other way to do that (besides your option probably, but that means reinventing a lot of the whole wheel). Having the option in the library instead would be way easier to implement, probably just a few lines of code.

If you want, I can try to provide a pull request. Tell me if you'd accept it, though, else it'd be wasted work.

My solve has been a view in the source DB to fix this, currently.

xuzhg:
@marcus905 We love to see your contribution.

marcus905:
Is there any news on this?

PR #1273, opened 2024-07-05
Add ReplaceIllegalFieldNameCharactersAttribute, add related tests, fixes issue #1269

marcus905:
As previously mentioned in #1269 this adds the attribute in the pull request title to replace illegal characters in field names with legal ones before serialization begins.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Formatter/Attributes/ReplaceIllegalFieldNameCharactersAttribute.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Attributes/ReplaceIllegalFieldNameCharactersAttribute.cs
@@ -1,0 +1,52 @@
+﻿using System;
+using System.Linq;
+
+namespace Microsoft.AspNetCore.OData.Formatter.Attributes
+{
+    [AttributeUsage(AttributeTargets.Class, Inherited = false, AllowMultiple = false)]
+    sealed class ReplaceIllegalFieldNameCharactersAttribute : Attribute
+    {
+        //constant collection of illegal characters
+        private static readonly string[] illegalChars = new string[] { "@", ":", "." };
+        public string ReplaceAt { get; }
+        public string ReplaceColon { get; }
+        public string ReplaceDot { get; }
+
+        public ReplaceIllegalFieldNameCharactersAttribute(string replaceAt, string replaceColon, string replaceDot)
+        {
+            //assert that the replacements are not illegal either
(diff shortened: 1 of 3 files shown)


marcus905:

src/Microsoft.AspNetCore.OData/Formatter/Attributes/ReplaceIllegalFieldNameCharactersAttribute.cs
```diff
@@ -0,0 +1,59 @@
+﻿using System;
+using System.Linq;
+
+namespace Microsoft.AspNetCore.OData.Formatter.Attributes
+{
+    [AttributeUsage(AttributeTargets.Class, Inherited = false, AllowMultiple = false)]
+    sealed class ReplaceIllegalFieldNameCharactersAttribute : Attribute
```
xuzhg, line 7: Use an attribute to change the property name looks weird to me. Since, it requires customers to change their data model by decorating this attribute. Most of time, it's not acceptable.

I think we can have a different option:

1) Create a service (an interface) named 'IEscapeCharacterService' or similar
2) Consume this service during 'AppendDynamicProperties'
3) If we have this service registered, calling it. Otherwise, let it go and no other impact for all existing customers

Where to register this service?
1) We can let it register into the DI, or
2) We can let it register into the Edm Model?

xuzhg (changes requested):

src/Microsoft.AspNetCore.OData/Formatter/Attributes/ReplaceIllegalFieldNameCharactersAttribute.cs
marcus905: If the user is already using the lib successfully, then there will be no need to apply that attribute and the renaming would not happen at all (no regressions on other tests, assert exception passed on the invalid mapped dynamic property dictionary) and so customers/user would be safe as no downstream code changes will be needed for normal operations.

If the user, instead, needs to fix the property names (and mind you, even if I do not foresee it being ever used in this way) this way, by annotating the model class, it would work on both dynamic and static properties which have a specific name set all the same.

I concede that the naming of the attribute might not be the best, but that can be changed.

Moreover, annotating entities in this way is a Model pattern classic (EF for example) used in several different libraries in other frameworks and languages.

I still consider this to be the best approach, but if this is deemed to not be acceptable for the project, I'll try yours.

Please let me know.

marcus905:
After some testing, the '#' char has a special meaning for most OData clients too (Power BI for example), so I added a way to replace it in the code.

I'm waiting for your review, thanks.

PR #1277, opened 2024-07-12
Update ODL versions to 8.0.0-rc.1

ElizabethOkerio:
Updated the ODL version to `8.0.0-rc.1`. Since `8.0.0-rc.1` had some breaking changes, I've also made some updates like: 

1. Updating `ODataResource.Properties` to a collection of `ODataPropertyInfo`

### DIFF
--- a/test/Microsoft.AspNetCore.OData.Tests/Formatter/Serialization/ODataResourceSerializerTests.cs
+++ b/test/Microsoft.AspNetCore.OData.Tests/Formatter/Serialization/ODataResourceSerializerTests.cs
@@ -1009,30 +1009,30 @@ namespace Microsoft.AspNetCore.OData.Tests.Formatter.Serialization
             Assert.Equal(6, resource.Properties.Count());
 
             // Verify the declared properties
-            ODataProperty street = Assert.Single(resource.Properties.Where(p => p.Name == "CustomerId"));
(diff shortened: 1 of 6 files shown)


habbes (approved):

test/Microsoft.AspNetCore.OData.Tests/Formatter/Serialization/ODataResourceSerializerTests.cs
```diff
@@ -1009,30 +1009,30 @@ public void CreateResource_Works_ToAppendDynamicProperties_ForOpenEntityType()
             Assert.Equal(6, resource.Properties.Count());
 
             // Verify the declared properties
-            ODataProperty street = Assert.Single(resource.Properties.Where(p => p.Name == "CustomerId"));
+            ODataProperty street = Assert.Single(resource.Properties.OfType<ODataProperty>().Where(p => p.Name == "CustomerId"));
```
gathogojr, line 1012: I think it makes better sense to do it this way:
```csharp
ODataProperty street = Assert.IsType<ODataProperty>(Assert.Single(resource.Properties.Where(p => p.Name == "CustomerId")));
```
This way, you're verifying that a single property with the name "CustomerId" exists in the collection and it's of type `ODataProperty`. We'd be able to catch a scenario where it exists but it's not of type `ODataProperty`.

I'd propose the same approach in other tests as well.

gathogojr (approved):

PR #1292, opened 2024-08-05
Fix AutoExpand Issues with Capitalized $Expand System Query Options

### LINKED ISSUE #1284: AutoExpand not working correctly with $Expand uri query parameters
**Assemblies affected**
Microsoft.AspNetCore.OData version 8.2.5

**Describe the bug**
According to [OASIS docs](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_SystemQueryOptions), system query options *must* support case-insensitive system query option names with or without a $ prefix.

When using AutoExpand or AutoSelect, and specifying a matching system query option that's not lower case (such as $Expand) an exception is thrown. The exception indicates that Expand is being added twice.

``` json
{
    "error": {
        "code": "",
        "message": "The query specified in the URI is not valid. Query option '$expand/expand' was specified more than once, but it must be specified at most once.",
        "details": [],
        "innererror": {
            "message": "Query option '$expand/expand' was specified more than once, but it must be specified at most once.",
            "type": "Microsoft.OData.ODataException",
            "stacktrace": "   at Microsoft.OData.UriParser.ODataQueryOptionParser.TryGetQueryOption(String name, String& value)\r\n   at Microsoft.OData.UriParser.ODataQueryOptionParser.ParseSelectAndExpand()\r\n   at Microsoft.AspNetCore.OData.Query.SelectExpandQueryOption.ProcessLevels()\r\n   at Microsoft.AspNetCore.OData.Query.SelectExpandQueryOption.get_ProcessedSelectExpandClause()\r\n   at Microsoft.AspNetCore.OData.Query.ODataQueryOptions.ApplySelectExpand[T](T entity, ODataQuerySettings querySettings)\r\n   at Microsoft.AspNetCore.OData.Query.ODataQueryOptions.ApplyTo(IQueryable query, ODataQuerySettings querySettings)\r\n   at Microsoft.AspNetCore.OData.Query.EnableQueryAttribute.ExecuteQuery(Object responseValue, IQueryable singleResultCollection, ControllerActionDescriptor actionDescriptor, HttpRequest request)\r\n   at Microsoft.AspNetCore.OData.Query.EnableQueryAttribute.OnActionExecuted(ActionExecutedContext actionExecutedContext, Object responseValue, IQueryable singleResultCollection, ControllerActionDescriptor actionDescriptor, HttpRequest request)"
        }
    }
}
```

I have managed to track the error to the function `AddAutoSelectExpandProperties` within `Microsoft.AspNetCore.Odata.Query.OdataQueryOptions.cs`.  

``` c#
	internal void AddAutoSelectExpandProperties()
	{
		bool containsAutoSelectExpandProperties = false;
		string autoExpandRawValue = GetAutoExpandRawValue();
		string autoSelectRawValue = GetAutoSelectRawValue();
		IDictionary<string, string> queryParameters = GetODataQueryParameters();
		if (!string.IsNullOrEmpty(autoExpandRawValue) && !autoExpandRawValue.Equals(RawValues.Expand, StringComparison.Ordinal))
		{
			queryParameters["$expand"] = autoExpandRawValue;
			containsAutoSelectExpandProperties = true;
		}
		else
		{
			autoExpandRawValue = RawValues.Expand;
		}
		if (!string.IsNullOrEmpty(autoSelectRawValue) && !autoSelectRawValue.Equals(RawValues.Select, StringComparison.Ordinal))
		{
			queryParameters["$select"] = autoSelectRawValue;
			containsAutoSelectExpandProperties = true;
		}
		else
		{
			autoSelectRawValue = RawValues.Select;
		}
		if (containsAutoSelectExpandProperties)
		{
			_queryOptionParser = new ODataQueryOptionParser(Context.Model, Context.ElementType, Context.NavigationSource, queryParameters, Context.RequestContainer);
			SelectExpandQueryOption originalSelectExpand = SelectExpand;
			SelectExpand = new SelectExpandQueryOption(autoSelectRawValue, autoExpandRawValue, Context, _queryOptionParser);
			if (originalSelectExpand != null && originalSelectExpand.LevelsMaxLiteralExpansionDepth > 0)
			{
				SelectExpand.LevelsMaxLiteralExpansionDepth = originalSelectExpand.LevelsMaxLiteralExpansionDepth;
			}
			else if (Context.ValidationSettings != null && Context.ValidationSettings.MaxExpansionDepth > 0)
			{
				SelectExpand.LevelsMaxLiteralExpansionDepth = Context.ValidationSettings.MaxExpansionDepth;
			}
		}
	}
```

The line below explicitly adds the auto expanded path values to the queryParameters dictionary with the key "$expand", in the case that the query parameters are also in lower case, this works, but if the query parameters are "$Expand", for example, instead of overriding the parameter a second queryParameter is added.
``` c#
queryParameters["$expand"] = autoExpandRawValue;
```

The simple solution may be to ensure that the query parameter is always in lower case...

``` c#
private IDictionary<string, string> GetODataQueryParameters()
  {
      Dictionary<string, string> result = new Dictionary<string, string>();

      foreach (var query in Request.Query)
      {
          string key = query.Key.Trim().ToLower();
          string value = query.Value.ToString();

      // code removed for brevity
      ...
     }
     return result;
}
```

**Expected behavior**
This shouldn't error and case on the Expand or Select query parameters should not matter.

WanjohiSammy:
### Issues

*This pull request fixes #1284.*

### Description

*When using **AutoExpand**, and specifying a system query option that's not lower case (such as $Expand) an exception is thrown. The exception indicates that Expand is being added twice.* 
```js
"message": "The query specified in the URI is not valid. Query option '$expand/expand' was specified more than once, but it must be specified at most once.",
```

This change is to resolve issues with the **AutoExpand** when using capitalized `$Expand` system query options

For example, given the following entities where `Friends` prop and `Trip` class are **AutoExpand**
```cs
public class Person
{
    public int Id { get; set; }

    public string? Name { get; set; }

    public IList<WeekDay>? WorkingDays { get; set; }

    [AutoExpand]
    public virtual ICollection<Person> Friends { get; set; }

    public virtual ICollection<Trip> Trips { get; set; }
}

[AutoExpand]
public class Trip
{
    public int Id { get; set; }

    public string? Name { get; set; }
}
```

With this change, we should be able to use `$Expand`(**capitalized of $expand**) with the **AutoExpand** navigation properties:
```cs
/People?$Expand=Friends
/People?$Expand=Trips
/People?$Expand=Friends,Trips
/People?$EXPAND=Friends
```

- [x] *Test cases added*

### Additional work necessary


src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
```diff
@@ -182,19 +183,7 @@ public static bool IsSystemQueryOption(string queryOptionName, bool isDollarSign
...
-                 fixedQueryOptionName.Equals("$count", StringComparison.Ordinal) ||
-                 fixedQueryOptionName.Equals("$expand", StringComparison.Ordinal) ||
-                 fixedQueryOptionName.Equals("$select", StringComparison.Ordinal) ||
-                 fixedQueryOptionName.Equals("$format", StringComparison.Ordinal) ||
-                 fixedQueryOptionName.Equals("$skiptoken", StringComparison.Ordinal) ||
-                 fixedQueryOptionName.Equals("$deltatoken", StringComparison.Ordinal) ||
-                 fixedQueryOptionName.Equals("$search", StringComparison.Ordinal) ||
-                 fixedQueryOptionName.Equals("$compute", StringComparison.Ordinal) ||
-                 fixedQueryOptionName.Equals("$apply", StringComparison.Ordinal);
+            return ODataQueryOptionConstants.QueryOptionSet.Contains(fixedQueryOptionName);
```
habbes, line 186: The intent here seems to improve readability/maintainability. But given that this list of comparisons is unlikely to change frequently (given that the number of query options is determine by the spec), I'm concerned whether the perceived improvement is worth the slight performance cost, especially given that this a method that we call for each query option of each request. Given this is a relatively small list of values to check, using a hashset is costlier than doing the comparisons even in the worst case (when the item is not a query option):

image


[Benchmark code](https://tinyurl.com/ycxdj9e5)

src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
```diff
@@ -182,19 +183,7 @@ public static bool IsSystemQueryOption(string queryOptionName, bool isDollarSign
                 fixedQueryOptionName = "$" + queryOptionName;
```
habbes, line 183: I think we can eliminate this allocation, but maybe that's outside the scope of this PR.

src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
```diff
@@ -873,7 +862,7 @@ internal void AddAutoSelectExpandProperties()
 
             foreach (var query in Request.Query)
             {
-                string key = query.Key.Trim();
+                string key = query.Key.Trim().ToLowerInvariant();
```
habbes, line 865: Is it possible to perform an `OrdinalIgnoreCase` comparison instead of doing `ToLowerInvariant()` to avoid extra string allocations?

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
```diff
@@ -86,6 +86,73 @@ public async Task QueryForAnResource_Includes_AutoExpandNavigationProperty(strin
...
+            string queryUrl = $"autoexpand/Customers(5)?{queryParams}";
+            HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Get, queryUrl);
+            request.Headers.Accept.Add(MediaTypeWithQualityHeaderValue.Parse("application/json;odata.metadata=none"));
+            HttpClient client = CreateClient();
+
+            // Act
+            HttpResponseMessage response = await client.SendAsync(request);
+
+            // Assert
+            Assert.NotNull(response);
```
habbes, line 104: Under what circumstances would `client.SendAsync(request)` return `null`? I think it throws an exception if it can't return a response, or?

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
```diff
@@ -488,5 +555,24 @@ public async Task PutCustomer_AutoExpandNavigationProperties()
                 }
             }
         }
+
+        private async Task VerifyNavigationPropertiesAndLevels(HttpResponseMessage response, int propCount)
+        {
+            var customer = await response.Content.ReadAsObject<JObject>().ConfigureAwait(false);
+            this.output.WriteLine(customer.ToString());
```
habbes, line 562: What is this writing to (and why)?

WanjohiSammy: @habbes This was writing the json response to Test Explorer console but it is not necessary. I will remove it.

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
WanjohiSammy: @habbes This is true, I will use `trycatch` and assert to ensure exception is null and response is not null:

```cs
...

HttpResponseMessage response = null;
Exception exception = null;

// Act
try
{
    response = await client.SendAsync(request);
}
catch (Exception ex)
{
    exception = ex;
}

// Assert
Assert.Null(exception);
Assert.NotNull(response);
```

src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
WanjohiSammy: @habbes This is possible. I have used `StringComparer.OrdinalIgnoreCase` instead of converting to `LowerCase()`
```cs
 Dictionary<string, string> result = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
```

src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
WanjohiSammy: @habbes Thanks for this analysis. I have updated the check to use string comparison instead. Thanks

WanjohiSammy:
> @WanjohiSammy does this issue only affect AutoExpand? Also do we have tests for query options without a $ prefix?

@habbes This addresses the issue of Select or Expand AutoExpand navigation property using capitalized `Select` or `Expand` with or without `$` prefix. I have added tests for query options without a $ prefix

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
```diff
@@ -86,6 +87,77 @@ public async Task QueryForAnResource_Includes_AutoExpandNavigationProperty(strin
...
+            {
+                response = await client.SendAsync(request);
+            }
+            catch (Exception ex)
+            {
+                exception = ex;
+            }
+
+            // Assert
+            Assert.Null(exception);
```
habbes, line 153: Why is this in a try catch if we expect the exception to be null? If no exception is expcted, maybe the try-catch is not necessary. If an exception is thrown and not caught, it would cause the test not to fail anyway.

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
```diff
@@ -86,6 +87,77 @@ public async Task QueryForAnResource_Includes_AutoExpandNavigationProperty(strin
...
```
habbes, line 115: Why is this in a try-catch that absorbs the exception if no exception is expected?

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
WanjohiSammy: @habbes No exception is expected. In case of any exception, the error will be caught, and the exception will no longer be null and hence the test (`Assert.Null(exception)`) will fail. I want the test to fail in case of any exception occurs in `client.SendAsync(request)`

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
WanjohiSammy: @habbes Same apply to here. No exception is expected. In case of any exception, the error will be caught, and the exception will no longer be null and hence the test (`Assert.Null(exception)`) will fail. I want the test to fail in case of any exception occurs in `client.SendAsync(request)`

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
```diff
@@ -86,6 +87,77 @@ public async Task QueryForAnResource_Includes_AutoExpandNavigationProperty(strin
...
+            string queryUrl = $"autoexpand/Customers(5)?{queryParams}";
+            HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Get, queryUrl);
+            request.Headers.Accept.Add(MediaTypeWithQualityHeaderValue.Parse("application/json;odata.metadata=none"));
+            HttpClient client = CreateClient();
+
+            HttpResponseMessage response = null;
+            Exception exception = null;
+
+            // Act
+            try
```
xuzhg, line 105: avoid the try and catch in the test case?

Why don't use "Assert.DoesThrow<T>" or "Assert.Throws<T>" ??

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
```diff
@@ -86,6 +87,77 @@ public async Task QueryForAnResource_Includes_AutoExpandNavigationProperty(strin
...
+            Assert.NotNull(response.Content);
+
+            await VerifyNavigationPropertiesAndLevels(response, propCount);
+        }
+
+        [Theory]
+        [InlineData("$Expand=Order & Select=Friend", 3)]
+        [InlineData("Expand=Order & $Select=Order", 3)]
+        [InlineData("$Expand=Order,HomeAddress/CountryOrRegion($Select=Name)", 4)]
+        [InlineData("$Expand=HomeAddress/CountryOrRegion,HomeAddress/Microsoft.AspNetCore.OData.E2E.Tests.AutoExpand.UsAddress/ZipCode,Order($Expand=Choice)", 4)]
```
xuzhg, line 127: does ODL support to case-insensitive query option key word?

WanjohiSammy: @xuzhg  `4.01` supports case-incentive system query options however if still using `4.0`, lower case query options are supported.

You can refer the Spec: [v4.01 #SystemQueryOptions](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_SystemQueryOptions)

Also here are sample examples in TripPin service:
- [/People?$Select=UserName](https://services.odata.org/TripPinRESTierService/(S(4dte2qgkqy1lkpudzkpnk4om))/People?$Select=UserName)
- [/People?$Filter=isof('Trippin.Employee')](https://services.odata.org/TripPinRESTierService/(S(4dte2qgkqy1lkpudzkpnk4om))/People?$Filter=isof(%27Trippin.Employee%27))

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
WanjohiSammy: @xuzhg Thanks for this.

src/Microsoft.AspNetCore.OData/Query/Query/ODataQueryOptionConstants.cs
```diff
@@ -0,0 +1,79 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.OData.Query.Query
+{
+    internal static class ODataQueryOptionConstants
+    {
+        /// <summary>
+        /// The $apply query option.
+        /// </summary>
+        public const string Apply = "$apply";
```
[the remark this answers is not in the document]
WanjohiSammy: @xuzhg I have used the strings directly as recommended

src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
WanjohiSammy: Can create another PR for optimizing this section

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
WanjohiSammy: @xuzhg This addresses the issue of Select or Expand `AutoExpand` navigation property using capitalized `Select` or `Expand` with or without $ prefix.

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
```diff
@@ -86,6 +87,55 @@ public async Task QueryForAnResource_Includes_AutoExpandNavigationProperty(strin
...
+            Assert.True(response.IsSuccessStatusCode);
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            Assert.NotNull(response.Content);
+
+            await VerifyNavigationPropertiesAndLevels(response, propCount);
+        }
+
+        [Theory]
+        [InlineData("$Expand=Order & Select=Friend", 3)]
+        [InlineData("Expand=Order & $Select=Order", 3)]
```
gathogojr, line 114: nit:
```suggestion
        [InlineData("$Expand=Order&Select=Friend", 3)]
        [InlineData("Expand=Order&$Select=Order", 3)]
```

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
```diff
@@ -86,6 +87,55 @@ public async Task QueryForAnResource_Includes_AutoExpandNavigationProperty(strin
...
+        [InlineData("$Expand=Order,HomeAddress/CountryOrRegion($Select=Name)", 4)]
+        [InlineData("$Expand=HomeAddress/CountryOrRegion,HomeAddress/Microsoft.AspNetCore.OData.E2E.Tests.AutoExpand.UsAddress/ZipCode,Order($Expand=Choice)", 4)]
+        [InlineData("Expand=HomeAddress/Microsoft.AspNetCore.OData.E2E.Tests.AutoExpand.UsAddress/ZipCode", 4)]
+        [InlineData("Expand=Friend(Select=Order) & $Select=Order", 3)]
+        [InlineData("Select=Order & Expand=Order(Expand=Choice(Select=Amount))", 3)]
```
gathogojr, line 119: nit:
```suggestion
        [InlineData("Select=Order&Expand=Order(Expand=Choice(Select=Amount))", 3)]
```

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
```diff
@@ -86,6 +87,55 @@ public async Task QueryForAnResource_Includes_AutoExpandNavigationProperty(strin
...
+        public async Task QueryForResources_WithExpandSelectQueryParamsCapitalized_IncludesAutoExpandNavigationProperty(string queryParams, int propCount)
```
gathogojr, line 120: I'm not sure I understand the parameters for this test. Is `$` optional for query options? Asking because I see `Expand` and `Select`. I think they'd not be treated as the system query options. They'd be ignored. However, if `AutoExpand` and/or `AutoSelect` are applied on the model, they'd be used.

src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
```diff
@@ -869,7 +869,7 @@ internal void AddAutoSelectExpandProperties()
 
         private IDictionary<string, string> GetODataQueryParameters()
         {
-            Dictionary<string, string> result = new Dictionary<string, string>();
+            Dictionary<string, string> result = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
```
gathogojr, line 872: I like this. Simple and elegant

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
```diff
@@ -488,5 +538,23 @@ public async Task PutCustomer_AutoExpandNavigationProperties()
                 }
             }
         }
+
+        private async Task VerifyNavigationPropertiesAndLevels(HttpResponseMessage response, int propCount)
```
gathogojr, line 542:
```suggestion
        private async Task VerifyNavigationProperties(HttpResponseMessage response, int propCount)
```

I also wish the method would accept a `JObject` instead of `HttpResponseMessage`, just like the other "Verify..." methods

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
```diff
@@ -86,6 +87,55 @@ public async Task QueryForAnResource_Includes_AutoExpandNavigationProperty(strin
             Assert.Null(friend["Order"]);
         }
 
+        [Theory]
+        [InlineData("$Select=Id", 4)]
+        [InlineData("$Select=Order", 3)]
+        public async Task QueryForResources_WithSelectQueryParamCapitalized_IncludesAutoExpandNavigationProperty(string queryParams, int propCount)
```
gathogojr, line 93: What about `AutoSelect`?

I think you should;
- Verify that properties specified on the `$select` (plus other variants, e.g. `$Select`, `$SELECT`, etc) query option together with properties decorated with the `AutoSelect` attribute on the model are handled together.
- Verify that properties specified on the `$expand` (plus other variants, e.g. `$Expand`, `$EXPAND`, etc) query option together with properties decorated with the `AutoExpand` attribute on the model are handled together.

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
WanjohiSammy: @gathogojr Thanks for pointing this out. I wanted to ensure also that the Query Params with no `$` sign, will still work. `EnableNoDollarQueryOptions` is set to `true` by default [ODataOptions.cs#EnableNoDollarQueryOptions](https://github.com/OData/AspNetCoreOData/blob/release-8.x/src/Microsoft.AspNetCore.OData/ODataOptions.cs#L288)

```cs
/// <summary>
/// Gets or sets whether or not the OData system query options should be prefixed with '$'.
/// </summary>
public bool EnableNoDollarQueryOptions { get; set; } = true;
```

src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
WanjohiSammy: This is also faster than converting the key to either LowerCase or UpperCase and checks if the dictionary contains that key.

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
WanjohiSammy: Fixed

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
WanjohiSammy: Applied

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
WanjohiSammy: @gathogojr This makes sense. I have added the tests:
- [AutoSelect](https://github.com/OData/AspNetCoreOData/blob/fix/1284-autoexpand-and-autoexpand-issue/test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs#L95)
- [AutoExpand](https://github.com/OData/AspNetCoreOData/blob/fix/1284-autoexpand-and-autoexpand-issue/test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs#L120)

test/Microsoft.AspNetCore.OData.E2E.Tests/AutoExpand/AutoExpandTests.cs
WanjohiSammy: @xuzhg `EnableCaseInsensitive` is by default true: https://github.com/OData/AspNetCoreOData/blob/release-8.x/src/Microsoft.AspNetCore.OData/ODataOptions.cs#L310

image

gathogojr (approved):

Issue #1294, opened 2024-08-06
ODataException when entity has DateOnly or TimeOnly member


andygjp:
**Assemblies affected**
Microsoft.AspNetCore.OData 9.0.0-rc.1
Microsoft.OData.Core 8.0.0-rc.1

**Describe the bug**
Cannot GET or POST entities that contain DateOnly and TimeOnly types after upgrading to v9 rc1

**Error**

Here is the error stacktrace:

```
An unhandled exception has occurred while executing the request.
      Microsoft.OData.ODataException: An ODataPrimitiveValue was instantiated with a value of type 'System.DateOnly'. ODataPrimitiveValue can only wrap values which can be represented as primitive EDM types.
         at Microsoft.OData.ODataPrimitiveValue..ctor(Object value)
         at Microsoft.AspNetCore.OData.Formatter.Serialization.ODataPrimitiveSerializer.CreatePrimitive(Object value, IEdmPrimitiveTypeReference primitiveType, ODataSerializerContext writeContext)
         at Microsoft.AspNetCore.OData.Formatter.Serialization.ODataPrimitiveSerializer.CreateODataPrimitiveValue(Object graph, IEdmPrimitiveTypeReference primitiveType, ODataSerializerContext writeContext)
         at Microsoft.AspNetCore.OData.Formatter.Serialization.ODataPrimitiveSerializer.CreateODataValue(Object graph, IEdmTypeReference expectedType, ODataSerializerContext writeContext)
         at Microsoft.AspNetCore.OData.Formatter.Serialization.ODataSerializerPropertyHelper.CreateProperty(IODataEdmTypeSerializer serializer, Object graph, IEdmTypeReference expectedType, String elementName, ODataSerializerContext writeContext)
         at Microsoft.AspNetCore.OData.Formatter.Serialization.ODataResourceSerializer.CreateStructuralProperty(IEdmStructuralProperty structuralProperty, ResourceContext resourceContext)
         at Microsoft.AspNetCore.OData.Formatter.Serialization.ODataResourceSerializer.CreateStructuralPropertyBag(SelectExpandNode selectExpandNode, ResourceContext resourceContext)
         at Microsoft.AspNetCore.OData.Formatter.Serialization.ODataResourceSerializer.CreateResource(SelectExpandNode selectExpandNode, ResourceContext resourceContext)
[...]

PR #1296, opened 2024-08-07
feat: adding support for frameworks greater than and equal to NET6

andygjp:
Fixes https://github.com/OData/AspNetCoreOData/issues/1294

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Common/TypeHelper.cs
+++ b/src/Microsoft.AspNetCore.OData/Common/TypeHelper.cs
@@ -93,7 +93,6 @@ namespace Microsoft.AspNetCore.OData.Common
             return Type.GetTypeCode(underlyingTypeOrSelf) == TypeCode.DateTime;
         }
 
-#if NET6_0_OR_GREATER
         /// <summary>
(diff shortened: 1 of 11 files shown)


andygjp:

src/Microsoft.AspNetCore.OData/Common/TypeHelper.cs
```diff
@@ -93,7 +93,7 @@ public static bool IsDateTime(Type clrType)
             return Type.GetTypeCode(underlyingTypeOrSelf) == TypeCode.DateTime;
         }
 
-#if NET6_0
+#if NET6_0_OR_GREATER
```
ElizabethOkerio, line 96: I think we need to get rid of the conditionals completely.

andygjp: I can do that

andygjp: Done

gathogojr (approved):

WanjohiSammy (approved):

PR #1308, opened 2024-09-06
Fix subtle bug when deserializing property without value

### LINKED ISSUE #1307: Subtle bug in deserializing payload with property without value
**Assemblies affected**
- Microsoft.AspNetCore.OData 9.x

**Describe the bug**
After introducing support for reading properties without value in OData .NET 9, there's a subtle bug in this line of code
https://github.com/OData/AspNetCoreOData/blob/1368836752963d88e920d9508d94807eb995894b/src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs#L492

The property type for `ODataResource.Properties` is `IEnumerable<ODataPropertyInfo>` and not `IEnumerable<ODataProperty>` as it was in OData .NET 8.

While the statement compiles okay, if an object of `ODataPropertyInfo` is present in the `Properties` collection property, an exception is thrown.

**Reproduce steps**
The issue can be reproduced using the following sample code:
```csharp
using System;

class ODataPropertyInfo
{
}

class ODataProperty : ODataPropertyInfo
{
    public object Value { get; set; }
}

class Program
{
    static void Main(string[] args)
    {
        var properties = new List<ODataPropertyInfo>()
        {
            new ODataProperty { Name = "Id", Value = 13 },
            new ODataPropertyInfo { Name = "Password" },
            new ODataProperty { Name = "LogonName", Value = "foobar" }
        };

        foreach (ODataProperty property in properties)
        {
            Console.WriteLine($"Name = {property.Name}, Value = {property.Value}");
        }
    }
}
```

Exception:
```
Unhandled Exception:
System.InvalidCastException: Specified cast is not valid.
  at Program.Main (System.String[] args) [0x0007a] in <86b2fbb858794e808f5e125cc32a50ce>:0 
[ERROR] FATAL UNHANDLED EXCEPTION: System.InvalidCastException: Specified cast is not valid.
  at Program.Main (System.String[] args) [0x0007a] in <86b2fbb858794e808f5e125cc32a50ce>:0
```

**Expected behavior**
The property without value (i,e,. `ODataPropertyInfo`) should be handled gracefully.

gathogojr:
Fixes #1307 by handling a property without value gracefully when deserializing OData payloads

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
@@ -489,8 +489,14 @@ public class ODataResourceDeserializer : ODataEdmTypeDeserializer
             throw new ArgumentNullException(nameof(resourceWrapper));
         }
 
-        foreach (ODataProperty property in resourceWrapper.Resource.Properties)
+        foreach (ODataPropertyInfo propertyInfo in resourceWrapper.Resource.Properties)
         {
+            if (!(propertyInfo is ODataProperty property))
+            {
+                // Cannot deserialize property without value
+                continue;
+            }
+
             ApplyStructuralProperty(resource, property, structuredType, readContext);
         }
     }
--- a/src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataReaderExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataReaderExtensions.cs
@@ -200,6 +200,10 @@ public static class ODataReaderExtensions
                 resourceSetParentWrapper.Items.Add(new ODataPrimitiveWrapper((ODataPrimitiveValue)reader.Item));
                 break;
 
+            case ODataReaderState.NestedProperty:
+                // Property without value - do nothing
+                break;
+
             default:
                 Contract.Assert(false, "We should never get here, it means the ODataReader reported a wrong state.");
                 break;
--- a/test/Microsoft.AspNetCore.OData.Tests/Formatter/Deserialization/ODataResourceDeserializerTests.cs
+++ b/test/Microsoft.AspNetCore.OData.Tests/Formatter/Deserialization/ODataResourceDeserializerTests.cs
@@ -1553,6 +1553,40 @@ public class ODataResourceDeserializerTests
             typeof(Product), _readContext).Wait(), "The property 'Concurrency' does not exist on type 'ODataDemo.Product'. Make sure to only use property names that are defined by the type or mark the type as open type.");
     }
 
+    [Fact]
+    public async Task ReadAsync_ForPropertyWithoutValue()
+    {
+        // Arrange
+        string content = "{" +
+            "\"ID\":0," +
+            "\"Name\":\"Bread\"," +
+            "\"Description\":\"Whole grain bread\"," +
+            "\"PublishDate\":\"1997-07-01\"," +
+            "\"ReleaseDate@odata.type\":\"#Edm.DateTimeOffset\"," + // OData annotation - Property without value
+            "\"DiscontinuedDate@Is.DateTimeOffset\":true," + // Custom annotation - Property without value
+            "\"Rating\":4," +
+            "\"Price\":2.5" +
+        "}";
+
+        ODataResourceDeserializer deserializer = new ODataResourceDeserializer(_deserializerProvider);
+
+        // Act
+        Product product = await deserializer.ReadAsync(
+            GetODataMessageReader(content, _edmModel),
+            typeof(Product),
+            _readContext) as Product;
+
+        // Assert
+        Assert.Equal(0, product.ID);
+        Assert.Equal(4, product.Rating);
+        Assert.Equal(2.5m, product.Price);
+        Assert.Equal(product.PublishDate, new Date(1997, 7, 1));
+        Assert.Equal("Bread", product.Name);
+        Assert.Equal("Whole grain bread", product.Description);
+        Assert.Null(product.ReleaseDate);
+        Assert.Null(product.DiscontinuedDate);
+    }
+
     private static ODataMessageReader GetODataMessageReader(IODataRequestMessage oDataRequestMessage, IEdmModel edmModel)
     {
         return new ODataMessageReader(oDataRequestMessage, new ODataMessageReaderSettings(), edmModel);


src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
```diff
@@ -489,8 +489,14 @@ public virtual void ApplyStructuralProperties(object resource, ODataResourceWrap
             throw new ArgumentNullException(nameof(resourceWrapper));
         }
 
-        foreach (ODataProperty property in resourceWrapper.Resource.Properties)
+        foreach (ODataPropertyInfo propertyInfo in resourceWrapper.Resource.Properties)
```
xuzhg, line 492: why don't use 'OfType<ODataProperty>', so we don't need line 494 to 499.

src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataReaderExtensions.cs
```diff
@@ -200,6 +200,10 @@ private static void ReadODataItem(ODataReader reader, Stack<ODataItemWrapper> it
                 resourceSetParentWrapper.Items.Add(new ODataPrimitiveWrapper((ODataPrimitiveValue)reader.Item));
                 break;
 
+            case ODataReaderState.NestedProperty:
+                // Property without value - do nothing
```
xuzhg, line 204: Why do nothing? need more details

test/Microsoft.AspNetCore.OData.Tests/Formatter/Deserialization/ODataResourceDeserializerTests.cs
```diff
@@ -1553,6 +1553,40 @@ public void ReadAsync_ThrowsOnUnknownEntityType()
...
+    public async Task ReadAsync_ForPropertyWithoutValue()
+    {
+        // Arrange
+        string content = "{" +
+            "\"ID\":0," +
+            "\"Name\":\"Bread\"," +
+            "\"Description\":\"Whole grain bread\"," +
+            "\"PublishDate\":\"1997-07-01\"," +
+            "\"ReleaseDate@odata.type\":\"#Edm.DateTimeOffset\"," + // OData annotation - Property without value
+            "\"DiscontinuedDate@Is.DateTimeOffset\":true," + // Custom annotation - Property without value
```
xuzhg, line 1566: How about the other type of properties with value? (Complex, enum)

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataResourceDeserializer.cs
gathogojr: It was deliberate. I believe `OfType` would be more expensive

gathogojr: @xuzhg Here's the code for a simple benchmark to compare the two:
```csharp
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Diagnosers;

[MemoryDiagnoser]
public class Benchmarks
{
    private IEnumerable<ODataPropertyInfo> properties;

    public Benchmarks()
    {
        List<ODataPropertyInfo> properties = new()
        {
            new ODataProperty { Name = "A", Value = 1 },
            new ODataProperty { Name = "A", Value = 1 },
            new ODataProperty { Name = "A", Value = 1 },
            new ODataProperty { Name = "A", Value = 1 },
            new ODataProperty { Name = "A", Value = 1 },
            new ODataProperty { Name = "A", Value = 1 },
            new ODataProperty { Name = "A", Value = 1 },
            new ODataProperty { Name = "A", Value = 1 },
            new ODataPropertyInfo { Name = "A" },
            new ODataProperty { Name = "A", Value = 1 },
            new ODataProperty { Name = "A", Value = 1 },
            new ODataProperty { Name = "A", Value = 1 },
            new ODataProperty { Name = "A", Value = 1 },
            new ODataPropertyInfo { Name = "A" },
            new ODataProperty { Name = "A", Value = 1 },
            new ODataProperty { Name = "A", Value = 1 },
        };

        this.properties = properties;
    }

    [Benchmark]
    public int CountPropertyValues_WithNestedCondition()
    {
        int count = 0;
        foreach (ODataPropertyInfo propertyInfo in properties)
        {
            if (propertyInfo is ODataProperty property) {
                count += property.Value;
            }
        }

        return count;
    }

    [Benchmark]
    public int CountPropertyValues_WithOfType()
    {
        int count = 0;
        foreach (ODataProperty property in properties.OfType<ODataProperty>())
        {
            count += property.Value;
        }

        return count;
    }
}

public class ODataPropertyInfo
{
}

public class ODataProperty : ODataPropertyInfo
{
    public int Value { get; set; }
}
```

And here are the results:
image

I think that LINQ overheads are responsible for the difference

src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataReaderExtensions.cs
xuzhg: I think we 'd add at least:
1) Why do we need to catch this reader state in this reading loop?
2) Why do we do nothing in this reader state?

xuzhg (approved):

src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataReaderExtensions.cs
habbes: Why didn't we need a case for this state before?

habbes (approved):

src/Microsoft.AspNetCore.OData/Formatter/Wrapper/ODataReaderExtensions.cs
gathogojr: @xuzhg @habbes The block for the `ODataReaderState.NestedProperty` should have been added when we upgraded the ODL dependency to the version that contained this change https://github.com/OData/odata.net/pull/2786.

In https://github.com/OData/odata.net/pull/2786, we introduced support for reading property without value. Before that, we would throw an exception if we came across a property without value. When we changed that, it meant that the `ODataReaderState.NestedProperty` state would be toggled going forward. It's just that after upgrading the dependency, we didn't add any test to verify behaviour when a payload containing a property without value was deserialized. It's easy to verify my theory here. If you include the test I added in ASP.NET Core OData `release-8.x` branch, the control flow will reach the `default` block and the `Debug.Assert` statement will fail. That's the reason we need the block that I introduced.

Issue #1223, opened 2024-04-23
Strange time zone propagation in the filter


cts-tradeit:
**Current**

Sample GET url http://localhost:51994/time-tracks?$filter=((dateFrom+eq+2024-04-17T09:45:00%2B02:00))&$orderby=dateFrom+desc&$skip=0&$top=25 contains time 2024-04-17T9:45:00.0000000+02:00

Produces SQL query that contains "2024-04-17T11:45:00.0000000+02:00". So it get's treated as if the date in the filter was UTC time.

The fault is in fact caused by DateTime being parsed as Kind.Unspecified (where expression in the query contains DateTime of this kind).
![image](https://github.com/OData/AspNetCoreOData/assets/71645109/42eee92d-3b89-4e07-a0cd-b04bd3d32079)

**Expected**
When oDataOptions.TimeZone is set in the configuration of the DateTime will be of Kind.Local and if the timezone in the filter is not the same as specified in TimeZone, the time will be shifted.

This would lead to result query with "2024-04-17T09:45:00.0000000+02:00".

cts-tradeit:
Related to #378 but not necesserily.

cts-tradeit:
Great solution would be to be able to specify custom deserializer for DateTime, or more precisely converter from DateTimeOffset to DateTime.

Yumitoya8569:
It is difficult to understand why such a serious and fundamental problem has not been taken seriously.
Though I don't really want to use the hacky method, it is the only way to fix the bug at present.
```C#
public static void HackDateTimeConvert()
{
    var targetType = AccessTools.TypeByName("Microsoft.AspNetCore.OData.Edm.EdmPrimitiveHelper");
    var targetMethod = AccessTools.Method(targetType, "ConvertPrimitiveValue", new Type[] { typeof(object), typeof(Type), typeof(TimeZoneInfo) });

    var prefix = new HarmonyMethod(typeof(ODataEdmStore).GetMethod("HackDateTimeConvertPrefix"));
    var harmony = new Harmony("Microsoft.AspNetCore.OData.Edm.Patch");
    harmony.Patch(targetMethod, prefix);
}

public static bool HackDateTimeConvertPrefix(ref object __result, object value, Type type, TimeZoneInfo timeZoneInfo)
{
    if (value.GetType() == type || value.GetType() == Nullable.GetUnderlyingType(type))
    {
        __result = value;
        return false;
    }

    if (type.IsInstanceOfType(value))
    {
[...]

PR #1316, opened 2024-09-20
Fix/datetime filter fix

VitaliyChaban:
fix error:
Cannot write DateTime with Kind=Unspecified to PostgreSQL type 'timestamp with time zone', only UTC is supported

also:
Fixes https://github.com/OData/AspNetCoreOData/issues/1223
Fixes https://github.com/OData/AspNetCoreOData/pull/1033
Fixes https://github.com/OData/AspNetCoreOData/issues/378

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Edm/EdmPrimitiveHelper.cs
+++ b/src/Microsoft.AspNetCore.OData/Edm/EdmPrimitiveHelper.cs
@@ -101,8 +101,9 @@ internal static class EdmPrimitiveHelper
             }
             else if (type == typeof(DateTime))
             {
-                if (value is DateTimeOffset dateTimeOffsetValue)
+                if (value is DateTimeOffset)
                 {
+                    DateTimeOffset dateTimeOffsetValue = (DateTimeOffset)value;
                     TimeZoneInfo timeZone = timeZoneInfo ?? TimeZoneInfo.Local;
                     dateTimeOffsetValue = TimeZoneInfo.ConvertTime(dateTimeOffsetValue, timeZone);
                     return DateTime.SpecifyKind(dateTimeOffsetValue.DateTime, DateTimeKind.Utc);
--- a/src/Microsoft.AspNetCore.OData/Edm/EdmPrimitiveHelper.cs
+++ b/src/Microsoft.AspNetCore.OData/Edm/EdmPrimitiveHelper.cs
@@ -105,8 +105,12 @@ internal static class EdmPrimitiveHelper
                 {
                     DateTimeOffset dateTimeOffsetValue = (DateTimeOffset)value;
                     TimeZoneInfo timeZone = timeZoneInfo ?? TimeZoneInfo.Local;
+
                     dateTimeOffsetValue = TimeZoneInfo.ConvertTime(dateTimeOffsetValue, timeZone);
-                    return DateTime.SpecifyKind(dateTimeOffsetValue.DateTime, DateTimeKind.Utc);
+
+                    var dateTimeKind = GetTargetDateTimeKind(timeZone);
+
+                    return DateTime.SpecifyKind(dateTimeOffsetValue.DateTime, dateTimeKind);
                 }
 
                 if (value is Date)
@@ -182,5 +186,16 @@ internal static class EdmPrimitiveHelper
                 }
             }
         }
+    }
+
+    private static DateTimeKind GetTargetDateTimeKind(TimeZoneInfo timeZone)
+    {
+        if (timeZone.Equals(TimeZoneInfo.Local))
+            return DateTimeKind.Local;
+
+        if (timeZone.Equals(TimeZoneInfo.Utc))
+            return DateTimeKind.Utc;
+
+        return DateTimeKind.Unspecified;
     }
 }
--- a/src/Microsoft.AspNetCore.OData/Edm/EdmPrimitiveHelper.cs
+++ b/src/Microsoft.AspNetCore.OData/Edm/EdmPrimitiveHelper.cs
@@ -101,12 +101,11 @@ internal static class EdmPrimitiveHelper
             }
             else if (type == typeof(DateTime))
             {
-                if (value is DateTimeOffset)
+                if (value is DateTimeOffset dateTimeOffsetValue)
                 {
-                    DateTimeOffset dateTimeOffsetValue = (DateTimeOffset)value;
                     TimeZoneInfo timeZone = timeZoneInfo ?? TimeZoneInfo.Local;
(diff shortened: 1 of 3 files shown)


VitaliyChaban:
Yes, I accept the agreement. What do I need to do for this?

@microsoft-github-policy-service agree company="DexTechnology"

VitaliyChaban:
Can someone review the changes and approve the pull request? 
There are very few changes, just one line not counting the auto-test. In terms of functionality, it only affects the date filtering with time and nothing else.

VitaliyChaban:
I added a more accurate setter for 'Kind' depending on TimeZoneInfo in the settings. 
All tests are passing, and there are new tests.

`    private static DateTimeKind GetTargetDateTimeKind(TimeZoneInfo timeZone)
    {
        if (timeZone.Equals(TimeZoneInfo.Local))
            return DateTimeKind.Local;

        if (timeZone.Equals(TimeZoneInfo.Utc))
            return DateTimeKind.Utc;

        return DateTimeKind.Unspecified;
    }`

xuzhg (dismissed):

VitaliyChaban:
@xuzhg 

Could you please let me know how often you conduct releases and when we can expect a new version with this fix? Will we have to wait for a scheduled release for some time, or can a version with these changes be released earlier (perhaps as a release candidate or something similar)?

6wanted9:
Hello! Any updates on this issue?

6wanted9:
@habbes @xuzhg take a look, please

6wanted9:
Guys, take a look please @xuzhg @marabooy @KenitoInc @habbes @mikepizzo @ElizabethOkerio @gathogojr

src/Microsoft.AspNetCore.OData/Edm/EdmPrimitiveHelper.cs
```diff
@@ -105,8 +105,12 @@ public static object ConvertPrimitiveValue(object value, Type type, TimeZoneInfo
                 {
                     DateTimeOffset dateTimeOffsetValue = (DateTimeOffset)value;
                     TimeZoneInfo timeZone = timeZoneInfo ?? TimeZoneInfo.Local;
+
                     dateTimeOffsetValue = TimeZoneInfo.ConvertTime(dateTimeOffsetValue, timeZone);
-                    return dateTimeOffsetValue.DateTime;
+
+                    var dateTimeKind = GetTargetDateTimeKind(timeZone);
```
WanjohiSammy, line 111: We prefer `explicit type` instead of `var`
```suggestion
                    DateTimeKind dateTimeKind = GetTargetDateTimeKind(timeZone);
```

src/Microsoft.AspNetCore.OData/Edm/EdmPrimitiveHelper.cs
```diff
@@ -183,4 +187,15 @@ public static object ConvertPrimitiveValue(object value, Type type, TimeZoneInfo
...
+
+    private static DateTimeKind GetTargetDateTimeKind(TimeZoneInfo timeZone)
+    {
+        if (timeZone.Equals(TimeZoneInfo.Local))
+            return DateTimeKind.Local;
+
+        if (timeZone.Equals(TimeZoneInfo.Utc))
+            return DateTimeKind.Utc;
+
+        return DateTimeKind.Unspecified;
```
WanjohiSammy, line 199: At what scenario will the `DateTimeKind.Unspecified` returned. Do you have a test for this.

WanjohiSammy (approved):

WanjohiSammy:
@VitaliyChaban @6wanted9 Could you resolved the failed unit tests.

VitaliyChaban:
> @VitaliyChaban @6wanted9 Could you resolved the failed unit tests.

Sure, I'll try to resolve the issue as soon as possible.

VitaliyChaban:
@WanjohiSammy 
Could you please run the tests on the server again?
I have accounted for the case where the server's local time is equivalent to UTC in the tests.

VitaliyChaban:
@WanjohiSammy 
I fixed it.
The behavior for implicitly specifying TimeZoneInfo.Local should be the same as for explicitly specifying it.

Sorry for bothering you with testing on the test bench) 
I wasn't able to reproduce the case where the system's local time = UTC right away

WanjohiSammy (dismissed):

gathogojr:
@VitaliyChaban Do you think you can include a unit test or E2E test for the scenario described [here](https://github.com/OData/AspNetCoreOData/issues/1223#issue-2259316131) that confirms that the date from the URL is parsed into the expected value?

gathogojr:
> > @VitaliyChaban Do you think you can include a unit test or E2E test for the scenario described [here](https://github.com/OData/AspNetCoreOData/issues/1223#issue-2259316131) that confirms that the date from the URL is parsed into the expected value?
> 
> I’m not very familiar with the context of E2E tests, so I’m not sure exactly what you mean.
> 
> The issue you’re describing should be fixed after my changes if you specify TimeZone = TimeZoneInfo.Utc in ODataSettings and in the query string use a dateTime with 'Z' at the end: '2024-04-17T09:45:00%2B02:00Z'
> 
> In that case, the original dateTime will be passed to EF, but with kind=utc, as expected.

Here's how the poster described the issue:
> Sample GET url http://localhost:51994/time-tracks?$filter=((dateFrom+eq+2024-04-17T09:45:00%2B02:00))&$orderby=dateFrom+desc&$skip=0&$top=25 contains time 2024-04-17T9:45:00.0000000+02:00
> 
> Produces SQL query that contains "2024-04-17T11:45:00.0000000+02:00". So it get's treated as if the date in the filter was UTC time.

The E2E test I had in mind is one that has a dependency on EF. One that would help us verify that your fix truly resolves the issue. You can check through the E2E tests in the repo to see if there are E2E tests with a dependency on EF that can guide you on how to proceed.

WanjohiSammy:
@VitaliyChaban Build is failing. Could you test this locally.
```
[xUnit.net 00:00:33.52]                                        ↓ (pos 641)
[xUnit.net 00:00:33.52]       Expected: ···" /><Property Name="PublishDay" Type="Edm.Date" /><Property N···
[xUnit.net 00:00:33.52]       Actual:   ···" /><Property Name="LastAction" Type="Edm.DateTimeOffset" Nul···
[xUnit.net 00:00:33.52]                                        ↑ (pos 641)

```

VitaliyChaban:
@WanjohiSammy 
I just revert it.
When saving a new field to the database, it is necessary to check it in other tests, and it's difficult to do this quickly during the workday, as  some of your tests use comparison with a constant string, while the database may slightly format the final value during serialization.

@gathogojr 
In any case, everything related to the conversion of the input dateTime (including kind and correct offset) is already covered by tests.

WanjohiSammy (approved):

PR #1322, opened 2024-10-07
Add dynamic properties support for any, all and in operators

gathogojr:
Add dynamic properties support for `any`, `all` and `in` operators.

Implements support for the following expressions
- `$filter=DynamicSingleValuedProperty in ('a','b','c')`
- `$filter=DynamicCollectionValuedProperty/any(d:d eq 4)`
- `$filter=DynamicCollectionValuedProperty/all(d: d gt 2)`

Closes https://github.com/OData/odata.net/issues/3059

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
@@ -97,6 +97,9 @@ public abstract partial class QueryBinder
             case QueryNodeKind.CollectionPropertyAccess:
                 return BindCollectionPropertyAccessNode(node as CollectionPropertyAccessNode, context);
 
+            case QueryNodeKind.CollectionOpenPropertyAccess:
+                return BindCollectionOpenPropertyAccessNode(node as CollectionOpenPropertyAccessNode, context);
+
             case QueryNodeKind.CollectionComplexNode:
                 return BindCollectionComplexNode(node as CollectionComplexNode, context);
 
@@ -108,7 +111,6 @@ public abstract partial class QueryBinder
 
             case QueryNodeKind.CollectionFunctionCall:
             case QueryNodeKind.CollectionResourceFunctionCall:
-            case QueryNodeKind.CollectionOpenPropertyAccess:
             default:
                 throw Error.NotSupported(SRResources.QueryNodeBindingNotSupported, node.Kind, typeof(QueryBinder).Name);
         }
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinderContext.cs
```diff
@@ -235,7 +235,18 @@ public void RemoveParameter(string name)
...
+            if (edmTypeReference != null)
+            {
+                clrType = Model.GetClrType(edmTypeReference, AssembliesResolver);
+            }
+            else
+            {
+                // Edm type reference will be null in a dynamic property scenario
+                clrType = typeof(object);
+            }
+
```
marabooy, line 248:
```suggestion
            // Edm type reference will be null in a dynamic property scenario hence we use the object type.
            Type clrType = null = edmTypeReference != null ? Model.GetClrType(edmTypeReference, AssembliesResolver) : typeof(object)
```
nit: Not sure if this is better but it compresses the initialization to a much smaller text window for readability, though ternary operator can also be less readable in some cases.

marabooy (approved):

test/Microsoft.AspNetCore.OData.E2E.Tests/DollarFilter/DollarFilterDataSource.cs
```diff
@@ -204,11 +208,624 @@ static DollarFilterDataSource()
         };
 
         #endregion Bad Vendors
+
+        #region Customers
+
+        customers = new List<Customer>
```
gathogojr, line 214: There's a deliberate aggressiveness with the tests and the test data. It was to ensure that as many scenarios as I could anticipate are adequately tested, especially given the dynamic nature of the LINQ expressions

src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
```diff
@@ -697,7 +735,7 @@ public virtual Expression BindAllNode(AllNode allNode, QueryBinderContext contex
     {
         CheckArgumentNull(allNode, context);
 
-       // context.EnterLambdaScope();
+        // context.EnterLambdaScope();
```
habbes, line 738: Do we still need this?

marabooy (approved):

src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
```diff
@@ -1487,42 +1546,42 @@ private Expression BindDynamicPropertyAccessExpression(SingleValueOpenPropertyAc
     /// <summary>
     /// Creates an expression for retrieving a dynamic property from the dynamic properties container property.
     /// </summary>
-    /// <param name="dynamicPropertiesContainerExpr">The dynamic properties container property access expression.</param>
+    /// <param name="containerPropertyAccessExpr">The dynamic properties container property access expression.</param>
```
WanjohiSammy, line 1549: Since you are passing `containerPropertyAccessExpr` and not `dynamicPropertiesContainerExpr`, should you also change the doc string for this parameter as well
```suggestion
    /// <param name="containerPropertyAccessExpr">The  properties container property access expression.</param>
```

src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
```diff
@@ -1487,42 +1546,42 @@ private Expression BindDynamicPropertyAccessExpression(SingleValueOpenPropertyAc
...
         IndexExpression dictionaryIndexExpr = Expression.Property(
-            dynamicPropertiesContainerExpr, DictionaryStringObjectIndexerName,
+            containerPropertyAccessExpr, DictionaryStringObjectIndexerName,
             Expression.Constant(propertyName));
 
         // ContainsKey method
         MethodCallExpression containsKeyExpression = Expression.Call(
-            dynamicPropertiesContainerExpr,
-            dynamicPropertiesContainerExpr.Type.GetMethod("ContainsKey"),
+            containerPropertyAccessExpr,
```
WanjohiSammy, line 1565: What is the different between `dynamicPropertiesContainerExpr` and `containerPropertyAccessExpr`. Or do they mean the same and only the name had changed.

habbes (dismissed):

WanjohiSammy (dismissed):

src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
gathogojr: @WanjohiSammy I just thought I should harmonize the name with how it's named elsewhere. Otherwise, both can be used to refer to the property of type `IDictionary<string, object>` (or any of it's implementing types) that we use to store dynamic properties. We can refer to it as the "dynamic properties container property" (mouthful/with 'property' repeated) or just "container property" (shorter).

src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
gathogojr: Maybe `The container properties property access expression`. Otherwise, `The properties container properties access expression` would sound inaccurate given the explanation I have provided here https://github.com/OData/AspNetCoreOData/pull/1322#discussion_r1832318661
"The dynamic properties container property access expression" is captures the definition accurately too

habbes (approved):

marabooy (approved):

WanjohiSammy (approved):

PR #1330, opened 2024-10-25
Add dynamic single-valued property support for filter expressions

gathogojr:
Add dynamic single-valued property support for filter expressions

Scenarios:
- $filter={DynamicSingleValuedProperty}/{DynamicSingleValuedProperty} eq {value}
- $filter={DynamicSingleValuedProperty}/{DynamicSingleValuedProperty}/{DynamicSingleValuedProperty} eq {value}
- $filter={DynamicSingleValuedProperty}/{DeclaredSingleValuedProperty}/{DynamicSingleValuedProperty} eq {value}
- $filter={DynamicSingleValuedProperty}/{DynamicSingleValuedProperty}/{DeclaredSingleValuedProperty} eq {value}
- $filter={DeclaredSingleValuedProperty}/{DynamicSingleValuedProperty} eq {value}
- $filter={DeclaredSingleValuedProperty}/{DynamicSingleValuedProperty}/{DynamicSingleValuedProperty} eq {value}
- $filter={DeclaredSingleValuedProperty}/{DeclaredSingleValuedProperty}/{DynamicSingleValuedProperty} eq {value}
- $filter={DeclaredSingleValuedProperty}/{DynamicSingleValuedProperty}/{DeclaredSingleValuedProperty} eq {value}
- $filter={DynamicSingleValuedProperty}/{DynamicSingleValuedProperty} in {collection}
- $filter={DynamicSingleValuedProperty}/{DynamicSingleValuedProperty}/{DynamicSingleValuedProperty} in {collection}
- $filter={DynamicSingleValuedProperty}/{DeclaredSingleValuedProperty}/{DynamicSingleValuedProperty} in {collection}
- $filter={DynamicSingleValuedProperty}/{DynamicSingleValuedProperty}/{DeclaredSingleValuedProperty} in {collection}
- $filter={DeclaredSingleValuedProperty}/{DynamicSingleValuedProperty} in {collection}
- $filter={DeclaredSingleValuedProperty}/{DynamicSingleValuedProperty}/{DynamicSingleValuedProperty} in {collection}
- $filter={DeclaredSingleValuedProperty}/{DeclaredSingleValuedProperty}/{DynamicSingleValuedProperty} in {collection}
- $filter={DeclaredSingleValuedProperty}/{DynamicSingleValuedProperty}/{DeclaredSingleValuedProperty} in {collection}

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
@@ -24,6 +24,7 @@ using Microsoft.OData;
 using Microsoft.OData.Edm;
 using Microsoft.OData.ModelBuilder;
 using Microsoft.OData.UriParser;
+using Microsoft.VisualBasic;
 
 namespace Microsoft.AspNetCore.OData.Query.Expressions;
 
@@ -297,30 +298,13 @@ public abstract partial class QueryBinder
             return Bind(computedProperty.Expression, context);
         }
 
-        PropertyInfo prop = GetDynamicPropertyContainer(openNode, context);
-
-        var propertyAccessExpression = BindPropertyAccessExpression(openNode, prop, context);
-        var readDictionaryIndexerExpression = Expression.Property(propertyAccessExpression,
-            DictionaryStringObjectIndexerName, Expression.Constant(openNode.Name));
-        var containsKeyExpression = Expression.Call(propertyAccessExpression,
-            propertyAccessExpression.Type.GetMethod("ContainsKey"), Expression.Constant(openNode.Name));
-        var nullExpression = Expression.Constant(null);
-
-        if (context.QuerySettings.HandleNullPropagation == HandleNullPropagationOption.True)
+        if (openNode.Source is SingleValueOpenPropertyAccessNode)
         {
-            var dynamicDictIsNotNull = Expression.NotEqual(propertyAccessExpression, Expression.Constant(null));
-            var dynamicDictIsNotNullAndContainsKey = Expression.AndAlso(dynamicDictIsNotNull, containsKeyExpression);
-            return Expression.Condition(
-                dynamicDictIsNotNullAndContainsKey,
-                readDictionaryIndexerExpression,
-                nullExpression);
+            return BindNestedDynamicPropertyAccessExpression(openNode, context);
         }
         else
         {
-            return Expression.Condition(
-                containsKeyExpression,
-                readDictionaryIndexerExpression,
-                nullExpression);
+            return BindDynamicPropertyAccessExpression(openNode, context);
         }
     }
 
@@ -978,7 +962,7 @@ public abstract partial class QueryBinder
     /// <summary>
     /// Gets property for dynamic properties dictionary.
     /// </summary>
-    /// <param name="openNode"></param>
+    /// <param name="openNode">The single-valued open property access node.</param>
     /// <param name="context">The query binder context.</param>
     /// <returns>Returns CLR property for dynamic properties container.</returns>
     protected static PropertyInfo GetDynamicPropertyContainer(SingleValueOpenPropertyAccessNode openNode, QueryBinderContext context)
@@ -993,22 +977,7 @@ public abstract partial class QueryBinder
             throw Error.ArgumentNull(nameof(context));
         }
 
-        IEdmStructuredType edmStructuredType;
-        IEdmTypeReference edmTypeReference = openNode.Source.TypeReference;
-        if (edmTypeReference.IsEntity())
-        {
-            edmStructuredType = edmTypeReference.AsEntity().EntityDefinition();
-        }
-        else if (edmTypeReference.IsComplex())
-        {
-            edmStructuredType = edmTypeReference.AsComplex().ComplexDefinition();
-        }
-        else
-        {
-            throw Error.NotSupported(SRResources.QueryNodeBindingNotSupported, openNode.Kind, typeof(QueryBinder).Name);
-        }
-
-        return context.Model.GetDynamicPropertyDictionary(edmStructuredType);
+        return GetDynamicPropertyContainer(openNode.Source.TypeReference, openNode.Kind, context.Model);
     }
 
     /// <summary>
@@ -1322,6 +1291,11 @@ public abstract partial class QueryBinder
             constantType = Nullable.GetUnderlyingType(constantType) ?? constantType;
         }
 
+        if (edmTypeReference != null && edmTypeReference.IsUntyped())
+        {
+            constantType = typeof(object);
+        }
+
         return constantType;
     }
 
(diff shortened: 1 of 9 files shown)


WanjohiSammy (approved):

src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
```diff
@@ -1453,6 +1469,180 @@ protected static Expression ApplyNullPropagationForFilterBody(Expression body, Q
...
+            // Get declared property
+            Expression getPropertyExpr = Expression.Call(
+                getTypeExpr,
+                typeof(Type).GetMethod("GetProperty", new[] { typeof(string) }),
+                Expression.Constant(openNode.Name));
+
+            // Get declared property value
+            Expression getDeclaredPropertyValueExpr = Expression.Convert(
+                Expression.Call(getPropertyExpr, "GetValue", Type.EmptyTypes, sourceExpr),
+                typeof(object));
```
habbes, line 1531: Does this convert the result of PropertyInfo.GetValue() into an `object`? Isn't the return type of `GetValue()` already `object`?

src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
```diff
@@ -1453,6 +1469,180 @@ protected static Expression ApplyNullPropagationForFilterBody(Expression body, Q
...
+            Expression getTypeExpr = Expression.Call(sourceExpr, "GetType", Type.EmptyTypes);
+
+            // Get declared property
+            Expression getPropertyExpr = Expression.Call(
+                getTypeExpr,
+                typeof(Type).GetMethod("GetProperty", new[] { typeof(string) }),
+                Expression.Constant(openNode.Name));
+
+            // Get declared property value
+            Expression getDeclaredPropertyValueExpr = Expression.Convert(
```
habbes, line 1529: Just for my own sanity check, this expression is the equivalent of `(object)DynamicParentProperty.GetType().GetProperty("DeclareProperty").GetValue(DynamicParentPropertyInstance)`?

habbes (approved):

src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
gathogojr: @habbes I'll check if the Convert is redundant here

src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
gathogojr: @gathogojr Think of it this way...
The call to `BindDynamicPropertyAccessExpression` resolves into something like:
```csharp
var dynamicPropertyValue = containerDict.Item["{dynamicPropertyName}"]
```
Then, the statements in this `else` block do something like:
```csharp
Type dynamicPropertyType = dynamicPropertyValue.GetType();

PropertyInfo propertyInfo = dynamicPropertyType.GetProperty("{nestedPropertyName}");

/* ... Check that propertyInfo is not null ... */

object nestedDeclaredPropertyValue = propertyInfo.GetValue(dynamicPropertyValue);
```
The `nestedDeclaredPropertyValue` is returned if indeed the nested property is a declared property.

The logic in scenario 2 on the other hand does the equivalent of:
```csharp
IEdmTypeReference edmTypeReference = context.Model.GetEdmTypeReference(dynamicPropertyType);

PropertyInfo containerPropertyInfo = GetDynamicPropertyContainer(edmTypeReference, nodeTypeKind, context.Model);

/* ... Check that containerPropertyInfo is not null ... */

Dictionary<string, object> nestedContainerDict = containerPropertyInfo.GetValue(dynamicPropertyValue);

object nestedDynamicPropertyValue = nestedContainerDict.Item["{nestedPropertyName}"]
```
The `nestedDynamicPropertyValue` is returned if indeed the nested property is a dynamic property property.

src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
```diff
@@ -1453,6 +1469,180 @@ protected static Expression ApplyNullPropagationForFilterBody(Expression body, Q
...
+        Expression sourceExpr = null;
+
+        if (openNode.Source is SingleValueOpenPropertyAccessNode openNodeParent)
+        {
+            sourceExpr = BindNestedDynamicPropertyAccessExpression(openNodeParent, context);
+        }
+
+        if (!(openNode.Source is SingleValueOpenPropertyAccessNode))
+        {
+            return BindDynamicPropertyAccessExpression(openNode, context);
```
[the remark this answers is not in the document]
gathogojr: No. This is `DynamicProperty/NestedDeclaredProperty` and `DynamicProperty/NestedDynamicProperty` scenarios.

Note however that based on the output of the parser, both `NestedDeclaredProperty` and `NestedDynamicProperty` will be dynamic! To the parser, everything that follows a dynamic segment is a dynamic segment. We can't know at parse time that nested segment might resolve into a declared segment.

src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
```diff
@@ -1453,6 +1469,180 @@ protected static Expression ApplyNullPropagationForFilterBody(Expression body, Q
...
+        }
+        else
+        {
```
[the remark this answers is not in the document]
gathogojr: @habbes I'll try the refactor that you suggested but let me first explain the logic in this method.
Consider either of these two expressions:
- `Products?$filter=DynamicSegment/NestedDynamicSegment1/NestedDynamicSegment2/NestedDynamicSegment3`
- `Products?$filter=DeclaredSegment/NestedDynamicSegment1/NestedDynamicSegment2/NestedDynamicSegment3`

In both cases, the query binder will first attempt to bind `NestedDynamicSegment3` to it's source (i.e. `NestedDynamicSegment2`). But we don't know the type for `NestedDynamicSegment2` because it's also an open node. It's the same challenge if we attempt to bind `NestedDynamicSegment2` to it's source (i.e. `NestedDynamicSegment3`).

For this reason, what this method first does is recursively call itself until we find **a node whose source is not an open node**.
Meaning recursion will stop if we find a node whose source is not `SingleValueOpenPropertyAccessNode`.
When we come across such a node, we bind it to its source which would either be a `SingleValueNode` or a `ResourceRangeVariableReferenceNode`.

Binding happens from the `BindDynamicPropertyAccessExpression` method that returns to us the expression that retrieves the dynamic property at the base of the path expression. The expression is the equivalent of:
```csharp
{productInstance}.containerDict.Item["DynamicSegment"]
```
for the first expression, and
```csharp
{productInstance}.DeclaredSegment.containerDict.Item["NestedDynamicSegment"]
```
for the second expression.

We then extend that expressions such that it can retrieve `NestedDynamicSegment2` and `NestedDynamicSegment2` as either declared or dynamic properties of the subsequent segment.

src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
gathogojr: @habbes Fixed

WanjohiSammy (approved):

habbes (approved):

PR #1334, opened 2024-10-28
Fixes #1332: Deep Comparison for Array-based Concurrency tokens.

### LINKED ISSUE #1332: IfNoneMatch.ApplyTo does not work for byte[] fields.
**Assemblies affected**
ASP.NET Core OData 8.x

**Describe the bug**

**Reproduce steps**

**Data Model**
```
public class ETagsCustomer
    {
        public int Id { get; set; }
        
        [ConcurrencyCheck]
        [Contained]
    }
```

**EDM (CSDL) Model**

```
        {
            return builder.GetEdmModel();
        }
```

**Request/Response**

<details>
<summary>Click me</summary>

```
        {
            // Arrange
            string eTag;

            // So, this case uses "4"
            int customerId = 4;

            // Act
            {
                // Assert

                Assert.NotNull(result);

            }

            // Arrange & Act
            {
                // Assert
            }
        }
```

</details>

**Expected behavior**

**Screenshots**
N/A

**Additional context**

anasik:
Fixes #1332

A popular convention today for concurrency check fields is to use a `byte[]` field called `RowVersion` with a `Timestamp` or `ConcurrencyCheck` attribute.

However, when using such a field, the `queryOptions.ifNoneMatch.ApplyTo` call was having absolutely no effect. That's because `ApplyTo` was internally using `Expression.Equal` which is perfect for value types but doesn't fare very well against reference types or particularly Arrays.

I fixed this by adding a case for arrays that, through a helper method, generates an expression that performs a deep comparison.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/ETag.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/ETag.cs
@@ -136,7 +136,11 @@ namespace Microsoft.AspNetCore.OData.Query
                 Expression value = itemValue != null
                     ? LinqParameterContainer.Parameterize(itemValue.GetType(), itemValue)
                     : Expression.Constant(value: null);
-                BinaryExpression equal = Expression.Equal(name, value);
+                Expression equal;
+                if (itemValue != null && itemValue.GetType().IsArray)
+                    equal = ExpressionHelpers.SequenceEquals(name, value);
+                else
+                    equal = Expression.Equal(name, value);
                 where = where == null ? equal : Expression.AndAlso(where, equal);
             }
 
--- a/src/Microsoft.AspNetCore.OData/Query/ETag.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/ETag.cs
@@ -133,14 +133,27 @@ namespace Microsoft.AspNetCore.OData.Query
             {
                 MemberExpression name = Expression.Property(param, item.Key);
                 object itemValue = item.Value;
-                Expression value = itemValue != null
-                    ? LinqParameterContainer.Parameterize(itemValue.GetType(), itemValue)
-                    : Expression.Constant(value: null);
+                
                 Expression equal;
-                if (itemValue != null && itemValue.GetType().IsArray)
-                    equal = ExpressionHelpers.SequenceEquals(name, value);
+                if (itemValue != null)
+                {
+                    Type itemType = itemValue.GetType();
+                    Expression value = LinqParameterContainer.Parameterize(itemType, itemValue);
+                    if (itemType.IsArray)
+                    {
+                        equal = ExpressionHelpers.SequenceEquals(name, value);
+                    }
+                    else
+                    {
+                        equal = Expression.Equal(name, value);
+                    }
+                }
                 else
+                {
+                    Expression value = Expression.Constant(value: null);
                     equal = Expression.Equal(name, value);
+                }
+                
                 where = where == null ? equal : Expression.AndAlso(where, equal);
             }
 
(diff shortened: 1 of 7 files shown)


anasik:
@WanjohiSammy , may I interest you in this?

anasik:
@xuzhg, it's been a little quiet here. Maybe I can help with something?

anasik:
@WanjohiSammy, bumping this up just in case.

test/Microsoft.AspNetCore.OData.E2E.Tests/ETags/DerivedETagTests.cs
```diff
@@ -60,6 +61,8 @@ private static IEdmModel GetDerivedEdmModel()
             EntitySetConfiguration<ETagsDerivedCustomer> eTagsDerivedCustomersSet = builder.EntitySet<ETagsDerivedCustomer>("ETagsDerivedCustomers");
             eTagsDerivedCustomersSet.HasRequiredBinding(c => c.RelatedCustomer, eTagsCustomersSet);
             eTagsDerivedCustomersSet.HasRequiredBinding(c => c.ContainedCustomer, eTagsCustomersSet);
+            eTagsDerivedCustomersSet.EntityType.Ignore(d=>d.RowVersion);
```
WanjohiSammy, line 64: What's the purpose of excluding this property from EdmModel

WanjohiSammy:
@anasik tests are missing. Can you add several tests for this Deep Comparison feature

anasik:
> @anasik tests are missing. Can you add several tests for this Deep Comparison feature

@WanjohiSammy,
All I did was prove that an existing test was failing under certain circumstances and then  I passed that test by adding the deep comparison feature. Since there's already a test that passes only when the feature exists, I am not sure what an additional test would look like?

If there's something specific you're looking for in terms of tests, please let me know and I'll see what I can do?

WanjohiSammy (approved):
I want to test and make sure that this is not a breaking change

anasik:
> I want to test and make sure that this is not a breaking change

@WanjohiSammy, makes perfect sense. Let me know if any tests are failing and I'll take a look.

anasik:
@WanjohiSammy are all existing build tests passing?

anasik:
@WanjohiSammy , are you waiting on me for something here?

WanjohiSammy (dismissed):

WanjohiSammy:
@habbes @gathogojr @marabooy please assist looking at this PR.

src/Microsoft.AspNetCore.OData/Query/ETag.cs
```diff
@@ -133,10 +133,27 @@ public virtual IQueryable ApplyTo(IQueryable query)
             {
                 MemberExpression name = Expression.Property(param, item.Key);
                 object itemValue = item.Value;
-                Expression value = itemValue != null
-                    ? LinqParameterContainer.Parameterize(itemValue.GetType(), itemValue)
-                    : Expression.Constant(value: null);
-                BinaryExpression equal = Expression.Equal(name, value);
+                
+                Expression equal;
+                if (itemValue != null)
```
habbes, line 138: @gathogojr made additional changes here to avoid duplicate calls to `itemValue.GetType()` and to improve overall readability the code.

habbes (approved):

gathogojr (approved):

gathogojr (approved):

anasik:
> Thank you @anasik for your contribution

@gathogojr , it was a pleasure. I hope to contribute more.

PR #1342, opened 2024-11-08
Bump version to 9.1.0

WanjohiSammy:
### Description
- Bump AspNetCore.OData version to 9.1.0

- [x] *Test cases added*


test/Microsoft.AspNetCore.OData.Tests/Query/Validator/FilterQueryValidatorTests.cs
```diff
@@ -820,7 +820,7 @@ public void OtherFunctions_SomeSingleParameterCasts_ThrowODataException(AllowedF
         {
             AllowedFunctions = AllowedFunctions.None,
         };
-        var expectedMessage = "Cast or IsOf Function must have a type in its arguments.";
+        var expectedMessage = $"Function '{unusedName}' is not allowed. To allow it, set the 'AllowedFunctions' property on EnableQueryAttribute or QueryValidationSettings.";
```
gathogojr, line 823: How come the error message here is quite different from what it was before?

gathogojr: Are we still testing the scenario requiring `Cast` or `IsOf` functions to have a type in the argument (quoted or unquoted)?

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/FilterBinderTests.cs
```diff
@@ -2639,38 +2643,56 @@ public void IsOfQuotedUndefinedTarget_ThrowsODataException(string filter)
...
     {
         // Arrange & Act & Assert
         ExceptionAssert.Throws<ODataException>(() => BindFilterAndVerify<Product>(filter), expectedMessage);
     }
 
+    [Theory]
+    [InlineData("isof(Microsoft.AspNetCore.OData.Tests.Models.DerivedProduct)")]
+    [InlineData("isof(Category,Microsoft.AspNetCore.OData.Tests.Models.DerivedCategory)")]
+    [InlineData("isof(Category, Microsoft.AspNetCore.OData.Tests.Models.DerivedCategory)")]
+    public void IsOfUnquotedEntityType_DoNotThrowsODataException(string filter)
```
gathogojr, line 2687: nit:
```suggestion
    public void IsOfUnquotedEntityType_DoesNotThrowODataException(string filter)
```

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/FilterBinderTests.cs
```diff
@@ -2639,38 +2643,56 @@ public void IsOfQuotedUndefinedTarget_ThrowsODataException(string filter)
...
+        var exception = Record.Exception(() => BindFilterAndVerify<Product>(filter));
+
+        // Assert
+        Assert.Null(exception);
+    }
+
+    [Theory]
     [InlineData("isof(null,Microsoft.AspNetCore.OData.Tests.Models.DerivedCategory)",
-        "Encountered invalid type cast. 'Microsoft.AspNetCore.OData.Tests.Models.DerivedCategory' is not assignable from 'Microsoft.AspNetCore.OData.Tests.Models.Product'.")]
+        "Encountered invalid type cast. 'Microsoft.AspNetCore.OData.Tests.Models.DerivedCategory' is not assignable from '<null>'.")]
```
gathogojr, line 2674: Why `<null>` and not the type that `Microsoft.AspNetCore.OData.Tests.Models.DerivedCategory` is not assignable to?

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/FilterBinderTests.cs
```diff
@@ -2639,38 +2643,56 @@ public void IsOfQuotedUndefinedTarget_ThrowsODataException(string filter)
...
 
         // Act & Assert
         ExceptionAssert.Throws<ODataException>(() => BindFilterAndVerify<Product>(filter), expectedMessage);
     }
 
     [Theory]
-    [InlineData("isof(Microsoft.AspNetCore.OData.Tests.Models.DerivedProduct)", "Cast or IsOf Function must have a type in its arguments.")]
+    [InlineData("isof(SupplierAddress,Microsoft.AspNetCore.OData.Tests.Models.Address)")]
+    [InlineData("isof(SupplierAddress, Microsoft.AspNetCore.OData.Tests.Models.Address)")]
+    public void IsOfUnquotedComplexType_DoNotThrowsODataException(string filter)
```
gathogojr, line 2663: nit:
```suggestion
    public void IsOfUnquotedComplexType_DoesNotThrowODataException(string filter)
```

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/FilterBinderTests.cs
```diff
@@ -2639,38 +2643,56 @@ public void IsOfQuotedUndefinedTarget_ThrowsODataException(string filter)
...
     [Theory]
-    [InlineData("isof(Microsoft.AspNetCore.OData.Tests.Models.Address)")]
-    [InlineData("isof(null,Microsoft.AspNetCore.OData.Tests.Models.Address)")]
-    [InlineData("isof(null, Microsoft.AspNetCore.OData.Tests.Models.Address)")]
-    [InlineData("isof(SupplierAddress,Microsoft.AspNetCore.OData.Tests.Models.Address)")]
-    [InlineData("isof(SupplierAddress, Microsoft.AspNetCore.OData.Tests.Models.Address)")]
-    public void IsOfUnquotedComplexType_ThrowsODataException(string filter)
+    [InlineData("isof(Microsoft.AspNetCore.OData.Tests.Models.Address)", "Microsoft.AspNetCore.OData.Tests.Models.Product")]
+    [InlineData("isof(null,Microsoft.AspNetCore.OData.Tests.Models.Address)", "<null>")]
+    [InlineData("isof(null, Microsoft.AspNetCore.OData.Tests.Models.Address)", "<null>")]
```
gathogojr, line 2648: Given the inputs of the `isof` function in this test, isn't the test name `IsOfUnquotedComplexType_ThrowsODataException` a bit misleading since your pull request actually introduces support for unquoted typename?

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/FilterBinderTests.cs
```diff
@@ -2448,36 +2457,31 @@ public void CastToQuotedComplexType_Succeeds(string filter)
...
+                    "cast(Category, Microsoft.AspNetCore.OData.Tests.Models.DerivedCategory)/DerivedCategoryName eq null"
                 },
             };
         }
     }
 
     [Theory]
     [MemberData(nameof(CastToUnquotedEntityType))]
-    public void CastToUnquotedEntityType_ThrowsODataException(string filter, string expectedMessage)
+    public void CastToUnquotedEntityType_DoNotThrowsODataException(string filter)
```
gathogojr, line 2478: nit:
```suggestion
    public void CastToUnquotedEntityType_DoesNotThrowODataException(string filter)
```

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/FilterBinderTests.cs
```diff
@@ -2391,33 +2390,43 @@ public void CastToQuotedPrimitiveType_Succeeds(string filter)
...
-            "'Microsoft.AspNetCore.OData.Tests.Models.Address' is not assignable from 'Microsoft.AspNetCore.OData.Tests.Models.Product'.";
+            $"'Microsoft.AspNetCore.OData.Tests.Models.Address' is not assignable from '{assignableFrom}'.";
 
         // Act & Assert
         ExceptionAssert.Throws<ODataException>(() => BindFilterAndVerify<Product>(filter), expectedMessage);
     }
 
+    [Theory]
+    [InlineData("cast(SupplierAddress, Microsoft.AspNetCore.OData.Tests.Models.Address) eq null")]
+    public void CastToUnquotedComplexType_DoNotThrowsODataException(string filter)
```
gathogojr, line 2421: nit:
```suggestion
    public void CastToUnquotedComplexType_DoesNotThrowODataException(string filter)
```

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/FilterBinderTests.cs
```diff
@@ -2240,10 +2240,9 @@ public void Cast_UnsupportedTarget_ReturnsNull(string filter)
     // in FilterQueryValidatorTest.  ODL's ODataQueryOptionParser and FunctionCallBinder call the code throwing these exceptions.
     [Theory]
     [InlineData("cast(null,Microsoft.AspNetCore.OData.Tests.Models.Address) ne null",
-        "Encountered invalid type cast. " +
-        "'Microsoft.AspNetCore.OData.Tests.Models.Address' is not assignable from 'Microsoft.AspNetCore.OData.Tests.Models.DataTypes'.")]
+        "Encountered invalid type cast. 'Microsoft.AspNetCore.OData.Tests.Models.Address' is not assignable from '<null>'.")]
```
gathogojr, line 2243: In general, I'm curious about the `<null>` in the error message. Are unable to determine the type that type in the cast function is not assignable to?

test/Microsoft.AspNetCore.OData.E2E.Tests/IsOf/IsOfContext.cs
```diff
@@ -0,0 +1,151 @@
...
+    public ProductsContext(DbContextOptions<ProductsContext> options)
+        : base(options)
+    {
+    }
+
+    public DbSet<Product> Products { get; set; }
+
+    protected override void OnModelCreating(ModelBuilder modelBuilder)
+    {
+        var splitStringConverter1 = new ValueConverter<IEnumerable<string>, string>(v => string.Join(";", v), v => v.Split(new[] { ';' }));
```
gathogojr, line 121: Where are we using `splitStringConverter1`?

test/Microsoft.AspNetCore.OData.Tests/Query/Validator/FilterQueryValidatorTests.cs
WanjohiSammy: The error message that was there before was because the ODL did not know how to handle `unquoted` type params.

```
Cast or IsOf Function must have a type in its arguments.
```

Now that ODL supports unquoted type params in `cast` and `isof`, no exception is thrown from ODL

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/FilterBinderTests.cs
WanjohiSammy: Thanks for pointing this out. I will change the name to something like this: `IsOfUnquotedComplexType_ThrowsNotAssignableFromException`

Based on the [isof spec](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_isof), `The two parameter` **isof** `function returns true if the object referred to by the expression is assignable to the type specified, according to the same rules, otherwise it returns false`, I can walk you through the below example:

```cs
namespace NS;

public class Product
{
    public int ProductID { get; set; }
    // Others properties
    public Address SupplierAddress { get; set; }
}

public class DerivedProduct : Address
{
    // Others properties
}

public class Address
{
    // Others properties
    public string City { get; set; }
}
```
With the examples below with `unquoted type params`
1. `/products?$filter=cast(NS.DerivedProduct)`
  ODL will cast Product to DerivedProduct and since DerivedProduct is of type Product, no exception will be thrown
 
2. `/products?$filter=cast(NS.Address)`
  ODL will cast Product to Address. This will throw an exception because the NS.Address is not assignable from NS.Product
 
3. ` /products?$filter=cast(null, NS.DerivedProduct)`
  ODL will try to cast null to DerivedProduct.  The error message:
  `Encountered invalid type cast. 'NS.DerivedProduct' is not assignable from '<null>'`
 
 4. `/products?$filter=cast(SupplierAddress, NS.DerivedProduct)`
  ODL will try to castSupplierAddress property to DerivedProduct. Since the type DerivedProduct cannot be assignable from SupplierAddress of type Address, we will get an exception
  `Encountered invalid type cast. 'NS.Address' is not assignable from 'NS.DerivedProperty'.`

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/FilterBinderTests.cs
WanjohiSammy: This is resolved Thanks

test/Microsoft.AspNetCore.OData.E2E.Tests/IsOf/IsOfContext.cs
WanjohiSammy: Thanks for pointing this out. I have removed the unnecessary variable

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/FilterBinderTests.cs
WanjohiSammy: Lets say you have the following data model:
```cs
namespace NS;

public class Product
{
    public int ProductID { get; set; }
    // Others properties
    public Address SupplierAddress { get; set; }
}

public class DerivedProduct : Product
{
   public string AnotherProperty { get; set; }
    // Others properties
}

public class Address
{
    // Others properties
    public string City { get; set; }
}
```

Executing the query `/products?$filter=cast(null,NS.DerivedProduct)/AnotherProperty eq 'something'` will give you the exception:
```
Encountered invalid type cast. 'NS.DerivedProduct' is not assignable from '<null>'.
```

This is so because the ODL will check if `NS.DerivedProduct` is related to `null` using the call to [UriEdmHelpers.CheckRelatedTo(parentType, childType)](https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/UriParser/Binders/DottedIdentifierBinder.cs#L104) where `parentType` will be **null** and `childType` will be **NS.DerivedProduct**

Since **parentType** is null, the **parentTypeName** will be set to `<null>`

Here is a screenshot of similar scenario:

image

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/FilterBinderTests.cs
WanjohiSammy: As describe in the https://github.com/OData/AspNetCoreOData/pull/1342#discussion_r1836592658 and https://github.com/OData/AspNetCoreOData/pull/1342#discussion_r1836290759, for `isof(null,Microsoft.AspNetCore.OData.Tests.Models.DerivedCategory)`, the ODL will try to check if the type of null is related to the type of `Microsoft.AspNetCore.OData.Tests.Models.DerivedCategory`

[CheckRelatedTo(IEdmType parentType, IEdmType childType)](https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/UriParser/UriEdmHelpers.cs#L51)

test/Microsoft.AspNetCore.OData.E2E.Tests/Cast/CastContext.cs
```diff
@@ -9,6 +9,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
+using System.Reflection.Metadata;
```
gathogojr, line 12: I don't see any change that would be using this namespace.

test/Microsoft.AspNetCore.OData.E2E.Tests/IsOf/IsOfTest.cs
```diff
@@ -0,0 +1,104 @@
...
+                combinations.Add(dataSourceType, "?$filter=isof(Location,'Microsoft.AspNetCore.OData.E2E.Tests.IsOf.MyOtherAddress')", 3);
+            }
+
+            return combinations;
+        }
+    }
+
+    [Theory]
+    [MemberData(nameof(Combinations))]
+    public async Task Cast_Query_WorksAsExpected(string dataSourceMode, string dollarFormat, int expectedEntityCount)
```
gathogojr, line 85: Should the second parameter be `dollarFormat` or `dollarFilter`?

test/Microsoft.AspNetCore.OData.E2E.Tests/IsOf/IsOfTest.cs
```diff
@@ -0,0 +1,104 @@
...
+        HttpResponseMessage response = await Client.GetAsync(requestUri);
+        JObject responseString = await response.Content.ReadAsObject<JObject>();
+
+        // Assert
+        Assert.True(HttpStatusCode.OK == response.StatusCode,
+            string.Format("Response status code, expected: {0}, actual: {1}, request url: {2}",
+            HttpStatusCode.OK, response.StatusCode, requestUri));
+
+        JArray value = responseString["value"] as JArray;
+        Assert.True(expectedEntityCount == value.Count,
```
gathogojr, line 100: See my comment here https://github.com/OData/AspNetCoreOData/pull/1342/files#r1836690256

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/FilterBinderTests.cs
```diff
@@ -2391,33 +2390,43 @@ public void CastToQuotedPrimitiveType_Succeeds(string filter)
...
-            "'Microsoft.AspNetCore.OData.Tests.Models.Address' is not assignable from 'Microsoft.AspNetCore.OData.Tests.Models.Product'.";
+            $"'Microsoft.AspNetCore.OData.Tests.Models.Address' is not assignable from '{assignableFrom}'.";
 
         // Act & Assert
         ExceptionAssert.Throws<ODataException>(() => BindFilterAndVerify<Product>(filter), expectedMessage);
     }
 
+    [Theory]
+    [InlineData("cast(SupplierAddress, Microsoft.AspNetCore.OData.Tests.Models.Address) eq null")]
+    public void CastToUnquotedComplexType_DoNotThrowODataException(string filter)
```
gathogojr, line 2421: Here and elsewhere prefer "DoesNotThrow" instead of "DoNotThrow"/"DoNotThrows"

test/Microsoft.AspNetCore.OData.E2E.Tests/IsOf/IsOfTest.cs
```diff
@@ -0,0 +1,104 @@
...
+    {
+        // Arrange
+        var requestUri = string.Format("{0}/Products{1}", dataSourceMode, dollarFormat);
+
+        // Act
+        HttpResponseMessage response = await Client.GetAsync(requestUri);
+        JObject responseString = await response.Content.ReadAsObject<JObject>();
+
+        // Assert
+        Assert.True(HttpStatusCode.OK == response.StatusCode,
```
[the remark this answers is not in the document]
WanjohiSammy: Used `Assert.True` because it allows custom user message as oppose to `Assert.Equal`
```cs
...
Assert.True(HttpStatusCode.OK == response.StatusCode,
    string.Format("Response status code, expected: {0}, actual: {1}, request url: {2}",
    HttpStatusCode.OK, response.StatusCode, requestUri));
...
Assert.True(expectedEntityCount == value.Count,
    string.Format("The entity count in response, expected: {0}, actual: {1}, request url: {2}",
    expectedEntityCount, value.Count, requestUri));

...
```
Source code:
[Assert.True(bool condition, string userMessage)](https://github.com/xunit/assert.xunit/blob/main/BooleanAsserts.cs#L115)

[Assert.Equal(T expected, T actual)](https://github.com/xunit/assert.xunit/blob/main/EqualityAsserts.cs#L116)

test/Microsoft.AspNetCore.OData.E2E.Tests/IsOf/IsOfTest.cs
WanjohiSammy: Same applied here,  I used `Assert.True` because it allows custom user message as opposed to `Assert.Equal`
```cs
...
Assert.True(expectedEntityCount == value.Count,
    string.Format("The entity count in response, expected: {0}, actual: {1}, request url: {2}",
    expectedEntityCount, value.Count, requestUri));

...
```

test/Microsoft.AspNetCore.OData.E2E.Tests/Cast/CastContext.cs
WanjohiSammy: Removed unnecessary imports

test/Microsoft.AspNetCore.OData.E2E.Tests/IsOf/IsOfTest.cs
WanjohiSammy: It should be `dollarFilter`. Resolved.

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/FilterBinderTests.cs
WanjohiSammy: Resolved.

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/FilterBinderTests.cs
habbes: Curious what the error message looks like when you have other arguments other than `null` that cannot be cast to the target type.

gathogojr (dismissed):

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/FilterBinderTests.cs
WanjohiSammy: @habbes I have added more tests to show the error message for scenarios where cast is not possible other `null`:

### isof
- `isof(Microsoft.AspNetCore.OData.Tests.Models.DerivedCategory)` 
  "Encountered invalid type cast. 'Microsoft.AspNetCore.OData.Tests.Models.DerivedCategory' is not assignable from 'Microsoft.AspNetCore.OData.Tests.Models.Product'."
- `isof(null,Microsoft.AspNetCore.OData.Tests.Models.DerivedCategory)` 
   "Encountered invalid type cast. 'Microsoft.AspNetCore.OData.Tests.Models.DerivedCategory' is not assignable from '<null>'."
- `isof(SupplierAddress, Microsoft.AspNetCore.OData.Tests.Models.DerivedCategory)` 
   "Encountered invalid type cast. 'Microsoft.AspNetCore.OData.Tests.Models.DerivedCategory' is not assignable from 'Microsoft.AspNetCore.OData.Tests.Models.Address'."
- `isof(Category, Microsoft.AspNetCore.OData.Tests.Models.Address)` 
   "Encountered invalid type cast. 'Microsoft.AspNetCore.OData.Tests.Models.Address' is not assignable from 'Microsoft.AspNetCore.OData.Tests.Models.Category'."

### cast
- `cast(Microsoft.AspNetCore.OData.Tests.Models.Address) eq null`
   "Encountered invalid type cast. 'Microsoft.AspNetCore.OData.Tests.Models.Address' is not assignable from 'Microsoft.AspNetCore.OData.Tests.Models.Product'."
- `cast(null, Microsoft.AspNetCore.OData.Tests.Models.Address) eq null`
   "Encountered invalid type cast. 'Microsoft.AspNetCore.OData.Tests.Models.Address' is not assignable from '<null>'."
- `cast(Category, Microsoft.AspNetCore.OData.Tests.Models.Product)/ProductID eq 1`
   "Encountered invalid type cast. 'Microsoft.AspNetCore.OData.Tests.Models.Product' is not assignable from 'Microsoft.AspNetCore.OData.Tests.Models.Category'."
- `cast(Category, Microsoft.AspNetCore.OData.Tests.Models.Address)/HouseNumber eq 1245`
   "Encountered invalid type cast.  'Microsoft.AspNetCore.OData.Tests.Models.Address' is not assignable from 'Microsoft.AspNetCore.OData.Tests.Models.Category'."

gathogojr (approved):

xuzhg (approved):

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/FilterBinderTests.cs
```diff
@@ -2639,38 +2642,59 @@ public void IsOfQuotedUndefinedTarget_ThrowsODataException(string filter)
...
+    [InlineData("isof(Microsoft.AspNetCore.OData.Tests.Models.Address)", "Microsoft.AspNetCore.OData.Tests.Models.Product")]
+    [InlineData("isof(null,Microsoft.AspNetCore.OData.Tests.Models.Address)", "<null>")]
+    [InlineData("isof(null, Microsoft.AspNetCore.OData.Tests.Models.Address)", "<null>")]
+    public void IsOfUnquotedComplexType_ThrowsNotAssignableFromException(string filter, string assignableFrom)
     {
         // Arrange
         var expectedMessage =
             "Encountered invalid type cast. " +
-            "'Microsoft.AspNetCore.OData.Tests.Models.Address' is not assignable from 'Microsoft.AspNetCore.OData.Tests.Models.Product'.";
+            $"'Microsoft.AspNetCore.OData.Tests.Models.Address' is not assignable from '{assignableFrom}'.";
```
habbes, line 2653: I wonder why `isof` throws an exception when the cast is not possible instead of returning `false`. Is that how it has always worked? It seems counter-intuitive to me.

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/FilterBinderTests.cs
habbes: @WanjohiSammy based on your explanation of the `isof` spec, I would expect it return `false` if the first operand cannot be cast to the second operand. Why does it throw an exception instead?

WanjohiSammy: @habbes Good question. The exception thrown is from ODL when unable to relate the types. For example, `/products?$filter=cast(SupplierAddress, NS.DerivedProduct)`, ODL will try to relate the type of `SupplierAddress` and the type of `NS.DerivedProduct` and in case they are not releated to each other, it will throw such an exception. The method responsible for that in ODL is [CheckRelatedTo(IEdmType parentType, IEdmType childType)](https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/UriParser/UriEdmHelpers.cs#L51)

Probably we should handle this exception in ODL for the case of `isof` and `cast` and return false instead of throwing this exception.

test/Microsoft.AspNetCore.OData.Tests/Query/Validator/FilterQueryValidatorTests.cs
WanjohiSammy: Will handle `Cast` and `IsOf` in a different PR

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/FilterBinderTests.cs
WanjohiSammy: @habbes @gathogojr Will handle `Cast` and `IsOf` in a different PR when updating ODL version

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/FilterBinderTests.cs
WanjohiSammy: Will handle `Cast` and `IsOf` in a different PR when updating ODL version

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/FilterBinderTests.cs
WanjohiSammy: @habbes @gathogojr Will handle `Cast` and `IsOf` in a different PR when updating ODL version

test/Microsoft.AspNetCore.OData.Tests/Query/Expressions/FilterBinderTests.cs
WanjohiSammy: I will create a different PR with `isof` and `cast` support

habbes (approved):

gathogojr (approved):

PR #1344, opened 2024-11-12
update DI settings

### LINKED ISSUE #1337: ODataMessageWriterSettings are registered as Singleton in DI but has ODataUri that are scoped to the request
**Assemblies affected**
ASP.NET Core OData 8.x

I have been having random 500 InternalServer errors when multiple responses are being written simultaneously. It often fails during construction of odata.context url as the ODataUri in the singleton instance of the ODataMessageWriterSettings are not matching the response being written.

I suspect that it is due to it being registered as Singleton instead of Scoped.

If I add it to DI as Scoped instead the problem seems to go away.

ODataServiceCollectionExtensions.cs line 56...

```
        services.AddSingleton(new ODataMessageWriterSettings
        {
            EnableMessageStreamDisposal = false,
            MessageQuotas = new ODataMessageQuotas { MaxReceivedMessageSize = Int64.MaxValue },
        });

```

ODataMessageWriterSettings.cs line 206...

```
        /// <summary>
        /// The OData Uri of an incoming request.  Call <see cref="ODataUriParser"/>'s methods,
        /// and assign properties (e.g., <see cref="ODataPath"/>) to <see cref="ODataUri"/>.
        /// </summary>
        public ODataUri ODataUri
        {
            get { return this.odataUri ?? (this.odataUri = new ODataUri()); }
            set { this.odataUri = value; }
        }

```

marabooy:
Message reader and Writer have the wrong scope in the DI container.
This could cause some issues whenever users retrieve and write to them in the scope of a request.

Fixes #1337 by updating the dependencies to 8.0.2 as well as moving reader and writer settings to scoped.
This includes https://github.com/OData/odata.net/pull/3058

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Abstracts/ODataServiceCollectionExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Abstracts/ODataServiceCollectionExtensions.cs
@@ -38,7 +38,7 @@ internal static class ODataServiceCollectionExtensions
         // ReaderSettings and WriterSettings are registered as prototype services.
         // There will be a copy (if it is accessed) of each prototype for each request.
 #pragma warning disable CS0618 // ReadUntypedAsString is obsolete in ODL 8.
-        services.AddSingleton(new ODataMessageReaderSettings
+        services.AddScoped(sp => new ODataMessageReaderSettings
         {
             EnableMessageStreamDisposal = false,
             MessageQuotas = new ODataMessageQuotas { MaxReceivedMessageSize = Int64.MaxValue },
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.OData/Abstracts/ODataServiceCollectionExtensions.cs
```diff
@@ -38,7 +38,7 @@ public static IServiceCollection AddDefaultWebApiServices(this IServiceCollectio
         // ReaderSettings and WriterSettings are registered as prototype services.
         // There will be a copy (if it is accessed) of each prototype for each request.
 #pragma warning disable CS0618 // ReadUntypedAsString is obsolete in ODL 8.
-        services.AddSingleton(new ODataMessageReaderSettings
```
habbes, line 41: I wonder why we had `AddSingleton` in the first place. Also check whether the comments above are still relevant, we no longer use "prototypes".

WanjohiSammy (approved):

PR #1346, opened 2024-11-13
Dynamic property support in filter expressions

gathogojr:
Dynamic property support in filter expressions

This pull request backports changes from the following pull requests:
- https://github.com/OData/AspNetCoreOData/pull/1330
- https://github.com/OData/AspNetCoreOData/pull/1322


src/Microsoft.AspNetCore.OData/Query/Validator/FilterQueryValidator.cs
```diff
@@ -617,6 +617,10 @@ protected virtual void ValidateCollectionNode(CollectionNode node, FilterValidat
                     ValidateCollectionPropertyAccessNode(propertyAccessNode, validatorContext);
                     break;
 
+                case QueryNodeKind.CollectionOpenPropertyAccess:
+                    // No identified validations for collection-valued open property yet
```
WanjohiSammy, line 621: When are we planning to add these validations if required

gathogojr: @WanjohiSammy If we ever identity specific validation that we can apply, then we will add. For example, I have wondering if filter restrictions can apply to dynamic properties. If they were applicable, we would plug those kind of validations here...

habbes (approved):

Issue #1327, opened 2024-10-21
Using DateOnly in a groupby with $apply results in object reference not set


computerwiz7791:
**Assemblies affected**
ASP.NET Core OData 9.x

**Describe the bug**
When querying a data model that contains a DateOnly property, where I want to group by that property to get summary or just the distinct records, I get an error saying that the object reference not set to an instance of an object.  When I change this to use DateTime the problem doesn't exist.  Currently using DateOnly as EF 8 now supports DateOnly (date database field).

**Project Code To Reproduce**
```

using Microsoft.AspNetCore.OData;
using Microsoft.AspNetCore.OData.Query;
using Microsoft.AspNetCore.OData.Routing.Controllers;
using Microsoft.OData.ModelBuilder;

var builder = WebApplication.CreateBuilder(args);

var modelBuilder = new ODataConventionModelBuilder();
var timeRecord = modelBuilder.EntitySet<TimeRecord>(nameof(TimeRecord)).EntityType;
timeRecord.Select().Expand(6).OrderBy().Filter().Count().Page();

// DI
builder.Services
    .AddControllers()
    .AddOData(options => options.EnableQueryFeatures(1000).AddRouteComponents(modelBuilder.GetEdmModel()));

// Configure
var app = builder.Build();
app.UseRouting();
app.UseEndpoints(endpoints => { endpoints.MapControllers();});
app.Run();

public class TimeRecord
{
    public long Id { get; set; }
    public DateOnly PeriodDate { get; set; }
    public decimal Hours { get; set; }
    public string Employee { get; set; }
}

public class TimeRecordController : ODataController
{
    [EnableQuery]
    public IQueryable Get()
    {
        return (new List<TimeRecord>()
        {
            new TimeRecord(){Id = 1, Employee = "John", PeriodDate = DateOnly.Parse("2024-10-01"), Hours = 35},
            new TimeRecord(){Id = 2, Employee = "John", PeriodDate = DateOnly.Parse("2024-10-08"), Hours = 15},
            new TimeRecord(){Id = 3, Employee = "Ellen", PeriodDate = DateOnly.Parse("2024-10-08"), Hours = 25},
            new TimeRecord(){Id = 4, Employee = "John", PeriodDate = DateOnly.Parse("2024-10-15"), Hours = 30},
            new TimeRecord(){Id = 5, Employee = "John", PeriodDate = DateOnly.Parse("2024-10-22"), Hours = 27},
        }).AsQueryable();
    }
}

[...]

PR #1353, opened 2024-11-15
Fix: Include DateOnly and TimeOnly in Non-standard to their standard EDM Primitive Conversions

WanjohiSammy:
### Issues

*This pull request fixes [#1327](https://github.com/OData/AspNetCoreOData/issues/1327).*

### Description

This PR included `DateOnly` and `TimeOnly` types as non-standard EDM primitive types in the `ConvertNonStandardPrimitives` method. This is to be able to convert these types to their standard EDM primitive types.

Given the following model:
```cs
public class DateOnyTimeOnlyModel
{
    public int Id { get; set; }

    public DateOnly Birthday { get; set; }

    public DateOnly EndDay { get; set; }

    public TimeOnly CreatedTime { get; set; }

    public TimeOnly? EndTime { get; set; }
}
```

With `DateOnly` and `TimeOnly` types added, the `HandleObjectType` method now supports the following:
- `$apply=groupby((EndDay))`
- `$apply=groupby((CreatedTime))`
- `$apply=groupby((EndTime),aggregate(Birthday with min as MinBirthday))`
- `?$apply=groupby((EndDay),aggregate(Id with max as MaxId))&$orderby=PublishDay`

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/Expressions/ExpressionBinderBase.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Expressions/ExpressionBinderBase.cs
@@ -1099,20 +1099,7 @@ public abstract class ExpressionBinderBase
                         break;
 
                     case TypeCode.Object:
-                        if (sourceType == typeof(char[]))
-                        {
-                            convertedExpression = Expression.New(typeof(string).GetConstructor(new[] { typeof(char[]) }), source);
-                        }
-                        else if (sourceType == typeof(XElement))
-                        {
-                            convertedExpression = Expression.Call(source, "ToString", typeArguments: null, arguments: null);
-                        }
-#if NETFX // System.Data.Linq.Binary is only supported in the AspNet version.
-                        else if (sourceType == typeof(Binary))
-                        {
-                            convertedExpression = Expression.Call(source, "ToArray", typeArguments: null, arguments: null);
-                        }
-#endif
+                        convertedExpression = HandleObjectType(sourceType, source);
                         break;
 
                     default:
@@ -1330,6 +1317,32 @@ public abstract class ExpressionBinderBase
             {
                 return NullConstant;
             }
+        }
+    }
+
+    private static Expression HandleObjectType(Type sourceType, Expression source)
+    {
+        switch (sourceType)
+        {
+            case Type t when t == typeof(char[]):
+                return Expression.New(typeof(string).GetConstructor(new[] { typeof(char[]) }), source);
+
+            case Type t when t == typeof(XElement):
+                return Expression.Call(source, "ToString", null, null);
+
+            case Type t when t == typeof(DateOnly):
+                return source;
+
+            case Type t when t == typeof(TimeOnly):
+                return source;
+
+#if NETFX
+        case Type t when t == typeof(Binary):
+            return Expression.Call(source, "ToArray", null, null);
+#endif
+
+            default:
+                throw new NotSupportedException($"Unsupported object type: {sourceType.Name}");
         }
     }
     #endregion
--- a/test/Microsoft.AspNetCore.OData.E2E.Tests/DateOnlyTimeOnly/DateAndTimeOfDayWithEfTest.cs
+++ b/test/Microsoft.AspNetCore.OData.E2E.Tests/DateOnlyTimeOnly/DateAndTimeOfDayWithEfTest.cs
@@ -236,6 +236,8 @@ public class DateOnlyAndTimeOnlyWithEfTest : WebApiTestBase<DateOnlyAndTimeOnlyW
     [InlineData("?$apply=groupby((Birthday))", 5)]
     [InlineData("?$apply=groupby((PublishDay))", 4)]
     [InlineData("?$apply=groupby((CreatedTime))", 5)]
+    [InlineData("?$apply=groupby((EndTime))", 4)]
+    [InlineData("?$apply=groupby((EndDay))", 5)]
     public async Task CanGroupBy_OnDateOnlyAndTimeOnlyProperties(string applyGroupBy, int expectCount)
     {
         // Arrange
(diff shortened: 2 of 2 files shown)


src/Microsoft.AspNetCore.OData/Query/Expressions/ExpressionBinderBase.cs
```diff
@@ -1332,5 +1319,31 @@ internal Expression BindCastToEnumType(Type sourceType, Type targetClrType, Quer
...
+
+            case Type t when t == typeof(DateOnly):
+                return source;
+
+            case Type t when t == typeof(TimeOnly):
+                return source;
+
+#if NETFX
+        case Type t when t == typeof(Binary):
+            return Expression.Call(source, "ToArray", null, null);
```
gathogojr, line 1341: nit: Indent off?

src/Microsoft.AspNetCore.OData/Query/Expressions/ExpressionBinderBase.cs
```diff
@@ -1099,20 +1099,7 @@ internal Expression ConvertNonStandardPrimitives(Expression source)
...
-                        {
-                            convertedExpression = Expression.Call(source, "ToString", typeArguments: null, arguments: null);
-                        }
-#if NETFX // System.Data.Linq.Binary is only supported in the AspNet version.
-                        else if (sourceType == typeof(Binary))
-                        {
-                            convertedExpression = Expression.Call(source, "ToArray", typeArguments: null, arguments: null);
-                        }
-#endif
+                        convertedExpression = HandleObjectType(sourceType, source);
```
gathogojr, line 1102: Curious, why d `DateOnly` and `TimeOnly` end up in the "catch-all" `TypeCode.Object` case, while `DateTime` goes to a specific `TypeCode.DateTime` case - https://github.com/OData/AspNetCoreOData/blob/9a0c7dbbfbc141f21f9bad50504f7f3978742ead/src/Microsoft.AspNetCore.OData/Query/Expressions/ExpressionBinderBase.cs#L1097

gathogojr, line 1102: A similar `ConventNonStandardPrimitives` method is in `QueryBinder` class - https://github.com/OData/AspNetCoreOData/blob/9a0c7dbbfbc141f21f9bad50504f7f3978742ead/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs#L1174

When it that method called and when is this one called?
Does that other method need the same cure you have applied here?

src/Microsoft.AspNetCore.OData/Query/Expressions/ExpressionBinderBase.cs
```diff
@@ -1332,5 +1319,31 @@ internal Expression BindCastToEnumType(Type sourceType, Type targetClrType, Quer
...
+            case Type t when t == typeof(TimeOnly):
+                return source;
+
+#if NETFX
+        case Type t when t == typeof(Binary):
+            return Expression.Call(source, "ToArray", null, null);
+#endif
+
+            default:
+                throw new NotSupportedException($"Unsupported object type: {sourceType.Name}");
```
gathogojr, line 1345: Shouldn't we be using a resource string here?

src/Microsoft.AspNetCore.OData/Query/Expressions/ExpressionBinderBase.cs
WanjohiSammy: This also made me wonder. 
[TypeCode Enum](https://learn.microsoft.com/en-us/dotnet/api/system.typecode?view=net-8.0) does not have DateOnly and TimeOnly as of yet.

You can view [typecode.cs,references](https://referencesource.microsoft.com/#mscorlib/system/typecode.cs,c60c8eedc11bf6fc,references)

WanjohiSammy: Thanks @gathogojr for pointing this out:

The check for `DateOnly` and `TimeOnly` was already added: https://github.com/OData/AspNetCoreOData/blob/9a0c7dbbfbc141f21f9bad50504f7f3978742ead/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs#L1192

Here is the PR: https://github.com/OData/AspNetCoreOData/commit/dad7c4cf04202a26e61b4e1517443ae9e81eb646#diff-d4b1c506e8df032b86d91d517b0df08ad83d090a2037bf5afa50198cda72484cL11

```cs
// rest of code
else if (TypeHelper.IsDateOnly(sourceType) || TypeHelper.IsTimeOnly(sourceType))
{
    convertedExpression = source;
}
// rest of code
```

src/Microsoft.AspNetCore.OData/Query/Expressions/ExpressionBinderBase.cs
WanjohiSammy: This has been moved to `SRResources`

habbes (approved):

src/Microsoft.AspNetCore.OData/Properties/SRResources.resx
```diff
@@ -681,6 +681,9 @@
   <data name="MissingRequiredParameterInOperation" xml:space="preserve">
     <value>Missing the required parameter '{0}' is in the operation '{1}' parameter mapping.</value>
   </data>
+  <data name="MissingNonStandardTypeSupportFor" xml:space="preserve">
+    <value>Missing Non-Standard type support for {0}.</value>
```
gathogojr, line 685:
```suggestion
    <value>Missing non-standard type support for '{0}'.</value>
```

gathogojr (dismissed):
Left a suggestion to improve a resource text otherwise LGTM

gathogojr (approved):

PR #1364, opened 2024-11-28
Enhancement/query actual type

VenkiDevelopers:


VenkiDevelopers:
Add query to the controller to get know the actual returntype of the endpoint Get.

PR #1365, opened 2024-12-03
Fix deltas

### LINKED ISSUE #1360: Address issues serializing a delta response payload from a function
**Assemblies affected**
ASP.NET Core OData 8.x

**Describe the bug**
OData has traditionally supported writing delta response payloads in two scenarios:
1) The response to a request to a delta link
2) The response to a PATCH request to a collection (aka bulk operation)
Recently, OData added a third scenario for returning a delta payload:
3) The result of a function whose returntype is annotated with the isDelta attribute.

**Reproduce steps**

Download and run the [DeltaFunctions Sample](https://github.com/OData/AspNetCoreOData/tree/DeltaFunctions/sample/DeltaFunctions)

**Data Model**

**EDM (CSDL) Model**
```xml
<edmx:Edmx xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx" Version="4.0">
    <edmx:DataServices>
        <Schema xmlns="http://docs.oasis-open.org/odata/ns/edm" Namespace="microsoft.graph">
            <EntityType Name="recoveryPreviewJob">
                <Key>
                    <PropertyRef Name="id"/>
                </Key>
                <Property Name="id" Type="Edm.String"/>
            </EntityType>
            <EntityType Name="recoveryChangeObject">
                <Key>
                    <PropertyRef Name="id"/>
                </Key>
                <Property Name="id" Type="Edm.String"/>
                <NavigationProperty Name="currentState" Type="Edm.EntityType" Nullable="false"/>
                <NavigationProperty Name="deltaFromCurrent" Type="Edm.EntityType" Nullable="false"/>
            </EntityType>
            <EntityType Name="directoryObject">
                <Key>
                    <PropertyRef Name="id"/>
                </Key>
                <Property Name="id" Type="Edm.String"/>
                <Property Name="displayName" Type="Edm.String"/>
                </EntityType>
            <EntityType Name="user" BaseType="microsoft.graph.directoryObject">
                <Property Name="emailName" Type="Edm.String"/>
            </EntityType>
            <EntityType Name="group" BaseType="microsoft.graph.directoryObject">
                <NavigationProperty Name="members" Type="Collection(microsoft.graph.directoryObject)"/>
            </EntityType>
            <Function Name="getChanges" IsBound="true" IsComposable="true">
                <Parameter Name="recoveryPreviewJob" Type="microsoft.graph.recoveryPreviewJob" Nullable="false"/>
                <ReturnType Type="Collection(microsoft.graph.recoveryChangeObject)" Nullable="false"/>
            </Function>
            <EntityContainer Name="graphService">
                <EntitySet Name="recoveryPreviewJobs" EntityType="microsoft.graph.recoveryPreviewJob"/>
                <EntitySet Name="recoveryChangeObjects" EntityType="microsoft.graph.recoveryChangeObject"/>
            </EntityContainer>
        </Schema>
    </edmx:DataServices>
</edmx:Edmx>
```

**Request/Response**
GET: http://[localhost:5219/recoveryPreviewJobs/1/getChanges()](http://localhost:5219/recoveryPreviewJobs/1/getChanges())

```json
{
    "@context": "http://localhost:5219/$metadata#recoveryChangeObjects/$delta",
    "value": [
        {
            "id": null,
            "currentState": {
                "@type": "#microsoft.graph.user",
                "id": "user1",
                "displayName": "William",
                "emailName": null   <=email should not be written; it was not set in the delta change object
            },
            "deltaFromCurrent": {
                "@type": "#microsoft.graph.user",
                "id": "user1",
                "displayName": "Bill",
                "emailName": null    <= email should not be written; it was not set in the delta change object
            }
        },
        {
            "id": "2",
            "currentState": {
                "@type": "#microsoft.graph.group",
                "id": "group1",
                "displayName": null,
                "members@delta": [                    <=this should be "members", not "members@delta -- it is the full collection
                    {
                        "@type": "#microsoft.graph.user",
                        "id": "user7",
                        "displayName": null,            <= displayName was not set and should not be written
                        "emailName": null                <=emailName was not set and should not be written
                    }
                ]
            },
            "deltaFromCurrent": {
                "@type": "#microsoft.graph.group",
                "id": "group1",
                "displayName": null,
                "members@delta": [
                    {
                        "@type": "#microsoft.graph.user",
                        "id": "user3",
                        "displayName": null,            <=displayName was not set and should not be written
                        "emailName": null                <=emailName was not set and should not be written
                    },
                    {
                        "@removed": {
                            "reason": "deleted"
                        },
                        "@id": "https://graph.microsoft.com/v1.0/users/4"    <= the deleted resource set other fields, like id, that should have been written
                    }
                ]
            }
        }
    ]
}
```
**Expected behavior**

1) Only properties of the change objects that were set should be written
2) Nested ODataResourceSets within the delta payload should be serialized using ODataDeltaResourceSetSerializer, not DeltaResourceSetSerializer 
3) Properties of deleted resources should be serialized
4) Other annotations, etc. of deleted and changed resources should be written just as would be for a regular resource (not shown)
 
```json
{
    "@context": "http://localhost:5219/$metadata#recoveryChangeObjects/$delta",
    "value": [
        {
            "id": null,
            "currentState": {
                "@type": "#microsoft.graph.user",
                "id": "user1",
                "displayName": "William"
            },
            "deltaFromCurrent": {
                "@type": "#microsoft.graph.user",
                "id": "user1",
                "displayName": "Bill"
            }
        },
        {
            "id": "2",
            "currentState": {
                "@type": "#microsoft.graph.group",
                "id": "group1",
                "members": [
                    {
                        "@type": "#microsoft.graph.user",
                        "id": "user7"
                    }
                ]
            },
            "deltaFromCurrent": {
                "@type": "#microsoft.graph.group",
                "id": "group1",
                "members@delta": [
                    {
                        "@type": "#microsoft.graph.user",
                        "id": "user3"
                    },
                    {
                        "@removed": {
                            "reason": "deleted"
                        },
                        "@type": "#microsoft.graph.user",
                        "@id": "https://graph.microsoft.com/v1.0/users/4",
                        "id": "user4"
                    }
                ]
            }
        }
    ]
}
```
**Additional context**
Attempting to implement the third pattern revealed some issues with how we serialize a delta payload. Many of these are there due to how the code evolved; originally delta payloads were completely separate from regular payloads, and deleted resources were completely different than regular resources. However, the 4.01 JSON delta format supported in 1), and exclusively in 2) and 3) makes delta payloads and regular payloads more similar than different, and deleted resources the same as regular resources with an additional removed control information at the beginning (and the requirement to include key fields or @id).

For these reasons, we should refactor the delta serializers to be more fully functioned and to share implementation with the non-delta serializers.

mikepizzo:
Fixes #1360


src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataDeletedResourceSerializer.cs
```diff
@@ -0,0 +1,240 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.OData.Deltas;
+
+namespace Microsoft.AspNetCore.OData.Formatter.Serialization;
+
+/// <summary>
+/// ODataSerializer for serializing instances of <see cref="IEdmDeltaDeletedResourceObject"/>/>
+/// </summary>
+[SuppressMessage("Microsoft.Maintainability", "CA1506:AvoidExcessiveClassCoupling", Justification = "Relies on many ODataLib classes.")]
+public class ODataDeletedResourceSerializer : ODataEdmTypeSerializer
```
xuzhg, line 25: DeletedResource typically is a 'Resource'?
1. Will we have other type of Resource?
2. Will we create corresponding Serializer also if #1 yes?
3. Why cannot use the ODataResourceSerializer?
4. Can we derive it from ODataResourceSerializer, not from ODataEdmTypeSerializer?

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataDeletedResourceSerializer.cs
```diff
@@ -0,0 +1,240 @@
...
+            await WriteDeletedResourceAsync(graph, writer, writeContext, expectedType).ConfigureAwait(false);
+        }
+    }
+
+    private async Task WriteDeletedResourceAsync(object graph, ODataWriter writer, ODataSerializerContext writeContext,
+    IEdmTypeReference expectedType)
+    {
+        Contract.Assert(writeContext != null);
+
+        //TODO: do we need this?
```
xuzhg, line 88: I don't think so if it's for DeletedResource only?

test/Microsoft.AspNetCore.OData.E2E.Tests/BulkOperation/BulkOperationTest.cs
```diff
@@ -75,7 +75,7 @@ public async Task DeltaSet_WithDeletedAndODataId_IsSerializedSuccessfully()
                 {'ID':2,'Name':'Employee2','Friends@odata.delta':[{'@id':'Friends(1)'}]}
             ]}";
 
-        string expectedResponse = "{\"@context\":\"http://localhost/convention/$metadata#Employees/$delta\",\"value\":[{\"ID\":1,\"Name\":\"Employee1\",\"Friends@delta\":[{\"@removed\":{\"reason\":\"changed\"},\"@id\":\"http://host/service/Friends(1)\"}]},{\"ID\":2,\"Name\":\"Employee2\",\"Friends@delta\":[{\"Id\":1}]}]}";
+        string expectedResponse = "{\"@context\":\"http://localhost/convention/$metadata#Employees/$delta\",\"value\":[{\"ID\":1,\"Name\":\"Employee1\",\"Friends@delta\":[{\"@removed\":{\"reason\":\"changed\"},\"@id\":\"http://host/service/Friends(1)\",\"id\":1}]},{\"ID\":2,\"Name\":\"Employee2\",\"Friends@delta\":[{\"Id\":1}]}]}";
```
xuzhg, line 78: why dose the expected response look different? and since it contains @id, why does it contain "id" again?

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -235,8 +218,28 @@ await writer.WriteStartAsync(new ODataResourceSet
...
+                    EdmCollectionTypeReference edmType = new EdmCollectionTypeReference(new EdmDeltaCollectionType(new EdmEntityTypeReference(itemType, false)));
+                    await serializer.WriteObjectInlineAsync(
+                        propertyValue,
+                        edmType,
+                        writer,
+                        nestedWriteContext).ConfigureAwait(false);
+                }
+                else
+                {
+                    ODataEdmTypeSerializer serializer = new ODataResourceSetSerializer(SerializerProvider);
```
xuzhg, line 240: retrieve the ODataResourceSetSerializer from DI?

test/Microsoft.AspNetCore.OData.E2E.Tests/BulkOperation/BulkOperationTest.cs
mikepizzo: Previously, we weren't returning properties of a deleted resource even if they were set by the service, which is why id was not being returned (which was a bug).

In this case, both the @id and the id were explicitly set in the controller (they were both set in the request payload which is echoed in the response payload), so they should both be returned in the response.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataDeletedResourceSerializer.cs
mikepizzo: I don't anticipate adding other types of resources. In ODL, we have a hierachy of two -- ODataResource and ODataDeletedResource both derive from ODataResourceBase, just as ODataResourceSet and ODataDeltaResourceSet derive from ODataResourceSetBase (and have separate serializers).

Deriving ODataDeletedResourceSerializer from ODataResourceSerializer could be interesting; then we could make some of the functions protected internal, rather than internal static. Let me play around with that...

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
mikepizzo: :thumbsup:

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataDeletedResourceSerializer.cs
mikepizzo: Updated to have ODataDeletedResourceSerializer derive from ODataResourceSerializer -- thanks for the suggestion!

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataDeletedResourceSerializer.cs
mikepizzo: thanks for the confirmation. removed dead code.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
mikepizzo: I just followed what the code was already doing for ODataDeltaResourceSetSerializer, but there was good reason for directly creating there, and not for the ODataResourceSetSerializer, so I updated the code -- thanks!

src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatterHelper.cs
```diff
@@ -89,8 +89,14 @@ internal static async Task WriteToStreamAsync(
         writerSettings.BaseUri = baseAddress;
 
         //use v401 to write delta payloads. 
-        if (serializer.ODataPayloadKind == ODataPayloadKind.Delta)
+        if (serializer.ODataPayloadKind == ODataPayloadKind.Delta && version < ODataVersion.V401)
         {
+            // Preserve setting of OmitODataPrefix
+            if (writerSettings.Version.GetValueOrDefault() == ODataVersion.V4)
+            {
+                writerSettings.SetOmitODataPrefix(writerSettings.GetOmitODataPrefix(ODataVersion.V4), ODataVersion.V401);
```
xuzhg, line 97: why don't we enable it using configuration?  not hard-coded? 
By default, it's ok to enable the configuration, but at least , maybe we need a way to let customer change the default?

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataDeletedResourceSerializer.cs
```diff
@@ -0,0 +1,173 @@
...
+
+            await writer.WriteStartAsync(odataDeletedResource).ConfigureAwait(false);
+            ODataResourceSerializer serializer = SerializerProvider.GetEdmTypeSerializer(expectedType) as ODataResourceSerializer;
+            if (serializer == null)
+            {
+                throw new SerializationException(
+                    Error.Format(SRResources.TypeCannotBeSerialized, expectedType.ToTraceString()));
+            }
+            await serializer.WriteResourceContent(writer, selectExpandNode, resourceContext, /*isDelta*/ true);
+            await writer.WriteEndAsync().ConfigureAwait(false);
```
xuzhg, line 126: shall we call 'ConfigureAwait' for all await?

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataDeltaResourceSetSerializer.cs
```diff
@@ -414,4 +442,11 @@ private static IEdmStructuredTypeReference GetResourceType(IEdmTypeReference fee
         string message = Error.Format(SRResources.CannotWriteType, typeof(ODataDeltaResourceSetSerializer).Name, feedType.FullName());
         throw new SerializationException(message);
     }
+
+    private bool WriteDeltaDeletedResourceAsyncIsOverridden()
+    {
+        MethodInfo method = GetType().GetMethod("WriteDeltaDeletedResourceAsync", new Type[] { typeof(object), typeof(ODataWriter), typeof(ODataSerializerContext) });
+        Contract.Assert(method != null, "WriteDeltaDeletedResourceAsync is not defined.");
```
xuzhg, line 449: 1. can we leave a details comment for this method?
2. Moreover, what about if customer creates/customize his own ODataDeletedResourceSerilizer?

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataDeletedResourceSerializer.cs
```diff
@@ -0,0 +1,173 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.OData.Deltas;
+
+namespace Microsoft.AspNetCore.OData.Formatter.Serialization;
+
+/// <summary>
+/// ODataSerializer for serializing instances of <see cref="IEdmDeltaDeletedResourceObject"/>/>
+/// </summary>
+[SuppressMessage("Microsoft.Maintainability", "CA1506:AvoidExcessiveClassCoupling", Justification = "Relies on many ODataLib classes.")]
+public class ODataDeletedResourceSerializer : ODataResourceSerializer
```
xuzhg, line 25: I didn't see to register this serializer into the DI? Will you do it later or miss that?

xuzhg (approved):
I leave couple comments, but overall, it looks good to me except the DI related.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataDeletedResourceSerializer.cs
mikepizzo: The regular method of using the EdmTypeSerializer doesn't work because it doesn't differentiate between writing a resource and writing a deleted resource.  So, I added code to the only place the DeletedResourceSerializer is created, which is from the DeltaResourceSetSerializer, to use DI, and registered the ODataDeletedResourceSerailizer in AddDefaultApiServices.  Note that DeltaResourceSetSerilizer falls back to directly creating the ODataDeletedResoruceSerializer if GetService returns null, for example, if the service isn't using our AddDefaultWebApiServices.

src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatterHelper.cs
mikepizzo: This code just fixes a bug where we weren't honoring the writing of the odata prefix in cases where we did change the version. I'd like to separately track if the user can override serializing deltas as 4.01, as this is existing behavior, and changing that behavior is outside of the scope of this PR.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataDeltaResourceSetSerializer.cs
mikepizzo: 1) Added comments
2) This is only here for backward compatibility. If the user customizes their own ODataDeletedResourceSerializer, they can provide whatever custom logic they need.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataDeletedResourceSerializer.cs
mikepizzo: We currently always use ConfigureAwait(false) to improve perf and avoid possible deadlocks because we are a library and not an app.

xuzhg (approved):

Issue #1359, opened 2024-11-25
Default ODataEnumSerializer and ODataEnumDeserializer Fails to Convert Multi-Value Flag Enum to Lower Camel Case


WanjohiSammy:
**Assemblies affected**
8.x

**Describe the bug**
The `ODataEnumSerializer` and `ODataEnumDeserializer` fails to convert multi-value Flag enums to `LowerCamelCase` as specified in `EnumMember attribute`. However, this works with single-value Flag enums.

### Example:

Consider the following Flag enum:
```cs
[Flags]
[DataContract(Name = "exampleEnum")]
public enum ExampleEnum
{
    [EnumMember(Value = "firstValue")]
    FirstValue = 1,

    [EnumMember(Value = "secondValue")]
    SecondValue = 2,

    [EnumMember(Value = "thirdValue")]
    ThirdValue = 4
}
```
With the default ODataEnumSerializer, a `single-value` Flag enum like `ExampleEnum.FirstValue` is correctly converted to ("`firstValue`"). However, a `multi-value` Flag enum like `ExampleEnum.FirstValue | ExampleEnum.SecondValue` is not converted to ("`firstValue, secondValue`") as expected. Instead, it remains in its original form as `"FirstValue, SecondValue"`

**Reproduce steps**
1. Clone the repo
    ```bash
    git clone https://github.com/WanjohiSammy/ODataWebApiApplicationSample.git
    ```
3. Build and run the application
4. Find the `ODataWebApiApplication15.http` and execute the following requests:

    - `GET /odata/Customers`: 
       Note that the `customerType` for customer `id=1` is "**regular**" but the `customerType` for customer `id=2` is "**Premium, VIP**" instead of "**premium, vip**" in `lower camel case`
    ```json
      {
        "@odata.context": "https://localhost:7085/odata/$metadata#Customers",
        "value": [
          {
            "id": 1,
            "name": "Customer1",
            "customerType": "regular"
          },
          {
            "id": 2,
            "name": "Customer2",
            "customerType": "Premium, VIP"
          },
          ...
        ]
      }
    ```

   - `POST /odata/Customers`
      - With `"customerType": "Regular, Premium"` - a new record is created successfully
      
      - With lower camel case `"customerType": "regular, premium"` - request is null
        ```json
          {
             "Id": 0,
             "Name": "John Doe",
             "CustomerType": "regular, premium"
          }
        ```

      - With lower camel case single-value `"customerType": "regular"` - works as expected
[...]

PR #1367, opened 2024-12-04
Fix Default ODataEnumSerializer and ODataEnumDeserializer Fails to Convert Multi-Value Flag Enum Values to Lower Camel Case

WanjohiSammy:
### Issues

*This pull request fixes [#1359](https://github.com/OData/AspNetCoreOData/issues/1359).*

### Description

Currently, the `ODataEnumSerializer` and `ODataEnumDeserializer` fails to convert multi-value Flag enums to LowerCamelCase as specified in `EnumMember` attribute. However, this works with single-value Flag enums.

### Example:

Consider the following Flag enum:
```cs
[Flags]
[DataContract(Name = "exampleEnum")]
public enum ExampleEnum
{
    [EnumMember(Value = "firstValue")]
    FirstValue = 1,

    [EnumMember(Value = "secondValue")]
    SecondValue = 2,

    [EnumMember(Value = "thirdValue")]
    ThirdValue = 4
}
```
With the default ODataEnumSerializer, a `single-value` Flag enum like `ExampleEnum.FirstValue` is correctly converted to ("`firstValue`"). However, a `multi-value` Flag enum like `ExampleEnum.FirstValue | ExampleEnum.SecondValue` is not converted to ("`firstValue, secondValue`") as expected. Instead, it remains in its original form as `"FirstValue, SecondValue"`

This change modifies both `ODataEnumSerializer` and `ODataEnumDeserializer` to correctly convert multi-value Flag enums to LowerCamelCase. This is to ensure that the `serialization` and `deserialization` of multi-value Flag enums are consistent with the single-value Flag enums.

With this change, the following are supported by default without having to write custom converters:
- `GET odata/examples`:
   ```json
  {
    "@odata.context": "https://localhost:7085/odata/$metadata#Examples",
    "value": [
      {
        "id": 1,
        "name": "Example1",
        "ExampleEnum": "firstValue"
      },
      {
        "id": 2,
        "name": "Example2",
        "ExampleEnum": "firstValue, secondValue" // support write-to multi-value Flags enums to Lower Camel Case
      },
      ...
    ]
  }
   ```

- `POST odata/examples`:
   ```json
  {
     "Id": 0,
     "Name": "John Doe Example",
     "ExampleEnum": "firstValue, thirdValue" // support read-in multi-value Flags enums in Lower Camel Case
  }
   ```

- `PATCH odata/examples(1)`:
   ```json
  {
       "ExampleEnum": "firstValue, thirdValue" // support read-in multi-value Flags enums in Lower Camel Case
  }
   ```

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataEnumDeserializer.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataEnumDeserializer.cs
@@ -6,6 +6,7 @@
 //------------------------------------------------------------------------------
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics.Contracts;
 using System.Linq;
 using System.Threading.Tasks;
@@ -128,41 +129,49 @@ public class ODataEnumDeserializer : ODataEdmTypeDeserializer
     {
         long result = 0;
         Type clrEnumType = TypeHelper.GetUnderlyingTypeOrSelf(clrType);
+        ReadOnlySpan<char> source = enumValue.Value.AsSpan().Trim();
 
-        // use `stackalloc` to allocate a Span<Range> on the stack, which avoids heap allocations.
-        Span<Range> ranges = stackalloc Range[enumValue.Value.Count(c => c == ',') + 1];
-        ReadOnlySpan<char> source = enumValue.Value.AsSpan();
-
-        // For flags enum, we need to split the value and convert it to the enum value.
-        int count = source.Split(ranges, ',', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
-        for(int i = 0; i < count; i++)
+        int start = 0;
+        while (start < source.Length)
         {
-            ReadOnlySpan<char> value = source[ranges[i]];
-
-            // Try to find the enum member in the EDM model.
-            IEdmEnumMember enumMember = null;
-            foreach (IEdmEnumMember member in enumType.Members)
+            // Find the end of the current value.
+            int end = start;
+            while (end < source.Length && source[end] != ',')
             {
-                if (value.Equals(member.Name, StringComparison.InvariantCultureIgnoreCase))
-                {
-                    enumMember = member;
-                    break;
-                }
+                end++;
             }
 
-            if (enumMember != null)
-            {
-                Enum clrEnumMember = memberMapAnnotation.GetClrEnumMember(enumMember);
-                result |= Convert.ToInt64(clrEnumMember);
-            }
-            else if (Enum.TryParse(clrEnumType, value, true, out var enumMemberParsed))
+            // Extract the current value.
+            ReadOnlySpan<char> currentValue = source[start..end].Trim();
+
+            bool parsed = Enum.TryParse(clrEnumType, currentValue, true, out var enumMemberParsed);
+            if (parsed)
             {
                 result |= Convert.ToInt64((Enum)enumMemberParsed);
             }
             else
             {
+                foreach (IEdmEnumMember enumMember in enumType.Members)
+                {
+                    // Check if the current value matches the enum member name.
+                    parsed = currentValue.Equals(enumMember.Name.AsSpan(), StringComparison.InvariantCultureIgnoreCase);
+                    if (parsed)
+                    {
+                        Enum clrEnumMember = memberMapAnnotation.GetClrEnumMember(enumMember);
+                        result |= Convert.ToInt64(clrEnumMember);
+                        break;
+                    }
+                }
+            }
+
+            // If still not valid, return null.
+            if (!parsed)
+            {
                 return null;
             }
+
+            // Move to the next value.
+            start = end + 1;
         }
 
         return result == 0 ? null : Enum.ToObject(clrEnumType, result);
--- a/src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataEnumDeserializer.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataEnumDeserializer.cs
@@ -13,6 +13,7 @@ using Microsoft.AspNetCore.OData.Edm;
 using Microsoft.AspNetCore.OData.Formatter.Value;
 using Microsoft.OData;
 using Microsoft.OData.Edm;
+using Microsoft.OData.ModelBuilder;
 
 namespace Microsoft.AspNetCore.OData.Formatter.Deserialization;
 
@@ -89,7 +90,7 @@ public class ODataEnumDeserializer : ODataEdmTypeDeserializer
         {
             if (enumValue != null)
             {
-                IEdmEnumMember enumMember = enumType.Members.FirstOrDefault(m => m.Name == enumValue.Value);
+                IEdmEnumMember enumMember = enumType.Members.FirstOrDefault(m => m.Name.Equals(enumValue.Value, StringComparison.InvariantCultureIgnoreCase));
                 if (enumMember != null)
                 {
                     var clrMember = memberMapAnnotation.GetClrEnumMember(enumMember);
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataEnumDeserializer.cs
```diff
@@ -98,10 +110,66 @@ public override object ReadInline(object item, IEdmTypeReference edmType, ODataD
...
+                clrEnumMember = (Enum)enumMemberParsed;
+            }
+            else
+            {
+                return null;
+            }
+
+            if (clrEnumMember != null)
+            {
+                result |= Convert.ToInt64(clrEnumMember);
```
derobuts, line 163: is 0 not a valid value in an enum in Odata or do all enum indices start at 1?

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataEnumDeserializer.cs
```diff
@@ -81,15 +84,24 @@ public override object ReadInline(object item, IEdmTypeReference edmType, ODataD
             return new EdmEnumObject(enumTypeReference, enumValue.Value);
         }
 
+        bool enablePropertyNameCaseInsensitive = 
+            readContext.Request != null &&
+            readContext.Request.ODataOptions() != null &&
+            readContext.Request.ODataOptions().RouteOptions != null &&
+            readContext.Request.ODataOptions().RouteOptions.EnablePropertyNameCaseInsensitive;
+
```
xuzhg, line 92: EnablePropertyNameCaseInsensitive is for the Route options. I don't think it's for the serialization/deserialization.

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataEnumDeserializer.cs
WanjohiSammy: @derobuts Not really. This is for enum with `Flags` attribute, where the constants should be power of 2 to avoid overlap for multi-value. The constant values can be 1,2,4,8,16 and so on.

You can fins more detail here:
https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-flagsattribute#attributes-of-the-flagsattribute

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataEnumDeserializer.cs
```diff
@@ -98,10 +110,66 @@ public override object ReadInline(object item, IEdmTypeReference edmType, ODataD
...
+    /// <param name="enablePropertyNameCaseInsensitive">The value indicating whether to enable case insensitive for the property name in conventional routing.</param>
+    /// <param name="memberMapAnnotation">The annotation containing the mapping of CLR enum members to EDM enum members.</param>
+    /// <returns>The deserialized flags enum value.</returns>
+    private object ReadFlagsEnumValue(ODataEnumValue enumValue, IEdmEnumType enumType, Type clrType, bool enablePropertyNameCaseInsensitive, ClrEnumMemberAnnotation memberMapAnnotation)
+    {
+        long result = 0;
+        Type clrEnumType = TypeHelper.GetUnderlyingTypeOrSelf(clrType);
+
+        // For flags enum, we need to split the value and convert it to the enum value.
+        string[] values = enumValue.Value.Split(',');
```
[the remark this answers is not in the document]
WanjohiSammy: Thanks for this. I have refactored the code like so:
```cs
// use `stackalloc` to allocate a Span<Range> on the stack, which avoids heap allocations.
Span<Range> ranges = stackalloc Range[enumValue.Value.Count(c => c == ',') + 1];
ReadOnlySpan<char> source = enumValue.Value.AsSpan();

// For flags enum, we need to split the value and convert it to the enum value.
int count = source.Split(ranges, ',', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
```

Doc: https://learn.microsoft.com/en-us/dotnet/api/system.memoryextensions.split?view=net-8.0

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataEnumDeserializer.cs
WanjohiSammy: I have removed the `EnablePropertyNameCaseInsensitive` and set the comparison to be `StringComparison.InvariantCultureIgnoreCase` by default.

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataEnumDeserializer.cs
```diff
@@ -98,10 +102,69 @@ public override object ReadInline(object item, IEdmTypeReference edmType, ODataD
...
+    /// <param name="clrType">The EDM enum CLR type.</param>
+    /// <param name="memberMapAnnotation">The annotation containing the mapping of CLR enum members to EDM enum members.</param>
+    /// <returns>The deserialized flags enum value.</returns>
+    private static object ReadFlagsEnumValue(ODataEnumValue enumValue, IEdmEnumType enumType, Type clrType, ClrEnumMemberAnnotation memberMapAnnotation)
+    {
+        long result = 0;
+        Type clrEnumType = TypeHelper.GetUnderlyingTypeOrSelf(clrType);
+
+        // use `stackalloc` to allocate a Span<Range> on the stack, which avoids heap allocations.
+        Span<Range> ranges = stackalloc Range[enumValue.Value.Count(c => c == ',') + 1];
```
habbes, line 133: It's dangerous to use a dynamic size for stackalloc. The size of the array here depends on the enumValue.Value.Count, which depends on user provided data. This could be arbitrarily large and blow up the stack.

If using stackalloc, always provide a fixed upper bound beyond which you fall back to a heap allocated array, or rent an array from a pool.

Here's a common pattern for using stackalloc:


```c#
const int StackAllocThreshold = 256;
Range[] array = null;
int size = enumValue.Value.Count(c => c == ',') + 1.
Span<Range> ranges = size <= StackAllocThreshold ? stackalloc Range[size]
     : array = ArrayPool<Range>.Shared.Rent(size);
ranges = ranges.Slice(0, size); // it's possible that larger array was rented than what was requested


// at the end
if (array != null)
{
    ArrayPool<Range>.Shared.Return(array);
}
```

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataEnumDeserializer.cs
```diff
@@ -98,10 +102,69 @@ public override object ReadInline(object item, IEdmTypeReference edmType, ODataD
...
+        ReadOnlySpan<char> source = enumValue.Value.AsSpan();
+
+        // For flags enum, we need to split the value and convert it to the enum value.
+        int count = source.Split(ranges, ',', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
```
habbes, line 137: You traverse the list of enum twice so far. Do we need to do that? Can wo go through the enum value once and split the strings as we go?

src/Microsoft.AspNetCore.OData/Formatter/Deserialization/ODataEnumDeserializer.cs
```diff
@@ -98,10 +102,69 @@ public override object ReadInline(object item, IEdmTypeReference edmType, ODataD
...
+        for(int i = 0; i < count; i++)
```
habbes, line 138: Maybe we could consider having a single loop that also handle the split of the strings, this also avoids the need to allocate the ranges array:

```c#
ReadOnlySpan<char> source = enumValue.AsSpan();
int currentRangeStart = 0;
for (int i = 0; i , source.Length; i++)
{
    if (source[i] != ',')
    {
       continue;
    }
    ReadOnlySpan<char> rawValue = source[currentRangeStart..i];
    ReadOnlySpan<char> value = rawValue.Trim();
    currentRangeStart = i;
   // handle value
}

// handle the last value
ReadOnlySpan<char> rawValue = source[currentRangeStart..];
ReadOnlySpan<char> value = rawValue.Trim();
```
PS: I didn't test it, verify for correctness. But the idea is we can get the ranges and the value in single scan. The Trim() is probably unnecessary if we check for spaces as well, but didn't want to complicate it too match especially since the Trim() is non-allocating.

WanjohiSammy: Refer to: https://github.com/OData/AspNetCoreOData/pull/1367#discussion_r1883706158

xuzhg (approved):

PR #1090, opened 2023-11-11
Custom method can now have nullable parameters and nullable arguments

bjelbo:
See issue #1084

PR #1374, opened 2024-12-16
Custom method can now have nullable parameters and nullable arguments for 9.x

bjelbo:
Fixes https://github.com/OData/AspNetCoreOData/issues/1084

This PR will make it possible to create custom function for nullable types.

See PR #1090 with similar change for 8.x

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.SingleValueFunctionCall.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.SingleValueFunctionCall.cs
@@ -754,12 +754,12 @@ public abstract partial class QueryBinder
         Expression[] arguments = BindArguments(node.Parameters, context);
         IEnumerable<Type> methodArgumentsType = arguments.Select(argument => argument.Type);
 
-        // Search for custom method info that are binded to the node name
-        MethodInfo methodInfo;
-        if (UriFunctionsBinder.TryGetMethodInfo(node.Name, methodArgumentsType, out methodInfo))
-        {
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.SingleValueFunctionCall.cs
```diff
@@ -754,12 +754,12 @@ protected virtual Expression BindCustomMethodExpressionOrNull(SingleValueFunctio
         Expression[] arguments = BindArguments(node.Parameters, context);
         IEnumerable<Type> methodArgumentsType = arguments.Select(argument => argument.Type);
 
-        // Search for custom method info that are binded to the node name
-        MethodInfo methodInfo;
-        if (UriFunctionsBinder.TryGetMethodInfo(node.Name, methodArgumentsType, out methodInfo))
-        {
-            return ExpressionBinderHelper.MakeFunctionCall(methodInfo, context.QuerySettings, arguments);
-        }
+            // Search for custom method info that are binded to the node name
```
xuzhg, line 757: alignment

xuzhg (dismissed):

src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.SingleValueFunctionCall.cs
bjelbo: @xuzhg code is now aligned. Also the AspNetCoreOData-main-rolling-1ES is not being automatically triggered

PR #1383, opened 2025-01-06
Make ICountOptionCollection public. Fixes #1382.

uffelauesen:

PR #1391, opened 2025-01-13
Encode segments. Fixes #1390

### LINKED ISSUE #1390: ODataPathSegmentHandler encoding of string values does not allign with ODataUriParser
**Assemblies affected**
ASP.NET Core OData 8.x++

**Describe the bug**
String key values end up unencoded encoded in odata.id, odata.readLink, odata.editLink as well as Location header. Client side on OData Client library you can end up with two representations of the same oblect in DataServiceContext's Entities tracked collection as the OData ID get constructed differently.

**Reproduce steps**
Given an entity type with a string key property ID. Create an object of that type with ID = "'24 25/&Foo,?\"".
Then do a GET request with OData Client for that object with and without selecting the ID, like...

```
var result1 = new DataServiceCollection<MyEntity>(
    from e in ctx.MyEntities 
    where e.ID == "'24 25/&Foo,?\"" 
    select new MyEntity { Title = e.Title } // AspNetCoreOData will generate and send the odata.id
);
var result2 = new DataServiceCollection<MyEntity>(
    from e in ctx.MyEntities 
    where e.ID == "'24 25/&Foo,?\"" 
    select new MyEntity { ID = e.ID } // OData Lib will generate the odata.id from the returned ID attribute
);

Assert.AreEqual(1, result1.Count);
Assert.AreEqual(1, result2.Count);
Assert.AreEqual(1, ctx.Entities.Count); // <- Exception here!!! ctx.Entities.Count is 2
Assert.AreEqual(result1.First(), result2.First());
```

**Data Model**
Given an entity type with a string key property ID. Create an object of that type with ID = "'24 25/&Foo,?\"".

**Expected behavior**
OData Clinet and AspNetCoreOData will generate the same Uri's.

**Additional context**
I originally reported this bug under OData Lib, but it is actually a bug in AspNetODataCore. You can find the original report as #2857 https://github.com/OData/odata.net/issues/2857

uffelauesen:


WanjohiSammy:
@uffelauesen 
Kindly add some description and the issue that this PR is meant to resolve.

This change seems to resolve the issue: [#1390](https://github.com/OData/AspNetCoreOData/issues/1390)

WanjohiSammy:
@uffelauesen 
What is the different between this PR and [PR #1383](https://github.com/OData/AspNetCoreOData/pull/1383)

uffelauesen:
Hi @WanjohiSammy 
Yes - I can see that I have just pushed changes for both issue to the same branch (my main branch) instead of new branches for the issues they fixes. Thats why they look the same as they actually are pointing to the same branch.
I can redo that and create different branches for the 2 issues if you need that?
Thanks
Uffe

uffelauesen:
@WanjohiSammy I had a mess in my branch structure. I am going to close this one and create a new seperate PR for this - hopefully Monday.
Thanks

uffelauesen:
@WanjohiSammy 
I have created a new PR for this. Please see #1396 
Thanks

PR #1393, opened 2025-01-17
Checking ITruncatedCollecton.IsTruncated after enumeration. Fixes #1384

### LINKED ISSUE #1384: NextLinks (paging) of expanded child collections are resolved before enumeration
**Assemblies affected**
ASP.NET Core OData 8.x

**Describe the bug**
The NextLinkGenerator to use for nested resource sets are resolved before we actually know if there are any more resources (that is before enumeration of the resources). The child resource set is checked for implementing ITruncatedCollection but also if it IsTruncated before it has been enumerated. In my provider I do not know if it has acually been truncated until after enumeration.

**Reproduce steps**
Implement ITruncatedCollection for some child collection property and observe that the ODataResourceSetSerializer will call it's IsTruncated before it has been enumerated. 

**Data Model**
NA

**EDM (CSDL) Model**
NA

**Request/Response**
NA

**Expected behavior**
The IsTruncated is called after enumeration of the child collection when the nextLink is about to be written.

**Screenshots**
NA

**Additional context**
I am trying to avoid having to override the default ODataResourceSetSerializer to get this to work as expected. The overriding of the serializer contains alot of replicated code that makes it hard to maintain as we need to revisit whenever the standard serializer changes (and it did just that recently). See also #1382

uffelauesen:
This PR fixes #1384 

The reading of ITruncatedCollection.IsTruncated (used by ODataResourceSetSerializer for nested resource sets) is postponed until after enumeration have completed.

This change is to support other implementations of ITruncatedCollection than the List<T> based TruncatedCollection<T>. Other implementations might not have the collection in memory when before enumeration starts and thus might not know if the data has been truncated like the List<T> based one.

Some tests have been added - both my attempt to write a unittest with Mocks and an E2E test.

### DIFF
--- a/test/Microsoft.AspNetCore.OData.E2E.Tests/ServerSidePaging/ServerSidePagingControllers.cs
+++ b/test/Microsoft.AspNetCore.OData.E2E.Tests/ServerSidePaging/ServerSidePagingControllers.cs
@@ -9,8 +9,11 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.OData.Extensions;
 using Microsoft.AspNetCore.OData.Query;
 using Microsoft.AspNetCore.OData.Routing.Controllers;
+using Microsoft.OData.Edm;
+using Microsoft.OData.UriParser;
 
 namespace Microsoft.AspNetCore.OData.E2E.Tests.ServerSidePaging;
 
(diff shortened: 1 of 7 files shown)


WanjohiSammy (approved):

rpallares:
Hi,

Is that MR (and a few other you propose that looks related) are in definitive to locally fix the server-driven paging feature that drastically reduce performances ?

See also https://github.com/OData/AspNetCoreOData/pull/1271#issuecomment-2209499421

I believe your recent PR they could help a lot for that topic.

uffelauesen:
Hi @WanjohiSammy 
I noticed that the pipeline failed as my added test failed.
I have fixed the test so I would expect the pipeline to be all green. 
Please have a look again.
Thanks

gathogojr (dismissed):

test/Microsoft.AspNetCore.OData.E2E.Tests/ServerSidePaging/ServerSidePagingControllers.cs
```diff
@@ -300,9 +359,9 @@ public class CollectionPagingCustomersController : ODataController
             Tags = new List<string> { "Tier 1", "Gen-Z", "HNW" },
             Categories = new List<CollectionPagingCategory>
             {
-                CollectionPagingCategory.Retailer,
-                CollectionPagingCategory.Wholesaler,
-                CollectionPagingCategory.Distributor
+            CollectionPagingCategory.Retailer,
+            CollectionPagingCategory.Wholesaler,
+            CollectionPagingCategory.Distributor
```
WanjohiSammy, line 364: nit:
Should be a tab inside

WanjohiSammy (dismissed):

WanjohiSammy (dismissed):

gathogojr (dismissed):

gathogojr (approved):

WanjohiSammy (approved):

PR #1396, opened 2025-01-20
Fixes #1390: Ensuring Url safe string key values. Aligning with ODL Client.

### LINKED ISSUE #1390: ODataPathSegmentHandler encoding of string values does not allign with ODataUriParser

uffelauesen:
ODataPathSegmentHandler is used to construct odata.id, odata.editLink, odata.readLink as well as Location header.
This PR ensures that key values are escaped for all unsafe url chars.
Uri.EscapeDataString is used to escape key values to allign with the escaping used by ODL Client.
This PR fixes issue #1390.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Routing/ODataPathSegmentHandler.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/ODataPathSegmentHandler.cs
@@ -318,7 +318,7 @@ public class ODataPathSegmentHandler : PathSegmentHandler
                 return string.Join(
                     ",",
                     keyValuePairs.Select(keyValuePair =>
-                        TranslateNode(keyValuePair.Value).EscapeBackSlashUriString()).ToArray());
+                        TranslateNode(keyValuePair.Value)).ToArray());
             }
         }
 
@@ -328,7 +328,7 @@ public class ODataPathSegmentHandler : PathSegmentHandler
             keyValuePairs.Select(keyValuePair =>
                 (keyValuePair.Key +
                  "=" +
-                 TranslateNode(keyValuePair.Value).EscapeBackSlashUriString())).ToArray());
+                 TranslateNode(keyValuePair.Value))).ToArray());
     }
 
     internal static string TranslateNode(object node)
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.OData/Routing/ODataPathSegmentHandler.cs
```diff
@@ -364,6 +364,17 @@ internal static string TranslateNode(object node)
             return parameterAliasNode.Alias;
         }
 
-        return ODataUriUtils.ConvertToUriLiteral(node, ODataVersion.V4);
+        string uriLiteral = ODataUriUtils.ConvertToUriLiteral(node, ODataVersion.V4);
+
+        if (node is string && uriLiteral.Length > 2)
```
xuzhg, line 369: "node is string" is right?  

Node should be a kind of 'OData Query Node', if it's a string value, it should be a constantnode with string value? Maybe I missed something?

xuzhg (approved):

src/Microsoft.AspNetCore.OData/Routing/ODataPathSegmentHandler.cs
uffelauesen: It should be. According to tests it is a string.

WanjohiSammy (approved):

PR #1398, opened 2025-01-23
Fixes #1397: Conventional routing cannot identify the endpoint for action to query property without the key

### LINKED ISSUE #1397: Conventional routing cannot identify the endpoint for action to query property without the key
**Assemblies affected**

**Describe the bug**

**Reproduce steps**

**Data Model**

```C#
public class CustomerService
{
    [Contained]
    public List<StoreCustomer> customers { get; set; } = new List<StoreCustomer>();
}

public class StoreCustomer
{
    public String id { get; set; }
}
```

Then we build the Edm model using a singleton as
```C#

modelBuilder.Singleton<CustomerService>("CustomerService");
```

The controller has the following:

```
    public class CustomerServiceController : ODataController
    {
        [EnableQuery]
        [HttpGet]
        public ActionResult<IEnumerable<StoreCustomer>> GetCustomers()
        {
            return Ok(CustomersController.customers);
        }
    }
```

**Expected behavior**
The conventional routing should identify 'GetCustomers' and build endpoints for it.

**Actual behavior**
There's no endpoint generated for this action.

**Additional context**
The root cause is that:

In the NavigationRoutingConvention, there's a logic as below:
https://github.com/OData/AspNetCoreOData/blob/main/src/Microsoft.AspNetCore.OData/Routing/Conventions/NavigationRoutingConvention.cs#L68-L74

```C#
bool hasKeyParameter = action.HasODataKeyParameter(entityType, context.Options?.RouteOptions?.EnablePropertyNameCaseInsensitive ?? false);
        if (!(context.Singleton != null ^ hasKeyParameter))
        {
            // Singleton, doesn't allow to query property with key
            // entityset, doesn't allow for non-key to query property
            return false;
        }
```

However, 'hasKeyParameter' should have 'false' but it has the 'true'.

`HasODataKeyParameter` has an issue to treat the entity type without key.

xuzhg:
Fixes #1397: Conventional routing cannot identify the endpoint for action to query property without the key

### DIFF
--- a/test/Microsoft.AspNetCore.OData.Tests/Routing/Conventions/NavigationRoutingConventionTests.cs
+++ b/test/Microsoft.AspNetCore.OData.Tests/Routing/Conventions/NavigationRoutingConventionTests.cs
@@ -176,6 +176,26 @@ public class NavigationRoutingConventionTests
     }
 
     [Theory]
+    [InlineData(0)]
+    [InlineData(1)]
+    [InlineData(2)]
+    public void NavigationRoutingConventionOnSingletonWithoutKeyDefinedOnEntityTypeAsExpectedBasedOnConfiguration(int paramerterCount)
+    {
+        // Arrange
+        ControllerModel controller = ControllerModelHelpers.BuildControllerModelWithAllActions<MainSupplierController>();
+        ActionModel action = controller.Actions.First(a => a.ActionName == "GetCustomers" && a.Parameters.Count == paramerterCount);
+
+        ODataControllerActionContext context = ODataControllerActionContextHelpers.BuildContext(string.Empty, EdmModel, controller);
+        context.Action = action;
+
+        // Act
+        bool returnValue = NavigationConvention.AppliesToAction(context);
+        Assert.True(returnValue);
+
+        Assert.Equal(["/MainSupplier/Customers", "/MainSupplier/Customers/$count"], action.Selectors.Select(s => s.AttributeRouteModel.Template));
+    }
+
+    [Theory]
     [InlineData("PostToName")]
     [InlineData("Get")]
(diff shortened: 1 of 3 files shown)


xuzhg:
@https://github.com/vchianese_microsoft

test/Microsoft.AspNetCore.OData.Tests/Routing/Conventions/NavigationRoutingConventionTests.cs
```diff
@@ -285,6 +305,23 @@ private static IEdmModel GetEdmModel()
         customers.AddNavigationTarget(subOrderNavProp, orders, new EdmPathExpression("NS.VipCustomer/SubOrder"));
         me.AddNavigationTarget(subOrderNavProp, orders, new EdmPathExpression("NS.VipCustomer/SubOrder"));
 
+        // Let's build a singleton using an entity type without a key
```
gathogojr, line 308: Just so I understand, the entity type used to build a singleton is not required to have a key? Would it make sense to refer to it as an entity type then? Wouldn't it be more correct to say that the singleton is built using a complex type?

WanjohiSammy (approved):

test/Microsoft.AspNetCore.OData.Tests/Routing/Conventions/NavigationRoutingConventionTests.cs
xuzhg: If an entity type is only used to define 'singleton' or single value navigation property, the key is not required.

I think I agree with you that a singleton can be built using 'structured type', it means it can be entity type, or complex type.  FYI for @mikepizzo

PR #1403, opened 2025-02-01
Set the time zone in ODataQuerySettings from the request context

### LINKED ISSUE #384: Time zone settings not propagated to query

icnocop:
Fixes OData#384

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/ODataQueryContextExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/ODataQueryContextExtensions.cs
@@ -26,7 +26,7 @@ internal static class ODataQueryContextExtensions
             returnSettings.CopyFrom(settings);
         }
 
-        returnSettings.TimeZone = context.Request.GetTimeZoneInfo();
+        returnSettings.TimeZone ??= context.Request.GetTimeZoneInfo();
 
         return returnSettings;
     }
--- a/src/Microsoft.AspNetCore.OData/Query/ODataQueryContextExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/ODataQueryContextExtensions.cs
@@ -7,6 +7,7 @@
 
 using System.Linq;
 using Microsoft.AspNetCore.OData.Abstracts;
+using Microsoft.AspNetCore.OData.Extensions;
 using Microsoft.AspNetCore.OData.Query.Expressions;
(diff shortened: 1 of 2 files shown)


WanjohiSammy:
Please add an End-to-End test with this scenario:
https://github.com/OData/AspNetCoreOData/tree/main/test/Microsoft.AspNetCore.OData.E2E.Tests

src/Microsoft.AspNetCore.OData/Query/ODataQueryContextExtensions.cs
```diff
@@ -25,6 +26,8 @@ public static ODataQuerySettings GetODataQuerySettings(this ODataQueryContext co
             returnSettings.CopyFrom(settings);
         }
 
+        returnSettings.TimeZone ??= context.Request.GetTimeZoneInfo();
```
WanjohiSammy, line 29: Would you like to place this line of code before the `copyFrom` during the initialization of `returnSettings`?

The change you've suggested will always override the `copyFrom` action. Please note that `null` is a valid value, and users might choose to set `returnSettings.TimeZone` to `null`.

mguinness:
Has this been addressed in https://github.com/OData/AspNetCoreOData/pull/1546?

ramiabughazaleh:
> Has this been addressed in #1546?

No

PR #1407, opened 2025-02-05
Fix date() and time() function translation (#792)

### LINKED ISSUE #792: date() function is not translated
**Assemblies affected**
ASP.NET Core OData 8.x

**Describe the bug**
The [`date()` function](https://learn.microsoft.com/en-us/odata/webapi/date-and-timeofday-support) is not translated correctly. It just returns the date itself instead of removing the time.

This is caused by the translation at https://github.com/OData/AspNetCoreOData/blob/4de92f52a346606a447ec4df96c5f3cd05642f50/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.SingleValueFunctionCall.cs#L696-L708 and https://github.com/OData/AspNetCoreOData/blob/4de92f52a346606a447ec4df96c5f3cd05642f50/src/Microsoft.AspNetCore.OData/Query/Expressions/ExpressionBinderBase.cs#L404-L416

The comments imply that it is not translated because EF does not support translating the `DateTime.Date`/`DateTimeOffset.Date` properties, but this is no longer the case with EF Core. I have confirmed that querying using the `Date` property works as expected with EF Core 6+ on Postgres, SQL Server, MySQL and SQLite (probably works on older versions of EF Core too).

**Reproduce steps**
Run a query with `$compute=date(someDateTime) as d` and `$select=d`, the values returned for `d` still contain the time:
Response

**Expected behavior**
The time should be removed from the DateTime/DateTimeOffset when the `date()` function is used:
Expected response

**Additional context**
This is a fairly simple thing to fix, it's a simple translation:
```c#
protected virtual Expression BindDate(SingleValueFunctionCallNode node, QueryBinderContext context)
{
    CheckArgumentNull(node, context, "date");

    Expression[] arguments = BindArguments(node.Parameters, context);

    // We should support DateTime & DateTimeOffset even though DateTime is not part of OData v4 Spec.
    Contract.Assert(arguments.Length == 1 && ExpressionBinderHelper.IsDateOrOffset(arguments[0].Type));

    return Expression.Property(arguments[0], nameof(DateTime.Date));
}
```

The above translation works with EF Core, however, I'm not sure how this should be handled since it would be a breaking change and would break compatibility with EF6. 

It should be possible to add EF6 compatibility using the [DbFunctions.TruncateTime](https://learn.microsoft.com/en-us/dotnet/api/system.data.entity.dbfunctions.truncatetime?view=entity-framework-6.2.0) method, assuming there is a way to detect that the data provider is EF6? I'm not sure if the other EF6 database providers support this method though, I've only ever used it with SQL Server.

It would also be fine to make this a configurable option?

On a slightly unrelated note, I noticed that the binding is duplicated. It appears that some cases still use the binding in `ExpressionBinderBase` (e.g. for aggregation and compute) instead of the binders provided by Dependency Injection as for filter, select/expand and orderby. This is a bit annoying since it means that I can't override the binding to add a workaround for the issue. I'm not sure if a refactor of the aggregation and compute binders is planned, some comments suggest that it is?

jamerst:
Fix #792 - `date()` function not being translated. Also fixes similar issue of `time()` function not being translated.

The comments previously implied that the `Date` and `TimeOfDay` properties of `DateTime` and `DateTimeOffset` were not supported by EF, but this is not the case with EF Core (see [function mappings documentation](https://learn.microsoft.com/en-us/ef/core/providers/sql-server/functions#date-and-time-functions)).

The translation has been updated to use these properties for the functions so that they work correctly. Test cases have been added for when they are used in $orderby specifically.

Note that this issue was not present in most circumstances when the `date()` function was used in $filter due to the type conversion used when dealing with Date types in binary expressions resulting in dates being compared via an integer representation instead: https://github.com/OData/AspNetCoreOData/blob/c860cdc6e4f63081bdefc16e4e21873952f8101f/src/Microsoft.AspNetCore.OData/Query/Expressions/ExpressionBinderHelper.cs#L425-L441

This won't work with EF6 as it doesn't support translating the `Date` property. Not sure if this is a problem, does this library even officially support EF6? Attempting to work around it feels somewhat out of scope.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/Expressions/ExpressionBinderBase.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Expressions/ExpressionBinderBase.cs
@@ -390,10 +390,12 @@ public abstract class ExpressionBinderBase
 
         // We should support DateTime & DateTimeOffset even though DateTime is not part of OData v4 Spec.
         Contract.Assert(arguments.Length == 1 && ExpressionBinderHelper.IsDateOrOffset(arguments[0].Type));
-
-        // EF doesn't support new Date(int, int, int), also doesn't support other property access, for example DateTime.Date.
-        // Therefore, we just return the source (DateTime or DateTimeOffset).
-        return arguments[0];
+        
+        Type type = Nullable.GetUnderlyingType(arguments[0].Type) ?? arguments[0].Type;
+        PropertyInfo property = type.GetProperty(nameof(DateTime.Date));
+        
+        Expression propertyAccessExpr = ExpressionBinderHelper.MakePropertyAccess(property, arguments[0], QuerySettings);
+        return ExpressionBinderHelper.CreateFunctionCallWithNullPropagation(propertyAccessExpr, arguments, QuerySettings);
     }
 
     private Expression BindNow(SingleValueFunctionCallNode node)
@@ -403,7 +405,7 @@ public abstract class ExpressionBinderBase
         // Function Now() does not take any arguments.
         Expression[] arguments = BindArguments(node.Parameters);
         Contract.Assert(arguments.Length == 0);
-
+        
         return Expression.Property(null, typeof(DateTimeOffset), "UtcNow");
     }
 
@@ -416,9 +418,11 @@ public abstract class ExpressionBinderBase
         // We should support DateTime & DateTimeOffset even though DateTime is not part of OData v4 Spec.
         Contract.Assert(arguments.Length == 1 && ExpressionBinderHelper.IsDateOrOffset(arguments[0].Type));
 
-        // EF doesn't support new TimeOfDay(int, int, int, int), also doesn't support other property access, for example DateTimeOffset.DateTime.
-        // Therefore, we just return the source (DateTime or DateTimeOffset).
-        return arguments[0];
+        Type type = Nullable.GetUnderlyingType(arguments[0].Type) ?? arguments[0].Type;
+        PropertyInfo property = type.GetProperty(nameof(DateTime.TimeOfDay));
+
+        Expression propertyAccessExpr = ExpressionBinderHelper.MakePropertyAccess(property, arguments[0], QuerySettings);
+        return ExpressionBinderHelper.CreateFunctionCallWithNullPropagation(propertyAccessExpr, arguments, QuerySettings);
     }
 
     private Expression BindFractionalSeconds(SingleValueFunctionCallNode node)
--- a/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.SingleValueFunctionCall.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.SingleValueFunctionCall.cs
@@ -699,9 +699,11 @@ public abstract partial class QueryBinder
         // We should support DateTime & DateTimeOffset even though DateTime is not part of OData v4 Spec.
         Contract.Assert(arguments.Length == 1 && ExpressionBinderHelper.IsDateOrOffset(arguments[0].Type));
 
-        // EF doesn't support new Date(int, int, int), also doesn't support other property access, for example DateTime.Date.
-        // Therefore, we just return the source (DateTime or DateTimeOffset).
-        return arguments[0];
+        Type type = Nullable.GetUnderlyingType(arguments[0].Type) ?? arguments[0].Type;
+        PropertyInfo property = type.GetProperty(nameof(DateTime.Date));
+        
+        Expression propertyAccessExpr = ExpressionBinderHelper.MakePropertyAccess(property, arguments[0], context.QuerySettings);
+        return ExpressionBinderHelper.CreateFunctionCallWithNullPropagation(propertyAccessExpr, arguments, context.QuerySettings);
     }
 
     /// <summary>
@@ -719,9 +721,11 @@ public abstract partial class QueryBinder
         // We should support DateTime & DateTimeOffset even though DateTime is not part of OData v4 Spec.
         Contract.Assert(arguments.Length == 1 && ExpressionBinderHelper.IsDateOrOffset(arguments[0].Type));
 
-        // EF doesn't support new TimeOfDay(int, int, int, int), also doesn't support other property access, for example DateTimeOffset.DateTime.
-        // Therefore, we just return the source (DateTime or DateTimeOffset).
-        return arguments[0];
+        Type type = Nullable.GetUnderlyingType(arguments[0].Type) ?? arguments[0].Type;
+        PropertyInfo property = type.GetProperty(nameof(DateTime.TimeOfDay));
+
+        Expression propertyAccessExpr = ExpressionBinderHelper.MakePropertyAccess(property, arguments[0], context.QuerySettings);
+        return ExpressionBinderHelper.CreateFunctionCallWithNullPropagation(propertyAccessExpr, arguments, context.QuerySettings);
     }
 
     /// <summary>
--- a/test/Microsoft.AspNetCore.OData.E2E.Tests/DateAndTimeOfDay/DateAndTimeOfDayTest.cs
+++ b/test/Microsoft.AspNetCore.OData.E2E.Tests/DateAndTimeOfDay/DateAndTimeOfDayTest.cs
@@ -273,6 +273,18 @@ public class DateAndTimeOfDayTest : WebApiTestBase<DateAndTimeOfDayTest>
 
                 new[] {"$orderby=NullableTimeOfDay", "3 > 1 > 2 > 4 > 5"},
                 new[] {"$orderby=NullableTimeOfDay desc", "5 > 4 > 2 > 1 > 3"},
+                
+                new[] {"$orderby=date(SameDayDateTime), Id desc", "5 > 4 > 3 > 2 > 1"},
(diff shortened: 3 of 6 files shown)


jamerst:

src/Microsoft.AspNetCore.OData/Query/Expressions/ExpressionBinderBase.cs
```diff
@@ -390,10 +390,12 @@ private Expression BindDate(SingleValueFunctionCallNode node)
...
-
-        // EF doesn't support new Date(int, int, int), also doesn't support other property access, for example DateTime.Date.
-        // Therefore, we just return the source (DateTime or DateTimeOffset).
-        return arguments[0];
+        
+        Type type = Nullable.GetUnderlyingType(arguments[0].Type) ?? arguments[0].Type;
+        PropertyInfo property = type.GetProperty(nameof(DateTime.Date));
+        
+        Expression propertyAccessExpr = ExpressionBinderHelper.MakePropertyAccess(property, arguments[0], QuerySettings);
+        return ExpressionBinderHelper.CreateFunctionCallWithNullPropagation(propertyAccessExpr, arguments, QuerySettings);
```
WanjohiSammy, line 398: Will this change break for users still using older versions of Entity Framework other than EF6? Why not add a condition to ensure it supports older versions of EF.

src/Microsoft.AspNetCore.OData/Query/Expressions/ExpressionBinderBase.cs
```diff
@@ -416,9 +418,11 @@ private Expression BindTime(SingleValueFunctionCallNode node)
...
         Contract.Assert(arguments.Length == 1 && ExpressionBinderHelper.IsDateOrOffset(arguments[0].Type));
 
-        // EF doesn't support new TimeOfDay(int, int, int, int), also doesn't support other property access, for example DateTimeOffset.DateTime.
-        // Therefore, we just return the source (DateTime or DateTimeOffset).
-        return arguments[0];
+        Type type = Nullable.GetUnderlyingType(arguments[0].Type) ?? arguments[0].Type;
+        PropertyInfo property = type.GetProperty(nameof(DateTime.TimeOfDay));
+
+        Expression propertyAccessExpr = ExpressionBinderHelper.MakePropertyAccess(property, arguments[0], QuerySettings);
+        return ExpressionBinderHelper.CreateFunctionCallWithNullPropagation(propertyAccessExpr, arguments, QuerySettings);
```
WanjohiSammy, line 425: Same here:
Older versions might not support new features from EF6. Adding a condition ensures compatibility.

src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.SingleValueFunctionCall.cs
```diff
@@ -699,9 +699,11 @@ protected virtual Expression BindDate(SingleValueFunctionCallNode node, QueryBin
...
         Contract.Assert(arguments.Length == 1 && ExpressionBinderHelper.IsDateOrOffset(arguments[0].Type));
 
-        // EF doesn't support new Date(int, int, int), also doesn't support other property access, for example DateTime.Date.
-        // Therefore, we just return the source (DateTime or DateTimeOffset).
-        return arguments[0];
+        Type type = Nullable.GetUnderlyingType(arguments[0].Type) ?? arguments[0].Type;
+        PropertyInfo property = type.GetProperty(nameof(DateTime.Date));
+        
+        Expression propertyAccessExpr = ExpressionBinderHelper.MakePropertyAccess(property, arguments[0], context.QuerySettings);
+        return ExpressionBinderHelper.CreateFunctionCallWithNullPropagation(propertyAccessExpr, arguments, context.QuerySettings);
```
WanjohiSammy, line 706: Same here:
https://github.com/OData/AspNetCoreOData/pull/1407/files#r1944232224

test/Microsoft.AspNetCore.OData.E2E.Tests/DateAndTimeOfDay/DateAndTimeOfDayTest.cs
```diff
@@ -273,6 +273,18 @@ public static TheoryDataSet<string, string, string> OrderByData
 
                 new[] {"$orderby=NullableTimeOfDay", "3 > 1 > 2 > 4 > 5"},
                 new[] {"$orderby=NullableTimeOfDay desc", "5 > 4 > 2 > 1 > 3"},
+                
+                new[] {"$orderby=date(SameDayDateTime), Id desc", "5 > 4 > 3 > 2 > 1"},
+                new[] {"$orderby=date(SameDayNullableDateTime), Id desc", "4 > 2 > 5 > 3 > 1"},
```
WanjohiSammy, line 278: Are you using EF here?

src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.SingleValueFunctionCall.cs
jamerst: Is it possible to tell if the query provider is EF6 from here?

The only place I can find that checks if it is EF6 is in `TransformationBinderBase`, and it uses the type of `IQueryable.Provider` to do so.

That would probably work for the translation in `ExpressionBinderBase` since `TransformationBinderBase` derives from it, but `QueryBinder` isn't exposed to the queryable at any point?

https://github.com/OData/AspNetCoreOData/blob/40caec2990964cb831158efde6d69efec52dfa38/src/Microsoft.AspNetCore.OData/Query/Expressions/TransformationBinderBase.cs#L49-L54

src/Microsoft.AspNetCore.OData/Query/Expressions/ExpressionBinderBase.cs
jamerst: @WanjohiSammy Added handling for EF6 (just falls back to the previous behaviour of returning the source).

This required moving the IsClassicEF method and ClassicEF properties from TransformationBinderBase up to ExpressionBinderBase.

src/Microsoft.AspNetCore.OData/Query/Expressions/ExpressionBinderBase.cs
jamerst: Resolved same as in https://github.com/OData/AspNetCoreOData/pull/1407/files#r1944232224

src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.SingleValueFunctionCall.cs
jamerst: @WanjohiSammy Added handling for EF6 here too, same as Resolved as in https://github.com/OData/AspNetCoreOData/pull/1407/files#r1944232224.

As previously mentioned QueryBinder wasn't exposed to the queryable at any point, to get around this an IsClassicEF property has been added to QueryBinderContext and a new constructor for it accepting IQueryProvider added.

This has been added as an additional constructor to preserve the public API, the new one could be changed to be internal if wanted but leaving it as public probably makes the most sense. The new constructor should now be in use everywhere that requires it.

test/Microsoft.AspNetCore.OData.E2E.Tests/DateAndTimeOfDay/DateAndTimeOfDayTest.cs
jamerst: These tests are for the in-memory queryable provider, there are a couple of tests for EF too below

src/Microsoft.AspNetCore.OData/Query/Expressions/ExpressionBinderBase.cs
```diff
@@ -390,10 +395,18 @@ private Expression BindDate(SingleValueFunctionCallNode node)
 
         // We should support DateTime & DateTimeOffset even though DateTime is not part of OData v4 Spec.
         Contract.Assert(arguments.Length == 1 && ExpressionBinderHelper.IsDateOrOffset(arguments[0].Type));
-
-        // EF doesn't support new Date(int, int, int), also doesn't support other property access, for example DateTime.Date.
-        // Therefore, we just return the source (DateTime or DateTimeOffset).
-        return arguments[0];
+        
+        // EF6 and earlier don't support translating the Date property, so just return the original value
+        if (ClassicEF)
```
KenitoInc, line 400: The new `QueryBinder` and `QueryBinderContext` were designed to remove the tight coupling with `IQueryable` that was present in `ExpressionBinderBase` and the older Binders.
Can you add a Custom `FilterBinder` and override this method with your custom implementation? That's a cleaner solution.
Reference https://devblogs.microsoft.com/odata/customizing-filter-for-spatial-data-in-asp-net-core-odata-8/#customizing-filterbinder

KenitoInc (changes requested):

src/Microsoft.AspNetCore.OData/Query/Expressions/ExpressionBinderBase.cs
jamerst: That would be the ideal solution, however that doesn't work for all operators. Aggregation and compute operations still use `ExpressionBinderBase` so can't be customised (unless that has changed since this PR was opened?)

PR #1413, opened 2025-02-07
Fixes #1412: $select query nested inside $expand was being ignored.

### LINKED ISSUE #1412: $select query option inside $expand is ignored
**Assemblies affected**
ASP.NET Core OData > 9.1.1

**Describe the bug**

When using $select query option inside $expand, OData query builder ignores it and includes ALL structural properties in _SelectExpandNode.SelectedStructuralProperties_ collection. This causes OData resource serializer to fail while trying to create default values for properties that are not populated on expanded EDM entity object(s) because, for example, string type doesn't have the default constructor. 

This behavior first appeared in ASP.NET Core OData 9.1.3 and didn't exist in ASP.NET Core OData <= 9.1.1 (in prior versions only selected properties were populated in _SelectExpandNode.SelectedStructuralProperties_). 

**Reproduce steps**
We use untyped EDM implementation. Any model that has a simple parent -> child relationship will work. For the model below, the following query will fail: _/Sites('3C3C802D-2D04-49D9-8E33-C09A64530A9C')/Plants('E63F816D-B7E7-4FC9-AE62-7E709AFF290E')?$select=Name&$expand=Areas($select=Name)_  if expanded Area _EdmEntityObject_ only has _Name_ property set as requested by the query. 

**Data Model**

**EDM (CSDL) Model**
```xml
<EntityType Name="Site" OpenType="true">
  <Key>
    <PropertyRef Name="Id"/>
  </Key>
  <Property Name="Id" Type="Edm.String" Nullable="false" MaxLength="max"/>
  <Property Name="ConcurrencyToken" Type="Edm.String" MaxLength="32"/>
  <Property Name="Name" Type="Edm.String" Nullable="false" MaxLength="64"/>
  <Property Name="Description" Type="Edm.String" MaxLength="max"/>
  <Property Name="Created" Type="Edm.DateTimeOffset" Nullable="false"/>
  <Property Name="Updated" Type="Edm.DateTimeOffset"/>
  <NavigationProperty Name="Plants" Type="Collection(NS.Plant)" ContainsTarget="true"/>
</EntityType>
<EntityType Name="Plant" OpenType="true">
  <Key>
    <PropertyRef Name="Id"/>
  </Key>
  <Property Name="Id" Type="Edm.String" Nullable="false" MaxLength="max"/>
  <Property Name="ConcurrencyToken" Type="Edm.String" MaxLength="32"/>
  <Property Name="Name" Type="Edm.String" Nullable="false" MaxLength="64"/>
  <Property Name="Description" Type="Edm.String" MaxLength="max"/>
  <Property Name="Created" Type="Edm.DateTimeOffset" Nullable="false"/>
  <Property Name="Updated" Type="Edm.DateTimeOffset"/>
  <NavigationProperty Name="Site" Type="NS.Site" Nullable="false"/>
  <NavigationProperty Name="Areas" Type="Collection(NS.Area)" ContainsTarget="true"/>
</EntityType>
<EntityType Name="Area" OpenType="true">
  <Key>
    <PropertyRef Name="Id"/>
  </Key>
  <Property Name="Id" Type="Edm.String" Nullable="false" MaxLength="max"/>
  <Property Name="ConcurrencyToken" Type="Edm.String" MaxLength="32"/>
  <Property Name="Name" Type="Edm.String" Nullable="false" MaxLength="64"/>
  <Property Name="Description" Type="Edm.String" MaxLength="max"/>
  <Property Name="Created" Type="Edm.DateTimeOffset" Nullable="false"/>
  <Property Name="Updated" Type="Edm.DateTimeOffset"/>
  <NavigationProperty Name="Plant" Type="NS.Site" Nullable="false"/>
</EntityType>
<EntityContainer Name="Api">
  <EntitySet Name="Sites" EntityType="NS.Site">
    <NavigationPropertyBinding Path="Plants/Areas/Plant" Target="Sites/Plants"/>
    <NavigationPropertyBinding Path="Plants/Site" Target="Sites"/>
  </EntitySet>
</EntityContainer>
```

**Request/Response**

/Sites('3C3C802D-2D04-49D9-8E33-C09A64530A9C')/Plants('E63F816D-B7E7-4FC9-AE62-7E709AFF290E')?$select=Name&$expand=Areas($select=Name)

HTTP 200 (OK) response with no body. The following error is thrown from OData resource serializer (notice that _SelectedStructuralProperties_ lists ALL properties  even though only Name was selected):

Image

**Expected behavior**
The request succeeds with expanded entity only having selected _Name_ property as shown in the screenshot below taken with ASP.NET Core OData 9.1.1
Image

**Screenshots**

**Additional context**

anasik:


anasik:
@gathogojr, I just added the tests for reproducing the bugs from the issue.

src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSerializer.cs
```diff
@@ -494,7 +494,10 @@ await writer.WriteEntityReferenceLinkAsync(new ODataEntityReferenceLink
                     }
                     else
                     {
-                        bool isDelta = graph is IDelta || graph is IEdmChangedObject;
+                        Type delta = typeof(Delta<>);
+                        Type graphType = graph.GetType();
+                        bool isGeneric = graphType.IsGenericType;
+                        bool isDelta = isGeneric && graphType.GetGenericTypeDefinition() == delta;
```
gathogojr, line 500: @anasik I'll illustrate my concern with this change by presenting a typeless scenario where the current logic does not correctly express the intended state.  

### Consider the following contrived OData sample service:  

#### **Edm Model**  
```csharp
var model = new EdmModel();

var orderEntityType = model.AddEntityType("ns", "Order");
var orderIdProperty = orderEntityType.AddStructuralProperty("Id", EdmPrimitiveTypeKind.Int32);
orderEntityType.AddKeys(orderIdProperty);
orderEntityType.AddStructuralProperty("Amount", EdmPrimitiveTypeKind.Decimal);

var entityContainer = model.AddEntityContainer("Default", "Container");
entityContainer.AddEntitySet("Orders", orderEntityType);
```

#### **Service Configuration**  
```csharp
builder.Services.AddControllers().AddOData(
    options => options.EnableQueryFeatures().AddRouteComponents("delta", model));
```

#### **Controller**  
```csharp
public class DeltaOrdersController : ODataController
{
    [HttpGet("delta/Orders")]
    public EdmChangedObjectCollection Get()
    {
        var model = Request.ODataFeature().Model;
        var orderEntityType = model.SchemaElements.SingleOrDefault(d => d.Name.Equals("Order")) as IEdmEntityType;

        var changedOrders = new EdmChangedObjectCollection(orderEntityType);

        var orderDeltaObject = new EdmDeltaResourceObject(orderEntityType);
        orderDeltaObject.TrySetPropertyValue("Id", 1);
        orderDeltaObject.TrySetPropertyValue("Amount", 130);

        changedOrders.Add(orderDeltaObject);

        return changedOrders;
    }
}
```

### **Observing the Behavior**  
If you send a **GET** request to the `delta/Orders` endpoint and place a breakpoint where your change was made, you'll see the following behavior:  

image  

The intention behind this check is to confirm whether we're writing a delta resource. While it correctly identifies `Delta<T>` (typed) scenarios, it does not work as expected for typeless scenarios.  

The intended behavior was for the check to return `true` for both cases—hence the inclusion of `graph is IEdmChangedObject`. However, the challenge is that **almost all types used in typeless scenarios implement this interface**.  

### **Background on the Issue**  
Originally, `EdmStructuredObject` (the base class for `EdmEntityObject` and `EdmComplexObject`) did **not** implement `IEdmChangedObject`. However, four years ago, a change was introduced through [[this commit](https://github.com/OData/AspNetCoreOData/commit/ee937986a197b9a231f63017790b3cfd3ce0621b)](https://github.com/OData/AspNetCoreOData/commit/ee937986a197b9a231f63017790b3cfd3ce0621b) to make `EdmStructuredObject` implement `IEdmChangedObject`. This allowed both `EdmEntityObject` and `EdmComplexObject` to represent delta payloads.  

As a result, `EdmDeltaResourceObject` and `EdmDeltaComplexObject` became redundant and should have been deprecated, as noted in this **TODO**:  
[[EdmDeltaResourceObject.cs#L24](https://github.com/OData/AspNetCoreOData/blame/c60d9e0e9f68d23ee8ccd84e854ef7e430fe6edb/src/Microsoft.AspNetCore.OData/Formatter/Value/EdmDeltaResourceObject.cs#L24)](https://github.com/OData/AspNetCoreOData/blame/c60d9e0e9f68d23ee8ccd84e854ef7e430fe6edb/src/Microsoft.AspNetCore.OData/Formatter/Value/EdmDeltaResourceObject.cs#L24)  

### **Why This Is Tricky**  
Determining whether a typeless object represents a standard or delta resource **purely by inspecting the object itself, its type, or its implemented interfaces** is unreliable. Without referencing the **context**, the check can easily return incorrect results.  

When considering the use of context to distinguish delta resources, we must also ensure that **anyone overriding public methods** can use it correctly. This is why we're approaching this more comprehensively—to ensure the logic is correct for all scenarios.  

Let me know if this makes sense or if you'd like to discuss further.
cc. @julealgon @mikepizzo @xuzhg @ElizabethOkerio

anasik: Are there any other types that could pose a similar problem or is it just `EdmDeltaResourceObject`?
Because if there's not any other case we need to look out for, we can just add another check before evaluating `isDelta`

anasik: One more thing, I ran your sample on both branches. And the result is the same. 

I do agree that the `isDelta` is returning false, but I'm confused about why that's a problem? We are, after all, moving towards blurring the lines between deltas and the rest.

In both cases, the output in the client is the exact same.

gathogojr: @anasik If it wasn't clear, the `EdmDeltaResourceObject` and `EdmDeltaComplexObject` are redundant and I created an issue to track their removal in the next major release https://github.com/OData/AspNetCoreOData/issues/1429. What that means is that `EdmEntityObject` should be used in place of `EdmDeltaResourceObject` and `EdmComplexObject` should be used in place of `EdmDeltaComplexObject`. It also means that `EdmEntityObject` and `EdmComplexObject` will be used in both delta and non-delta scenarios, so you can't inspect an instance of either and determine if it's a delta resource. The context within which you're serializing the object will provide that information.

> I do agree that the isDelta is returning false, but I'm confused about why that's a problem?
I'm not understand this reasoning... If you're writing a delta resource and you're performing a check to confirm that and it returns false, is that not a problem?

> We are, after all, moving towards blurring the lines between deltas and the rest
Where are you getting this from? Can you expound?

> In both cases, the output in the client is the exact same and all tests are passing
Like I said, this is not the only area where the delta serialization logic is buggy. We're looking at fixing this more comprehensively. It is not enough that tests pass if the underlying logic isn't sound

PR #1431, opened 2025-03-04
Enable MinimalApi for OData design

xuzhg:
This is the design draft to support minimal API.

Seeking for reviews and feedbacks. 

Don't hesitate to leave your comments in the PRs.

### DIFF
--- a/docs/designs/Enable.MinimalApi.md
+++ b/docs/designs/Enable.MinimalApi.md
@@ -1,0 +1,320 @@
+# Enable OData Minimal API (Design Draft)
+See issue: https://github.com/OData/AspNetCoreOData/issues/578
+
+## Problem
+Minimal APIs are a simplified approach for building HTTP APIs fast with ASP.NET Core. Customers can build fully functioning REST endpoints with minimal code and configuration, especially without controller, action and even the formatters.
+OData works with controller/action, but OData developers also want to enable OData query on minimal API pattern.
+
+## Prerequisites
+
+Here’s the basic minimal API service app:
+```C#
+var builder = WebApplication.CreateBuilder(args);
+var app = builder.Build(); 
+app.MapGet("/customers", () => new Customer[] { …. });
+app.Run();
+```
+This service provides a Web API as:
+
+`GET {host}/customers`
+
+This API returns a collection of customers, then OData developers want to enable OData query functionalities on this API, for example:
+
+`GET {host}/customers?$select=name&$top=1`
+
+We should figure out ways to empower developers enabling all OData query options. 
+
+### Route handlers
+The lambda expression in preceding `MapGet` is called <strong>Route Handler</strong>. They are methods that execute when the route matches. Route handlers can be:
+
+- A lambda expression, 
+- A local function, 
+- An instance method
+- A static method
+- Or a RequestDelegate
+  
+where:
+```C#
+public delegate Task RequestDelegate(HttpContext context);
+```
+
+Route handlers can be synchronous or asynchronous.
+
+## Scenarios
+We want to support the following scenarios for OData Minimal API:
+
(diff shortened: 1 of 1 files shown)


docs/designs/Enable.MinimalApi.md
```diff
@@ -0,0 +1,320 @@
...
+builder.AddEndpointFilter(queryFilter);
+```
+
+c)	Use  ‘UseModel(IEdmModel)’ for Endpoint Filter.
+
+I prefer not to do ‘b’ since we can achieve the same thing using ‘c’.
+
+## Serialization
+
+The result of OData query will be serialized as normal JSON payload. Be noted, it doesn’t contain the OData control metadata, for example @odata.context.
```
mguinness, line 273: Would that also include `@odata.count`, since that's essential in paging situations

xuzhg: > Would that also include `@odata.count`, since that's essential in paging situations

If you return 'PageResult<T>' from the route handler, it should contains the 'count' and 'nextlink' properties. I will update the sample to refresh that.

xuzhg: What's the better way to specify the PageSize (Server side) value for the route handler? 
I think maybe to use the metadata. for example

```C#
app.MapGet("/customers", () => ....).WithPageSize(5);
```
@mguinness

xuzhg: Oh, Ok. I remembered that I have the overload for the 'AddODataQueryEndpointFilter' to config the query settings. 
I updated the docs to refresh this part.

mguinness: Wasn't aware of [PageResult](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.odata.results.pageresult) so that works.  Regarding page size, that would suffice if `EnableQuery` attribute isn't possible.

docs/designs/Enable.MinimalApi.md
```diff
@@ -0,0 +1,359 @@
...
+    app.MapGet("/customers", (ODataQueryOptions<Customers> options) => {
+          // …
+       }).UseOData("customized");
+```
+
+    Cons: 
+        1. First, we need the “prefix” twice to map the route handler and Edm Model. 
+        2. Second, it could be confusing with the existing controller/action mode.
+
+- Build the Edm model on the fly (Edm model-less)
```
habbes, line 131: If we build an edm model on the fly, so we can support model-less scenarios, I think this makes a case of support model-less in the query options parsers, expression trees and other parts of our stack. That way we can achieve model-less without paying the cost of building and caching a model on the fly. This is separate issue from this design, we can investigate it separately. It's something to consider is we're currently investigating perf improvements to the parser.

docs/designs/Enable.MinimalApi.md
```diff
@@ -0,0 +1,359 @@
...
+
+I prefer not to do ‘b’ since we can achieve the same thing using ‘c’.
+
+## Serialization
+
+The result of OData query will be serialized as normal JSON payload. Be noted, it doesn’t contain the OData control metadata, for example @odata.context.
+
+I think that’s true because:
+
+- It’s minimal API, developers only need the data with query functionalities.
```
habbes, line 316: I understand that there are customers for whom query options support is sufficient and they don't need an OData compliant service. But this statement rules out customers building OData services (including OData-compliant responses) that would prefer or benefit from using minimal APIs over controllers. Are the insights that there's not demand from this group? Or is that we should prioritize this use case later? For example, the comment from @mguinness already suggests that some OData conventions and semantics are desirable in the response.

Lonli-Lokli:
How will it work with versioning? Using this https://github.com/dotnet/aspnet-api-versioning

Kebechet:
Is there an ETA when this can be released ? at least as a prerelase version or temporarely as completely different nuget ?

PR #1451, opened 2025-03-27
CA2254 fixes possible formatting errors

sherlock1982:
According to [CA2254 docs](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2254) template of a log message should be a static expression.
There are a number of places where instead of logging it might cause and causes in my case subsequent formatting exception.


WanjohiSammy:
@sherlock1982 I believe we need to do this change across the entire codebase where there is logging

xuzhg (approved):

PR #1453, opened 2025-04-07
Handle IAsyncEnumerable Results When Returning Non-Generic IActionResult

etherfactor:
Adds an additional check in the [`ODataResourceSetSerializer`](https://github.com/OData/AspNetCoreOData/compare/main...etherfactor:bugfix/1450-async-iterator?expand=1#diff-4b47a17b59881163a375d365ce2d4f9177b8eabe5240be89ec8281584aba2d71) that handles a non-generic `IActionResult` response type.

The [`ObjectResultExecutor`](https://github.com/dotnet/aspnetcore/blob/main/src/Mvc/Mvc.Core/src/Infrastructure/ObjectResultExecutor.cs
) infers the return type from the generic `ActionResult<>`, but falls back on the explicit type of the returned object if the action result is the non-generic `IActionResult`. When an instance of `IAsyncEnumerable<>` is returned in an `IActionResult` response, its type definition is always the generated state machine class, not `IAsyncEnumerable<>`, causing the original condition to fail.

Also adds a test that covers this particular case.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
+++ b/src/Microsoft.AspNetCore.OData/Formatter/Serialization/ODataResourceSetSerializer.cs
@@ -9,6 +9,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics.Contracts;
+using System.Linq;
 using System.Runtime.Serialization;
 using System.Threading;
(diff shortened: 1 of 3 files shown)


etherfactor:

WanjohiSammy:
Resolve via this PR: https://github.com/OData/AspNetCoreOData/pull/1467

PR #1470, opened 2025-04-23
Fixes #580 Change PageResult<T> property names on serialization

### LINKED ISSUE #580: Change PageResult<T> property names on serialization
Hi,

I have a .NET 5 web API with OData 8.0.6 installed.
Currently I'm hitting an endpoint and I get the result as:
```
{
  "items": [ ... ],
  "count": 10
}
```

I want the the JSON response to be Uppercased like "Items" and "Count". I see there is an extension method EnableLowerCamelCase but I don't see any other method witch does the opposite like pascal casing.

ConfigureServices goes like:

```
services.AddControllers(options =>
{
    options.Filters.Add(typeof(MyCustomFilter));
})
.AddJsonOptions(options =>
{
    // I also applied pascal casing for API
    options.JsonSerializerOptions.PropertyNamingPolicy = null;
})
.AddOData(opt =>
{
    var builder = new ODataConventionModelBuilder();

    // this is the data source for PagingResult I'm returning
    builder.Function("Myfunction").ReturnsCollection<MyfunctionResult>();

    opt.AddRouteComponents(builder.GetEdmModel());
});
```

How can I achieve a paging result response to look like this?
```
{
  "Items": [ ... ],
  "Count": 10
}
```

xuzhg:
Fixes #580

Take the suggestion from https://github.com/OData/AspNetCoreOData/issues/580#issuecomment-2822486834 to update/change the property name on serialization for PageResult< T >

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Results/PageResultValueConverter.cs
+++ b/src/Microsoft.AspNetCore.OData/Results/PageResultValueConverter.cs
@@ -57,21 +57,26 @@ internal class PageResultConverter<TEntity> : JsonConverter<PageResult<TEntity>>
     public override void Write(Utf8JsonWriter writer, PageResult<TEntity> value, JsonSerializerOptions options)
     {
         writer.WriteStartObject();
-        writer.WritePropertyName("items");
+        writer.WritePropertyName(ConvertName(options, "items"));
         JsonSerializer.Serialize(writer, value.Items, options);
 
         if (value.NextPageLink != null)
         {
-            writer.WritePropertyName("nextpagelink");
+            writer.WritePropertyName(ConvertName(options, "nextpagelink"));
             writer.WriteStringValue(value.NextPageLink.OriginalString);
         }
 
         if (value.Count != null)
         {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.OData/Results/PageResultValueConverter.cs
```diff
@@ -57,21 +57,26 @@ public override PageResult<TEntity> Read(ref Utf8JsonReader reader, Type typeToC
...
-            writer.WritePropertyName("count");
+            writer.WritePropertyName(ConvertName(options, "count"));
             writer.WriteNumberValue(value.Count.Value);
         }
 
         writer.WriteEndObject();
     }
+
+    private static string ConvertName(JsonSerializerOptions options, string name)
+        => options != null && options.PropertyNamingPolicy != null
```
WanjohiSammy, line 79: Is this only targeting `PropertyNamingPolicy`

WanjohiSammy (approved):

PR #1482, opened 2025-05-30
Fix case insensitive property selection

CAPCHIK:
On types where the names of properties differ only in case, the choice of a property at the CLR level breaks down, while OData allows you to have such names and processes them correctly.

The change does not change the number of test statuses. Before making the change, there were 7851 passed tests and 8 failed tests on my PC, and these numbers remained after making the changes. And the logic of selecting a property started working correctly, since by default, getting a property does not have an [indication to ignore the case](https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/System.Linq.Expressions/src/System/Linq/Expressions/MemberExpression.cs#L211-L212).

The behavior is reproduced in the [repository](https://github.com/CAPCHIK/ODataCaseSensitiveExample), and the exception can be observed in the [github action](https://github.com/CAPCHIK/ODataCaseSensitiveExample/actions/runs/15353393955/job/43206829355#step:5:9).


gathogojr:
Hello @CAPCHIK. Could you please add some test coverage to this pull request? It would help us better understand the specific scenario your change is addressing and ensure the behavior is preserved going forward.

PR #1485, opened 2025-06-02
Add TryValidate for ODataQueryOptions

### LINKED ISSUE #1419: Add ODataQueryOptions.TryValidate
There is currently no way to validate query options without throwing an exception. I'd like a `TryValidate` method, so I don't need to catch an exception to find out that the query options don't satisfy the settings. Ideally it would have an `out` parameter which includes the disallowed query option.

```csharp
public ActionResult< IQueryable > Get(
	ODataQueryOptions< TResult > queryOptions )
{
	ODataValidationSettings validationSettings =
		new( )
		{
			AllowedQueryOptions = AllowedQueryOptions.Select | AllowedQueryOptions.SkipToken,
			MaxTop = 1000,
		};

	try
	{
		queryOptions.Validate( validationSettings );
	}
	catch ( Exception )
	{
		return BadRequest( "Unsupported query option" );
	}
}
```

WanjohiSammy:
### Issues

*This pull request fixes #1419.*

### Description

This PR introduces a `TryValidate` methods to `ODataQueryOptions` addressing the current limitation where query options can only be validated by throwing exceptions.

The `TryValidate` method provides a way to validate query options without relying on exception handling. Instead of relying on exception handling, this method returns a bool indicating success or failure and provides detailed validation errors through an out parameter.

### Example Usage:
```cs
var filterOption = new FilterQueryOption("Name eq 'John'", context);
var settings = new ODataValidationSettings();

if (_validator.TryValidate(filterOption, settings, out IEnumerable<string> validationErrors))
{
    Console.WriteLine("Validation succeeded!");
}
else
{
    Console.WriteLine("Validation failed:");
    foreach (var error in validationErrors)
    {
        Console.WriteLine(error);
    }
}
```

- [x] *Test cases added*

### Additional work necessary


src/Microsoft.AspNetCore.OData/Query/Validator/Interfaces/IComputeQueryValidator.cs
```diff
@@ -14,9 +16,18 @@ namespace Microsoft.AspNetCore.OData.Query.Validator;
...
     void Validate(ComputeQueryOption computeQueryOption, ODataValidationSettings validationSettings);
+
+    /// <summary>
+    /// Attempts to validate the <see cref="ComputeQueryOption" />.
+    /// </summary>
+    /// <param name="computeQueryOption">The $compute query.</param>
+    /// <param name="validationSettings">The validation settings.</param>
+    /// <param name="validationErrors">Contains a collection of validation errors encountered, or an empty collection if validation succeeds.</param>
+    /// <returns><see langword="true"/> if the validation succeeded; otherwise, <see langword="false"/>.</returns>
+    bool TryValidate(ComputeQueryOption computeQueryOption, ODataValidationSettings validationSettings, out IEnumerable<string> validationErrors);
```
xuzhg, line 32: In this pattern, shall we call 'Validate' within 'TryValidate' and avoid duplicate the validation codes?

xuzhg: Is it the time to add 'async' interface method?

PR #1488, opened 2025-06-12
Fixes #1487 : Minimal API TimeZoneInfo for Serialization

### LINKED ISSUE #1487: Minimal API TimeZoneInfo for Serialization
I know you can set TimeZoneInfo on the ODataQuerySettings, but this appears to only be used in the ExpressionBinderHelper class and not for serialization.

It'd be nice to add the option to configure TimeZoneInfo for serialization like we have when using OData with controllers. Currently the result from an endpoint always comes back with DateTime displayed using the local time of the server.

Controller Service Configuration:

```cs
builder.Services.AddControllers()
    .AddOData(opt =>
    {
        opt.EnableQueryFeatures();
        opt.AddRouteComponents("odata", EdmModelHelper.GetEdmModel());
        opt.TimeZone = TimeZoneInfo.Utc; // Missing from Minimal API Service Configuration
    });
```

Minimal API Service Configuration Example:

```cs
builder.Services.AddOData(q =>
{
    q.EnableAll();
    q.SetCaseInsensitive(true);
    q.SetTimeZoneInfo(TimeZoneInfo.Utc); // Add something like this to be able to set DateTime serialization options.
});
```

xuzhg:
1. Fixes #1487 : Minimal API TimeZoneInfo for Serialization in minimal APi
2. Enable ODataActionParameter and ODataUntypedActionParameter binding

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Extensions/HttpContextExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Extensions/HttpContextExtensions.cs
@@ -223,7 +223,7 @@ public static class HttpContextExtensions
         ODataMiniMetadata metadata = endpoint.Metadata.GetMetadata<ODataMiniMetadata>();
         if (metadata is null || metadata.Model is null || metadata.PathFactory is null)
         {
-            throw new ODataException($"Please call WithODataModel() and WithODataPathFactory() for the endpoint");
+            throw new ODataException(SRResources.ODataMustBeSetOnMinimalAPIEndpoint);
         }
 
         Type parameterType = parameter.ParameterType;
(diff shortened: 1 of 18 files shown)


src/Microsoft.AspNetCore.OData/Extensions/HttpContextExtensions.cs
```diff
@@ -196,6 +213,74 @@ internal static IServiceProvider GetOrCreateServiceProvider(this HttpContext htt
...
+                toDispose).ConfigureAwait(false);
+
+            foreach (IDisposable obj in toDispose)
+            {
+                obj.Dispose();
+            }
+        }
+        catch (Exception ex)
+        {
+            throw new ODataException($"Cannot bind parameter '{parameter.Name}'. the error is: {ex.Message}");
```
WanjohiSammy, line 261: Same here. Should we move the text to `SRResources`

WanjohiSammy (approved):

src/Microsoft.AspNetCore.OData/Extensions/HttpContextExtensions.cs
xuzhg: #Resolved, Thanks.

WanjohiSammy (approved):

PR #1492, opened 2025-06-25
Fix/Improve TruncatedCollectionOfT

WanjohiSammy:


src/Microsoft.AspNetCore.OData/Query/Container/TruncatedCollectionOfT.cs
```diff
@@ -6,170 +6,352 @@
...
+    private static async Task<TruncatedCollection<T>> CreateInternalAsync(IAsyncEnumerable<T> source, int pageSize, long? totalCount, CancellationToken cancellationToken)
+    {
+        ValidateArgs(source, pageSize);
+
+        int capacity = pageSize > 0 ? pageSize : (totalCount > 0 ? (totalCount < int.MaxValue ? (int)totalCount : int.MaxValue) : DefaultCapacity);
+        var buffer = new List<T>(capacity);
+
+        bool isTruncated = false;
+        int count = 0;
+        await foreach (var item in source.Take(checked(capacity + 1)).WithCancellation(cancellationToken).ConfigureAwait(false))
```
rpallares, line 247: I don't quite understand. Isn't it strange to enumerate an IAsyncENumerable to put it in a list and then put it back in a new IAsyncEnumerable?

Shouldn't we stream the data completely? From what I understood, the only real problem is that the count was accessed at the beginning, but this is no longer the case since https://github.com/OData/AspNetCoreOData/pull/1393

Shouldn't we introduce a TrucatedEnumerable instead?

WanjohiSammy: Thanks @rpallares for pointing this out. I have updated it. Could you take another look.

peter-bertok:
We've been waiting for this issue to be fixed for years now. I just ran a quick test of this PR against a particularly problematic application that streams ~100K rows per page:

- Memory usage dropped from 1,700 MB to 75 MB.
- Time to first byte dropped from 2 seconds to 75 milliseconds.
- Time to last byte dropped by 7%.

That is a _huge_ improvement! I think it's well worth investing the time to merge this PR!

NOTE: The PR alone is insufficient, because it just "preps the ground", but doesn't actually wire up the new `TruncatedAsyncEnumerable<T>` type. I wired it up using a vibe-coded patch of the OData library to enable testing.

PR #1507, opened 2025-07-29
fixed NoDollar top and skip in nextlink

### LINKED ISSUE #1505: "?top" and "?skip" unmodified in nextLink when EnableNoDollarQueryOptions is true.
**Assemblies affected**
ASP.NET Core OData 8.x
ASP.NET Core OData 9.x

**Describe the bug**
The `@odata.nextLink` in the response is supposed to modify both the `$top` and `$skip` options if present in the query.

If `$top` > `MaxPageSize`, the `$top` value gets decremented by the pageSize before being included in the `@odata.nextLink`
If `$skip` is present and `$skiptoken` is enabled, the `$skip` is simply not included in the `@odata.nextLink`

With `EnableNoDollarQueryOptions = true`, which is default behaviour, we can use `top` and `skip` in our queries, in addition to the standard `$top` and `$skip`.

However, if `top` and `skip`, i.e. the no dollar versions, are present in the query, they are simply copied, as is, into the `@odata.nextLink` rendering it incorrect.

**Request/Response**
```
http://localhost:5203/odata/Product?top=15&skip=1
```

**Expected behavior**
Assuming `MaxPageSize = 2`
```
...
"@odata.nextLink": "http://localhost:5203/odata/Product?top=13&$skiptoken=Id-15"
}
```

What's actually happening is 

```
...
"@odata.nextLink": "http://localhost:5203/odata/Product?top=15&skip=1&$skiptoken=Id-15"
}
```

anasik:
Also, fixes #1505 but for `9.x`

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Extensions/GetNextPageHelper.cs
+++ b/src/Microsoft.AspNetCore.OData/Extensions/GetNextPageHelper.cs
@@ -44,6 +44,7 @@ internal static class GetNextPageHelper
             switch (key)
             {
                 case "$top":
+                case "top":
                     int top;
                     if (Int32.TryParse(value, out top))
                     {
@@ -60,6 +61,7 @@ internal static class GetNextPageHelper
                     }
                     break;
                 case "$skip":
+                case "skip":
                     if (useDefaultSkip)
                     {
                         //Need to increment skip only if we are not using skiptoken 


src/Microsoft.AspNetCore.OData/Extensions/GetNextPageHelper.cs
```diff
@@ -44,6 +44,7 @@ internal static Uri GetNextPageLink(Uri requestUri, IEnumerable<KeyValuePair<str
             switch (key)
             {
                 case "$top":
+                case "top":
```
xuzhg, line 47: Shall we check the configuration?

anasik: You mean only match the no dollar case if nodollarqueryoptions is enabled? 

I thought about it at the time and I thought it would make the code ugly and the changes of anyone using "top" for anything other than this were incredibly slim.

But I can add a check.

xuzhg (approved):

anasik:
@xuzhg, same test failing here too?

xuzhg:
> @xuzhg, same test failing here too?

I didn't check this, but I think it should be the same. Please run those test cases at your side to verify.

xuzhg (approved):

PR #1529, opened 2025-09-22
Fix #1518: Add additional check in IsMinimalEndpoint to prevent false positives

### LINKED ISSUE #1518: EnableNoDollarQueryOptions not working as expected
**Assemblies affected**
only affects ASP.NET Core OData 9.4.0
ASP.NET Core OData 9.3.2 and previous: working as expected

**Describe the bug**
When configuring EnableNoDollarQueryOptions=false, and adding an object of type ODataQueryOptions<T> to an endpoint, the ODataQueryOptions.Top property is being filled and contains the top query parameter value, whether a $ was appended as a prefix to it or not.

**Reproduce steps**
* Set EnableNoDollarQueryOptions=false
```
services.AddControllers()
                .AddOData(option =>
                {
                    option.AddRouteComponents("odata", new ODataConventionModelBuilder().GetEdmModel());
                    option.Count()
                        .Filter()
                        .OrderBy()
                        .Expand()
                        .Select()
                        .SetMaxTop(null);
                    option.EnableNoDollarQueryOptions = false;
                });
```
* Add some endpoint
```
        [HttpGet]
        [ODataRouteComponent("Table")]
        [Route("Table")]
        [AllowAnonymous]
        public async Task<IActionResult> GetAll(ODataQueryOptions<object> options, int skip = 0, int top = 22, bool latestData = false)
        {
            return Ok(options.Top is null);
        }
```
* Call the endpoint, ex. `GET http://localhost:5000/api/test/Table?top=25`

**Expected behavior**
Expected output: options.Top should be null. Under 9.4.0, the Top property is populated and reads the value from the top query parameter directly.
Before 9.4.0, options.Top is null.

Gilbert32:
When Minimal Apis are not used, and controllers are, `ODataQueryOptions<TEntity>` always calls `PopulateMetadata()`, which will always add metadata of type `ODataMiniMetadata` to the endpoint.

Currently, when `httpContext.IsMinimalEndpoint()` is called, it checks whether `ODataMiniMetadata` exists which is always the case.

This commits makes it additionally check for an `HttpMethodAttribute` in the endpoint metadata.

Additionally, when `PopulateMetadata()` was called, the endpointBuilder.Metadata list was empty, so it was not possible to check for other attributes.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Extensions/HttpContextExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Extensions/HttpContextExtensions.cs
@@ -11,6 +11,7 @@ using System.Linq;
 using System.Reflection;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc.Routing;
 using Microsoft.AspNetCore.OData.Abstracts;
 using Microsoft.AspNetCore.OData.Edm;
 using Microsoft.AspNetCore.OData.Formatter;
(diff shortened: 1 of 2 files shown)


Gilbert32:

PR #1534, opened 2025-10-17
Set read untyped as string using reflection

xuzhg:
ODL will remove 'ReadUntypedAsString', 

However, ASP.NET Core OData is used in OData.Net E2E test case which 'temporally' set the 'ReadUntypedAsString=false'.

For back-compatibility, We don't remove it completely in current version and let's use the reflection to replace the property setting directly.

In the next major version, let's remove such codes.

I don't think it's a bigger performance. 

I do update the version to 9.4.1 and will use this version in the OData.NET E2E test cases.

### DIFF
--- a/tool/builder.versions.settings.targets
+++ b/tool/builder.versions.settings.targets
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <VersionMajor Condition="'$(VersionMajor)' == ''">9</VersionMajor>
     <VersionMinor Condition="'$(VersionMinor)' == ''">4</VersionMinor>
-    <VersionBuild Condition="'$(VersionBuild)' == ''">0</VersionBuild>
+    <VersionBuild Condition="'$(VersionBuild)' == ''">1</VersionBuild>
     <VersionRelease Condition="'$(VersionRelease)' == ''"></VersionRelease>
   </PropertyGroup>
   


mikepizzo:
I don't understand why we need this -- ReadUntypedAsString="false" should be the default. That is, once we fully deprecate read untyped as string, ODL should *never* read untyped as string -- it should always read as structured.

So why do we need to set a flag through reflection to do what should be the default?

xuzhg:
> ReadUntypedAsString=false should be the default when we remove the property from ODL. We shouldn't need to set it to false.

@mikepizzo So, do you think 'https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/ODataMessageReaderSettings.cs#L56' is wrong?

WanjohiSammy (approved):

mikepizzo:
> > ReadUntypedAsString=false should be the default when we remove the property from ODL. We shouldn't need to set it to false.
> 
> @mikepizzo So, do you think 'https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/ODataMessageReaderSettings.cs#L56' is wrong?

@xuzhg  -- Yes -- as part of deprecating ReadUntypedAsString, we should *always* default to false (that is, we should always read untyped as structured values).  We should *never* read untyped values as raw strings, regardless the OData version. Whether or not to read as string really isn't dependent upon the version -- we just added that so that, at least for new payload versions, we would never have the problem. Deprecating was removing that behavior from *all* versions.

PR #1546, opened 2025-12-03
Fixes #1545: ODataOptions.TimeZone is ignored; dates are always serialized using TimeZoneInfo.Local

### LINKED ISSUE #1545: ODataOptions.TimeZone is ignored; dates are always serialized using TimeZoneInfo.Local
**Assemblies affected**
Version 9.4.1 (Updating from version 9.3.0)

**Describe the bug**
When adding ODataQueryOptions to an [HttpGet] endpoint, `opt.TimeZone = TimeZoneInfo.Utc` is ignored, and ODataMiniMetadata.TimeZone (which defaults to Local) is used instead

**Reproduce steps**
Minimal repro:

```
using Microsoft.AspNetCore.OData;
using Microsoft.AspNetCore.OData.Query;
using Microsoft.AspNetCore.OData.Routing.Controllers;
using Microsoft.OData.ModelBuilder;

var builder = WebApplication.CreateBuilder(args);

builder.Services.AddControllers().AddOData(opt =>
    {
        opt.EnableQueryFeatures();
        opt.TimeZone = TimeZoneInfo.Utc; // Ignored when using DataQueryOptions<TestModel> parameter

        var edmBuilder = new ODataConventionModelBuilder();
        edmBuilder.EntitySet<TestModel>("Test");

        opt.AddRouteComponents("v1", edmBuilder.GetEdmModel());
    });

var app = builder.Build();

app.UseRouting();
app.MapControllers();

app.Run();

public class TestController : ODataController
{
    private static readonly IList<TestModel> _models = new[]
    {
        new TestModel { Id = 1, Name = "Test111", Date = DateTime.UtcNow },
        new TestModel { Id = 2, Name = "Testaaa", Date = DateTime.UtcNow.AddDays(1) }
    };

    [HttpGet]
    public IActionResult Get(ODataQueryOptions<TestModel> options)
    {
        return Ok(_models.AsQueryable());
    }
}

public class TestModel
{
    public int Id { get; set; }
    public DateTime Date { get; set; }
}
```

When removing the ODataQueryOptions<TestModel> parameter, dates are serialized in UTC format.

**Request/Response**
GET https://localhost/v1/Test

**Expected behavior**
Dates should be serialized in UTC format, but they are currently being serialized in local time

xuzhg:
Fixes #1545: 

Root cause:

1) If `ODataQueryOptions<T>` is used as parameter in the controller action, for example:

```C#
  [HttpGet]
  public IActionResult Get(ODataQueryOptions<TestModel> options)
  {
      return Ok(_models.AsQueryable());
  }
```

or minimal API as:
```C#
app.MapGet("/v2/test", (ODataQueryOptions<TestModel> options) =>
{
    IList<TestModel> _model = new[]
        {
        new TestModel { Id = 1, Name = "Test111", Date = DateTime.UtcNow },
        new TestModel { Id = 2, Name = "Testaaa", Date = DateTime.UtcNow.AddDays(1) }
    };

    return options.ApplyTo(_model.AsQueryable());
})
    .WithODataResult();
```
Then, the following code is executed ahead:

https://github.com/OData/AspNetCoreOData/blob/main/src/Microsoft.AspNetCore.OData/Query/ODataQueryOptionsOfT.cs#L132

```C#
    public static void PopulateMetadata(ParameterInfo parameter, EndpointBuilder builder)
    {
        // Make sure we have the metadata added into the endpoint.
        // Shall we build the 'EdmModel' here? ==> Emm...No, because the 'convention' runs after this population.
        // If developer calls 'WithODataModel()', then any model created here will be replaced. So, no need/required to create model here.
        ODataEndpointConventionBuilderExtensions.ConfigureODataMetadata(builder, null);
   }
```

The processing codes is to add 'ODataMiniMetadata' into the endpoint metadata container, even it's controller-based application and minimal api application. 

Then, a request coming as: `GET {{WebApplication2_HostAddress}}/v1/Test`

The library tries to retrieve the 'TimeZone' configuration from the options, here: https://github.com/OData/AspNetCoreOData/blob/main/src/Microsoft.AspNetCore.OData/Extensions/HttpRequestExtensions.cs#L102

but, https://github.com/OData/AspNetCoreOData/blob/main/src/Microsoft.AspNetCore.OData/Extensions/HttpContextExtensions.cs#L92 only checks 'ODataMiniMetadata'. 

in this case, controller-based application is also treated as minimal api. That's wrong.

This PR is to fix that by checking 'ODataRoutingMetadata' first. In the controller-based, the endpoint contains 'ODataRoutingMetadata', meanwhile, it's not in the minimal api.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Extensions/HttpContextExtensions.cs
+++ b/src/Microsoft.AspNetCore.OData/Extensions/HttpContextExtensions.cs
@@ -16,6 +16,7 @@ using Microsoft.AspNetCore.OData.Abstracts;
 using Microsoft.AspNetCore.OData.Edm;
 using Microsoft.AspNetCore.OData.Formatter;
 using Microsoft.AspNetCore.OData.Results;
+using Microsoft.AspNetCore.OData.Routing;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
 using Microsoft.OData;
@@ -98,6 +99,11 @@ public static class HttpContextExtensions
 
         // Check if the endpoint is a minimal endpoint.
         var endpoint = httpContext.GetEndpoint();
+        if (endpoint?.Metadata.GetMetadata<IODataRoutingMetadata>() != null)
+        {
+            return false;
+        }
+
         return endpoint?.Metadata.GetMetadata<HttpMethodAttribute>() == null
                && endpoint?.Metadata.GetMetadata<ODataMiniMetadata>() != null;
     }


WanjohiSammy (approved):
Is it possible to add some tests for this case to ensure the issue is fixed

mguinness:
Still seeing the problem using latest version [10.0.0-preview.2](https://www.nuget.org/packages/Microsoft.AspNetCore.OData/10.0.0-preview.2) with repro at https://github.com/OData/AspNetCoreOData/issues/384#issuecomment-3882128062.

leoerlandsson:
Still seeing this problem in 9.4.1 and in 10.0 preview [xuzhg](https://github.com/xuzhg)

PR #1560, opened 2026-02-20
Fix: Make EnsureStableSortOrderBy respect the ordering direction for added fields

jbaehr:
Previously, the fields added to the $orderby clause to enable stable ordering were always in ascending order -- no matter what the original sorting order was. This has led to two undesired side effects:
- As a user, I expect my results to be returned in reverse order if I switch from asc to desc. Yet the order of items sharing the same value for my original sorting criterion were not reversed.
- As a developer, I expect the same indices to be used (just traversed in reverse order) if the sorting order from the user query is reversed. Yet the generated query is not simply reversed, preventing the same composite indices to be used. E.g. for MongoDB this currently requires an additional index.

This commit addresses both issues by propagating the sorting order of the last user-provided ordering criterion to all additional criteria required to stabilize the ordering. So when the original query was "$orderby=Foo" I got (and still get) an ordering of "Foo asc, Id asc". Reversing this to "$orderby=Foo desc" I previously got "Foo desc, Id asc" and now get "Foo desc, Id desc". This means an index on {Foo:1, Id:1} can be used for both queries while previously an additional index on {Foo:-1, Id:1} was required (example for MongoDB).
And for the user a result of "(A,1);(A,2); ... (Z,55);(Z,66)" reversed is now "(Z,66);(Z55); ... (A,2);(A,1)" as intuitively expected. Previously it was "(Z55);(Z,66); ... (A,1);(A,2)".

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/ODataQueryOptions.cs
@@ -748,7 +748,8 @@ public class ODataQueryOptions
             var propertyPathsToAdd = applySortOptions.Where(p => !usedPropertyNames.Contains(p)).OrderBy(p => p);
             if (propertyPathsToAdd.Any())
             {
-                var orderByRaw = orderBy.RawValue + "," + String.Join(",", propertyPathsToAdd);
+                var suffix = ExtractDirectionSuffix(orderBy);
+                var orderByRaw = orderBy.RawValue + "," + String.Join(",", propertyPathsToAdd.Select(p => p + suffix));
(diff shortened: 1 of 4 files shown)


jbaehr:
@microsoft-github-policy-service agree company="Rohde & Schwarz GmbH & Co. KG"

jbaehr:
@xuzhg @gathogojr should I create a dedicated issue for the misbehavior fixed here or is the PR enough?

jbaehr:
@xuzhg @gathogojr could you please provide a rough estimate on when you could take a look at this PR?

jbaehr:
@WanjohiSammy, I've seen that you lately reviewed and merged some PRs here. Could you maybe provide a rough estimate on when to expect some feedback this PR?

PR #1562, opened 2026-02-22
Tests on linux fail because of hardcoded line-endings

PatrickDahlin:
This fixes tests using hardcoded returnvalue comparisons using preset newlines in windows format (CRLF).
Handling here is mosly using ReplaceLineEndings and ignoreLineEndingDifferences parameter to use the environment newlines instead.


PatrickDahlin:

PR #1565, opened 2026-04-02
Fix: Handle null propagation for nullable navigation properties when expanding and ordering

MartinTopfstedtSinc:
## Summary

Fixes null propagation handling when expanding nullable navigation properties and ordering by their nullable properties.

## Changes

- Added `IsNullableNavigationProperty()` extension method to `IEdmProperty` to properly identify nullable navigation properties
- Updated `QueryBinder.CreatePropertyAccessExpression()` to include nullable navigation properties in null propagation checks
- The fix ensures that when `HandleNullPropagation` is enabled, both nullable value types AND nullable navigation properties are properly handled during query binding

## Why this matters

Previously, when expanding a nullable navigation property and then ordering by an integer property of that same nullable navigation property, the null propagation logic would not correctly handle the nullable navigation property case, potentially causing runtime errors or incorrect query results.

The updated logic now checks for either:
1. Nullable value types (`ExpressionBinderHelper.IsNullable(source.Type)`)
2. Nullable navigation properties (`property.IsNullableNavigationProperty()`)

This ensures consistent null handling across both scenarios.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Edm/EdmHelpers.cs
+++ b/src/Microsoft.AspNetCore.OData/Edm/EdmHelpers.cs
@@ -520,4 +520,18 @@ internal static class EdmHelpers
         Contract.Assert(model != null);
         return model.GetAnnotationValue<OperationTitleAnnotation>(operation);
     }
+
+    /// <summary>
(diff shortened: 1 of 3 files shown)


MartinTopfstedtSinc:
@microsoft-github-policy-service agree company="SINC"

PR #1578, opened 2026-05-19
Add totalseconds canonical function binding

RicardoGS98:
## Summary

Implements the OData V4 canonical function `totalseconds(Edm.Duration)`, which is recognized by the URI parser but is not bound by the LINQ expression binder — currently producing `Unknown function 'totalseconds'` at query time (see `SRResources.ODataFunctionNotSupported`).

Per the spec, `totalseconds` returns an `Edm.Decimal` representing the duration as a fractional number of seconds. The CLR mapping is `TimeSpan.TotalSeconds` cast to `decimal`.

This also makes the common diff-between-dates pattern work out of the box:

```
$filter=totalseconds(EndDate sub StartDate) gt 3600
```

because the OData parser types `DateTimeOffset sub DateTimeOffset` as `Edm.Duration`, and the binary subtract binder already produces a `TimeSpan` LINQ expression, which `BindTotalSeconds` then consumes.

## Changes

- `ClrCanonicalFunctions`: add `TotalSecondsFunctionName` and `TimeSpanTotalSeconds` `PropertyInfo`.
- `QueryBinder.SingleValueFunctionCall`: route `totalseconds` to a new `virtual BindTotalSeconds` method.
- `AllowedFunctions`: add `TotalSeconds = 0x40000000`; include it in `AllDateTimeFunctions` (and transitively in `AllFunctions`).
- `FilterQueryValidator` + `QueryValidatorHelpers`: map `"totalseconds"` to `AllowedFunctions.TotalSeconds` in both (they were duplicated).
- Tests:
  - `FilterBinderTests`: covers `totalseconds(TimeSpanProp)` and `totalseconds(DateTimeOffset sub DateTimeOffset)`.
  - `FilterQueryValidatorTests` and `AllowedFunctionsTests`: data sets updated so the group/exclude tests cover the new flag.
- `PublicAPI.Unshipped.txt` and `PublicApi.bsl` baselines updated.

## Note on `totaloffsetminutes`

I considered including `totaloffsetminutes` in the same change since it tends to come up together, but mapping it requires the sign bit (`0x80000000`) of the `AllowedFunctions` enum, which makes `AllFunctions` negative and breaks the `value > AllowedFunctions.AllFunctions` range check in `ODataValidationSettings`. That would need either widening the underlying type to `long` (binary breaking) or a separate validation strategy, so I left it out of this PR.

## Test plan

- [x] `dotnet build src/Microsoft.AspNetCore.OData/Microsoft.AspNetCore.OData.csproj` — 0 errors.
- [x] `dotnet test test/Microsoft.AspNetCore.OData.Tests/Microsoft.AspNetCore.OData.Tests.csproj` — 6336 passed, 5 skipped, 0 failed (locally).
- [x] PublicAPI baseline regenerated and matched.
- [ ] Reviewer to verify EDM/URI parser already exposes the canonical signature for `totalseconds(Edm.Duration)` without further model registration (it does on the cases I tested; flagging for confirmation).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


RicardoGS98:

WanjohiSammy:
@RicardoGS98 
Please add E2E tests

RicardoGS98:
> @RicardoGS98 Please add E2E tests

Thanks for the review — E2E coverage added in `test/Microsoft.AspNetCore.OData.E2E.Tests/DurationFunctions/`
(5 new files, no existing E2E test modified). 37 E2E cases over a 5-row in-memory dataset.

PR #1282, opened 2024-07-16
Fixes #1286: String Collections support in $select. 

anasik:
[EF Core 8 comes with support for primtive collections in models. ](https://learn.microsoft.com/en-us/ef/core/what-is-new/ef-core-8.0/whatsnew#primitive-collections)

Naturally, OData should pick up on that support. As of now, if you create a `List<string>` field in your model and then fetch all the records for that model, it works flawlessly. You see an array of strings in the response. 

However, if you try to `$select` that field, it fails with the following error: 

<details>
  <summary>Click to expand error</summary>
```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.EntityFrameworkCore.Query.RelationalShapedQueryCompilingExpressionVisitor.ShaperProcessingExpressionVisitor.CreateGetValueExpression(ParameterExpression dbDataReader, Int32 index, Boolean nullable, RelationalTypeMapping typeMapping, Type type, IPropertyBase property)
   at Microsoft.EntityFrameworkCore.Query.RelationalShapedQueryCompilingExpressionVisitor.ShaperProcessingExpressionVisitor.VisitExtension(Expression extensionExpression)
   at Microsoft.EntityFrameworkCore.Query.RelationalShapedQueryCompilingExpressionVisitor.ShaperProcessingExpressionVisitor.ProcessShaper(Expression shaperExpression, RelationalCommandCache& relationalCommandCache, IReadOnlyList`1& readerColumns, LambdaExpression& relatedDataLoaders, Int32& collectionId)
   at Microsoft.EntityFrameworkCore.Query.RelationalShapedQueryCompilingExpressionVisitor.ShaperProcessingExpressionVisitor.VisitExtension(Expression extensionExpression)
   at System.Linq.Expressions.ExpressionVisitor.VisitMemberAssignment(MemberAssignment node)
   at System.Linq.Expressions.ExpressionVisitor.VisitMemberBinding(MemberBinding node)
   at System.Linq.Expressions.ExpressionVisitor.Visit[T](ReadOnlyCollection`1 nodes, Func`2 elementVisitor)
   at System.Linq.Expressions.ExpressionVisitor.VisitMemberInit(MemberInitExpression node)
   at System.Linq.Expressions.ExpressionVisitor.VisitMemberAssignment(MemberAssignment node)
   at System.Linq.Expressions.ExpressionVisitor.VisitMemberBinding(MemberBinding node)
[...]

Issue #1268, opened 2024-06-25
Disabling Automatic Expansion in Nested OData Entity Objects


anjanasagathiya:
**Subject:** Disabling Automatic Expansion in Nested OData Entity Objects

**Problem Description:**

We are using AspNetCoreOData version 7.20.0 and need to disable the automatic expansion of nested objects in our OData entities. Here FYI, the data retrieval is from a single table.

**Example** Below is an example illustrating the issue.

```
Public class OdataEntity
{
  [DataMember(Name = "id")]
  public long Id { get; set; }
  
  [DataMember(Name = "credit_memo")]
  public CreditMemo CreditMemo { get; set; }
}

[ComplexType] 
[DataContract]
public class CreditMemo
{
   [DataMember(Name = "id")]
   public long Id { get; set; }

   [DataMember(Name = "name")]
   public string Name { get; set; }
}

```
In the above OdataEntity, we want to prevent the automatic expansion of the CreditMemo object. If the user explicitly includes $expand=CreditMemo in the OData query, then the CreditMemo object should be included in the response.

Could you please guide us on how to achieve this?

We appreciate any insights or solutions that can help resolve this issue efficiently.

WanjohiSammy:
@anjanasagathiya `ComplexType` cannot be directly used as a navigation property because navigation properties imply entity relationships, and `ComplexTypes` do not have identities or keys. 
[Complex types](https://learn.microsoft.com/en-us/odata/concepts/data-model?view=odata-edm-7.0#complex-types)

However, it is possible to have navigation properties under `ComplexType`.

You can review the doc: 
- [Complex Type with Navigation Property](https://learn.microsoft.com/en-us/odata/webapi/complextypewithnavigationproperty)
- [Navigation property in complex type](https://learn.microsoft.com/en-us/odata/odatalib/navigation-under-complex)

anjanasagathiya:
@WanjohiSammy My question is related to expansion of navigation property. How can I disable automatic expansion of navigation property if data is retrieved from same table?

julealgon:
@anjanasagathiya you should consider renaming your question to make it explicit you are talking about a complex type model. Normal navigations are _not_ expanded by default.

Have you tried just calling `AutoExpand(false)` in the EDM configuration for your property?

WanjohiSammy:
@anjanasagathiya
Update us on whether you’ve resolved the issue related to navigation properties. 
[...]

PR #1580, opened 2026-06-05
Fixes #1332, #1268: $select on Primitive Collections and Array-based Concurrency tokens. 

### LINKED ISSUE #1332: IfNoneMatch.ApplyTo does not work for byte[] fields.

anasik:
I opened the issues #1268, and #1332 in late 2024 and early 2025 respectively and submitted a PR each into the `release-8.x` branch. The PRs were #1282 and #1334 respectively.

I noticed for a long time that the bugs were never fixed in 9.x and 10.x. 

This PR knocks out both of them for 9.x.

**Please also let me know if I need to create a separate PR for 10.x because it appears to me that 10.x and 9.x have been very much in sync.**

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Common/TypeHelper.cs
+++ b/src/Microsoft.AspNetCore.OData/Common/TypeHelper.cs
@@ -386,6 +386,26 @@ internal static class TypeHelper
     }
 
     /// <summary>
+    /// Check whether the given type is a primitive type or known type.
+    /// </summary>
+    /// <param name="type">The type to validate.</param>
+    /// <returns>True if type is primitive or known type, otherwise False.</returns>
+    public static bool IsPrimitiveOrKnownType(Type type)
+    {
+        return type.IsPrimitive 
+               || type == typeof(string)
+               || type == typeof(Uri)
+               || type == typeof(DateTime)
+#if NET6_0_OR_GREATER
+    || type == typeof(DateOnly)
+    || type == typeof(TimeOnly)
+#endif
+               || type == typeof(DateTimeOffset)
+               || type == typeof(Guid)
+               || type == typeof(Decimal);
+    }
+
+    /// <summary>
     /// Determines whether the specified <see cref="Type"/> represents an <see cref="IAsyncEnumerable{T}"/> type.
     /// </summary>
     /// <param name="clrType">The <see cref="Type"/> to evaluate.</param>
--- a/test/Microsoft.AspNetCore.OData.E2E.Tests/Microsoft.AspNetCore.OData.E2E.Tests.csproj
+++ b/test/Microsoft.AspNetCore.OData.E2E.Tests/Microsoft.AspNetCore.OData.E2E.Tests.csproj
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>net8.0</TargetFrameworks>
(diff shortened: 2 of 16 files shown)


anasik:
@WanjohiSammy , @xuzhg , @habbes

anasik:
@gathogojr

anasik:
@WanjohiSammy please let me know if this looks okay or if I need to change anything

anasik:
@julealgon any idea why I'm failing to get attention here?

test/Microsoft.AspNetCore.OData.E2E.Tests/Microsoft.AspNetCore.OData.E2E.Tests.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>net8.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1;net6.0;net8.0</TargetFrameworks>
```
WanjohiSammy: we should only target net8.0 here

src/Microsoft.AspNetCore.OData/Common/TypeHelper.cs
```diff
@@ -385,6 +385,26 @@ Type collectionInterface
...
+    /// </summary>
+    /// <param name="type">The type to validate.</param>
+    /// <returns>True if type is primitive or known type, otherwise False.</returns>
+    public static bool IsPrimitiveOrKnownType(Type type)
+    {
+        return type.IsPrimitive 
+               || type == typeof(string)
+               || type == typeof(Uri)
+               || type == typeof(DateTime)
+#if NET6_0_OR_GREATER
```
WanjohiSammy: No need of this preprocessor directive here. This is because main branch targets net8.0

anasik:
@WanjohiSammy Please also let me know if I need to create a separate PR for 10.x

WanjohiSammy (approved):

WanjohiSammy:
LGTM - could we resolve the conflicts

anasik:
> LGTM - could we resolve the conflicts

@WanjohiSammy apparently there's no conflicts. 

Please also let me know if I need to create a separate PR for 10.x

WanjohiSammy:
> Please also let me know if I need to create a separate PR for 10.x

Yes please, you could cherrypick this commit and create a PR into dev-10.x

anasik:
> > Please also let me know if I need to create a separate PR for 10.x
> 
> Yes please, you could cherrypick this commit and create a PR into dev-10.x

@WanjohiSammy, done! Opened #1603.

PR #1587, opened 2026-07-09
Fix matching multiple OData routing metadata

### LINKED ISSUE #753: ODataRoutingMatcherPolicy Does Not Consider All Routing Metadata
**Assemblies affected**
ASP.NET Core OData 8.x

**Describe the bug**
Various paths through `ODataRoutingApplicationModelProvider` can add multiple `IODataRoutingMetadata` to an `Endpoint`, but  `ODataRoutingMatcherPolicy` will only ever consider the first one:

https://github.com/OData/AspNetCoreOData/blob/61ae3232d4ad687bc552fc583b32103f1aeff41a/src/Microsoft.AspNetCore.OData/Routing/ODataRoutingMatcherPolicy.cs#L86

This can result in some endpoints to not match when they should. There doesn't appear to be any documentation that indicates this is the expected behavior.

**Reproduce steps**
Consider a versioned API, which defines:

- 1.0
- 2.0
- 3.0
- 0.9 (deprecated)

Each version has its own EDM and ultimately adds 4 `ODataRoutingMetadata` instances to the `Endpoint`. The appropriate EDM is matched by the applied `ApiVersionAnnotation` to the incoming request. Consider the following controller. 

```c#
[ApiVersionNeutral]
public class FunctionsController : ODataController
{
    [HttpGet( "api/GetSalesTaxRate(PostalCode={postalCode})" )]
    [ProducesResponseType( typeof( double ), 200 )]
    public IActionResult GetSalesTaxRate( int postalCode ) => Ok( 5.6 );
}
```

A _version-neutral_ controller can match any API version, including none at all. OData, however, must have an EDM.  In this scenario, a developer is expected to use the same function definition for each version, but that's their discretion. This configuration yields the following results:

| Request | Result |
| --------- | ------ |
| `api/GetSalesTaxRate(PostalCode=98052)` | 200 |
| `api/GetSalesTaxRate(PostalCode=98052)?api-version=1.0` | 200 |
| `api/GetSalesTaxRate(PostalCode=98052)?api-version=2.0` | 404 |
| `api/GetSalesTaxRate(PostalCode=98052)?api-version=3.0` | 404 |

This happens because once an explicit version is specified, it cannot match up to the correct EDM as only the first set of metadata is considered. This is only one example, but there are any number of other cases where this _could_ happen.

**Expected behavior**
`ODataRoutingMatcherPolicy.ApplyAsync` should consider all `IODataRoutingMetadata` before invalidating a candidate.

```diff
- IODataRoutingMetadata metadata = candidate.Endpoint.Metadata.OfType<IODataRoutingMetadata>().FirstOrDefault();
+ IODataRoutingMetadata[] metadata = candidate.Endpoint.Metadata.OfType<IODataRoutingMetadata>().ToArray();
- if (metadata == null)
+ if (metadata.Length == 0)
{
    continue;
}

if (odataFeature.Path != null)
{
    // If it's odata endpoint, and we have a path set, let other odata endpoints invalid.
    candidates.SetValidity(i, false);
    continue;
}

- ODataTemplateTranslateContext translatorContext =
-     new ODataTemplateTranslateContext(httpContext, candidate.Endpoint, candidate.Values, metadata.Model);

- ODataPath odataPath = _translator.Translate(metadata.Template, translatorContext);

+ ODataPath odataPath = null;

+ for (var j = 0; odataPath == null && j < metadata.Length; i++)
+ {
+     ODataTemplateTranslateContext translatorContext =
+         new ODataTemplateTranslateContext(httpContext, candidate.Endpoint, candidate.Values, metadata[j].Model);
+     odataPath = _translator.Translate(metadata[j].Template, translatorContext);
+ }

if (odataPath != null)
{
    odataFeature.RoutePrefix = metadata.Prefix;
    odataFeature.Model = metadata.Model;
    odataFeature.Path = odataPath;

    MergeRouteValues(translatorContext.UpdatedValues, candidate.Values);

    // Shall we break the remaining candidates?
    // So far the answer is no. Because we can use this matcher to obsolete the unmatched endpoint.
    // break;
}
else
{
    candidates.SetValidity(i, false);
}
```

lpperras:
Fixes #753.

`ODataRoutingMatcherPolicy` previously considered only the first
`IODataRoutingMetadata` entry attached to an endpoint.

The matcher now evaluates metadata entries in order and uses the first
successful translation, including its EDM model, route prefix, and updated
route values.

Added focused unit coverage for:
- falling through from a failed metadata entry to a successful one
- stopping after the first successful match
- invalidating a candidate only when all metadata entries fail

The implementation approach is based on the solution proposed by
@commonsensesoftware in the original issue.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Routing/ODataRoutingMatcherPolicy.cs
+++ b/src/Microsoft.AspNetCore.OData/Routing/ODataRoutingMatcherPolicy.cs
@@ -83,8 +83,8 @@ internal class ODataRoutingMatcherPolicy : MatcherPolicy, IEndpointSelectorPolic
                 continue;
             }
 
-            IODataRoutingMetadata metadata = candidate.Endpoint.Metadata.OfType<IODataRoutingMetadata>().FirstOrDefault();
-            if (metadata == null)
+            IODataRoutingMetadata[] metadataItems = candidate.Endpoint.Metadata.OfType<IODataRoutingMetadata>().ToArray();
+            if (metadataItems.Length == 0)
             {
                 continue;
             }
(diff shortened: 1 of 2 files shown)


lpperras:

lpperras:
@xuzhg @WanjohiSammy we might need to shift our versioning strategy for our product depending on if this will ever be approved. We don't want to run of a fork but need this fix to move forward. Could you please let me know if there is not chance this makes it to the product so that we can align accordingly on our side?

PR #1592, opened 2026-07-17
fix: Add opt-in diagnostic logging for OData query validation failures

WanjohiSammy:
## Summary

When an OData request has a `$select`/`$expand` that queries a property that does not exist on the target type, the query parser throws while validating: the **controller** path (`EnableQueryAttribute`) returns `HTTP 400` and the **minimal API** path surfaces the `ODataException` from `ODataQueryEndpointFilter`. The parse throws **before** the parsed query options are populated, so telemetry that reads them sees an empty set and the failure gives no hint of the offending property, type, options, or endpoint.

This PR adds an **opt-in** structured `ILogger` diagnostic on the failure path (shared by both paths) that captures those details from request state that is still available. It is **off by default**; the HTTP response, its body, and the thrown exception are never changed.

## Configuration (default is a no-op)

Off by default (`EnableQueryValidationErrorLogging = false`).

| Setting | Controller | Minimal API |
|---|---|---|
| Enable per action/endpoint | `[EnableQuery(EnableQueryValidationErrorLogging = true)]` | `.WithODataOptions(o => o.SetQueryValidationErrorLogging(true))` |
| Enable globally | `ODataOptions.EnableQueryValidationErrorLogging = true` | `services.AddOData(o => o.SetQueryValidationErrorLogging(true))` |
| Level (global, default `Warning`) | `ODataOptions.QueryValidationErrorLogLevel = ...` | `services.AddOData(o => o.SetQueryValidationErrorLogLevel(...))` |

## After - opt-in

Example (console renderer, default level):

```
warn: Microsoft.AspNetCore.OData.Query.EnableQueryAttribute[0]
      OData query validation failed. Endpoint: odata/Customers, Type: NS.Customer, Query options: $expand=NoSuchOrders. Could not find a property named 'NoSuchOrders' on type 'NS.Customer'.
      Microsoft.OData.ODataException: Could not find a property named 'NoSuchOrders' on type 'NS.Customer'.
```

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/EnableQueryAttribute.Config.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/EnableQueryAttribute.Config.cs
@@ -7,6 +7,7 @@
 
 using System;
 using Microsoft.AspNetCore.OData.Query.Validator;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.OData.Query;
 
@@ -20,6 +21,7 @@ public partial class EnableQueryAttribute
     // validation settings
     private ODataValidationSettings _validationSettings;
     private string _allowedOrderByProperties;
+    private bool? _enableQueryValidationErrorLogging;
 
     // query settings
     private ODataQuerySettings _querySettings;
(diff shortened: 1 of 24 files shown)


xuzhg (dismissed):

src/Microsoft.AspNetCore.OData/Query/QueryValidationErrorLogger.cs
```diff
@@ -0,0 +1,111 @@
...
+
+    /// <summary>
+    /// Writes the diagnostic entry for a failed query validation at the specified level. Does nothing when no
+    /// logger is available or the level is not enabled.
+    /// </summary>
+    /// <param name="logger">The logger to write to, or <c>null</c> when none is available.</param>
+    /// <param name="logLevel">The level at which the diagnostic is written.</param>
+    /// <param name="httpContext">The <see cref="HttpContext"/> for the current request.</param>
+    /// <param name="exception">The exception raised while validating the query.</param>
+    internal static void LogWarningOrError(ILogger logger, LogLevel logLevel, HttpContext httpContext, Exception exception)
```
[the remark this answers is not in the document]
WanjohiSammy: Applied

src/Microsoft.AspNetCore.OData/Query/EnableQueryAttribute.Config.cs
```diff
@@ -20,6 +21,7 @@ public partial class EnableQueryAttribute
     // validation settings
     private ODataValidationSettings _validationSettings;
     private string _allowedOrderByProperties;
+    private bool? _enableQueryValidationErrorLogging;
```
xuzhg, line 24: can we use the ILogger directly?
I mean can we not use this configuration, instead, can we rely on customer's setting about the ILogger? 
Fetch the certain 'ILogger' , if set, use it, otherwise, not log the error?

xuzhg: @copilot

xuzhg (approved):

PR #1603, opened 2026-07-31
Fixes #1332, #1268 for 10.x: $select on Primitive Collections and Array-based Concurrency tokens.

### LINKED ISSUE #1332: IfNoneMatch.ApplyTo does not work for byte[] fields.

anasik:
I opened the issues https://github.com/OData/AspNetCoreOData/issues/1268, and https://github.com/OData/AspNetCoreOData/issues/1332 in late 2024 and early 2025 respectively and submitted a PR each into the release-8.x branch. The PRs were https://github.com/OData/AspNetCoreOData/pull/1282 and https://github.com/OData/AspNetCoreOData/pull/1334 respectively.

I noticed for a long time that the bugs were never fixed in 9.x and 10.x.

#1580 just got closed that fixes them in 9.x

This PR knocks out both of them for 10.x.

(cherry picked from commit a05e1ad06bc7da68629b03ec86c3115aa3fc170a)


anasik:
@WanjohiSammy

WanjohiSammy (approved):

PR #1604, opened 2026-08-01
Fixes #1581: Nullable enum with 'in' operator throws when collection contains starts with a null value

### LINKED ISSUE #1581: Nullable enum with 'in' operator throws ArgumentException when collection contains both null and enum values
When using the in operator with a nullable enum property and a collection containing both null and enum values (e.g., $filter=SkinColor in (null, 'Red')), the LINQ filter binder throws:
 
  System.ArgumentException: The value "Red" is not of type 
  "System.Nullable`1[Color]" and cannot be used in this generic collection.
 
 The OData URI parser correctly parses the filter (fixed in OData/odata.net#3491). However, ASP.NET OData's filter binder creates a typed collection that doesn't properly handle the mix of null and enum values for nullable enum properties.
 
 Steps to Reproduce
 
  1. Entity with nullable enum: Color? SkinColor
  2. Query: GET /Products?$filter=SkinColor in (null, 'Red')
 
 Expected: Returns entities where SkinColor is null or Red
 Actual: Throws ArgumentException
 
 Notes:
 
  - SkinColor in (null) alone works ✓
  - SkinColor in ('Red') alone works ✓
  - Only the combination fails

anasik:
## Summary                                                                                                                                                                                                     
                                                                                                                                                                                                                 
  Fixes #1581. `$filter=SomeEnum in (null, 'Value')` on a **nullable** enum property threw:                                                                                                                      
                                                                                                                                                                                                                 
  System.ArgumentException: The value "Value" is not of type "System.Nullable`1[...]"                                                                                                                            
  and cannot be used in this generic collection.                                                                                                                                                                 
                                                                                                                                                                                                                 
  ...whenever a `null` appeared *before* a real enum value in the `in (...)` list. The                                                                                                                           
  reverse order (`in ('Value', null)`) already worked, which is why this slipped                                                                                                                                 
  through existing coverage.                                                                                                                                                                                     
                                                                                                                                                                                                                 
  ## Root cause                                                                                                                                                                                                  
                                                                                                                                                                                                                 
  `QueryBinder.BindCollectionConstantNode` determines the CLR item type for the whole                                                                                                                            
  `in (...)` list from the **first** element only. `RetrieveClrTypeForConstant` only                                                                                                                             
  unwrapped `Nullable<TEnum>` → `TEnum` when the value it was inspecting was non-null.                                                                                                                           
  So when the first item was `null`, the CLR type stayed `Nullable<TEnum>` for the rest                                                                                                                          
  of the loop, `constantType.IsEnum` came back `false` (always false for `Nullable<T>`),                                                                                                                         
  and the next real item's `ODataEnumValue` got added directly into a `List<TEnum?>`                                                                                                                             
  instead of being converted first.                                                                                                                                                                              
                                                                                                                                                                                                                 
  ## Fix                                                                                                                                                                                                         
                                                                                                                                                                                                                 
  `QueryBinder.cs` only — hoisted the `Nullable<TEnum>` unwrap into its own preceding                                                                                                                            
  check that runs regardless of whether the value being inspected is null; the                                                                                                                                   
  value-conversion step still only runs when non-null. No other logic touched.                                                                                                                                   
                                                                                                                                                                                                                 
  ## Tests                                                                                                                                                                                                       
                                                                                                                                                                                                                 
  - Unit: `EnumInExpression_NullableEnum_WithNullValueFirst` in `FilterBinderTests.cs`                                                                                                                           
    — the null-first ordering not covered by the existing                                                                                                                                                        
    `EnumInExpression_NullableEnum_WithNullValue` test.                                                                                                                                                          
  - E2E: nullable `MaritalStatus` property added to the `Enums` test model/controller,                                                                                                                           
    with a test filtering `MaritalStatus in (null,'Married')` through the real pipeline.                                                                                                                         
                                                                                                                                                                                                                 
  Both reproduce the exact reported exception when the fix is reverted, and pass with it.

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
@@ -1492,44 +1492,44 @@ public abstract partial class QueryBinder
         if (edmTypeReference != null && edmTypeReference.IsEnum())
         {
             constantType = Nullable.GetUnderlyingType(constantType) ?? constantType;
+        }
 
-            if (value != null)
+        if (value != null && edmTypeReference != null && edmTypeReference.IsEnum())
+        {
+            ODataEnumValue odataEnumValue = (ODataEnumValue)value;
+            string strValue = odataEnumValue.Value;
+            Contract.Assert(strValue != null);
+
+            IEdmEnumType enumType = edmTypeReference.AsEnum().EnumDefinition();
+            ClrEnumMemberAnnotation memberMapAnnotation = context.Model.GetClrEnumMemberAnnotation(enumType);
+            if (memberMapAnnotation != null)
             {
-                ODataEnumValue odataEnumValue = (ODataEnumValue)value;
-                string strValue = odataEnumValue.Value;
-                Contract.Assert(strValue != null);
-
-                IEdmEnumType enumType = edmTypeReference.AsEnum().EnumDefinition();
-                ClrEnumMemberAnnotation memberMapAnnotation = context.Model.GetClrEnumMemberAnnotation(enumType);
-                if (memberMapAnnotation != null)
+                IEdmEnumMember enumMember = enumType.Members.FirstOrDefault(m => m.Name == strValue);
+                if (enumMember == null)
                 {
-                    IEdmEnumMember enumMember = enumType.Members.FirstOrDefault(m => m.Name == strValue);
-                    if (enumMember == null)
-                    {
-                        enumMember = enumType.Members.FirstOrDefault(m => m.Value.ToString() == strValue);
-                    }
+                    enumMember = enumType.Members.FirstOrDefault(m => m.Value.ToString() == strValue);
+                }
 
-                    if (enumMember != null)
+                if (enumMember != null)
+                {
+                    Enum clrMember = memberMapAnnotation.GetClrEnumMember(enumMember);
+                    if (clrMember != null)
                     {
-                        Enum clrMember = memberMapAnnotation.GetClrEnumMember(enumMember);
-                        if (clrMember != null)
-                        {
-                            value = clrMember;
-                        }
-                        else
-                        {
-                            throw new ODataException(Error.Format(SRResources.CannotGetEnumClrMember, enumMember.Name));
-                        }
+                        value = clrMember;
                     }
                     else
                     {
-                        value = Enum.Parse(constantType, strValue);
+                        throw new ODataException(Error.Format(SRResources.CannotGetEnumClrMember, enumMember.Name));
                     }
                 }
                 else
                 {
                     value = Enum.Parse(constantType, strValue);
                 }
+            }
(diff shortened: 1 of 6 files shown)


anasik:
@WanjohiSammy I'll cherry pick this onto main once this is merged

anasik:
@xuzhg

xuzhg:
@anasik The latest ODL 9.0.0 was RTM on nuget.org. In that version, I changed the query option parser result what is related to the collection constant. I'd like to update the ASP.NET Core OData 10.x to depend on the ODL 9.0.0 first, then back to this PR. 

FYI, @WanjohiSammy

anasik:
> @anasik The latest ODL 9.0.0 was RTM on nuget.org. In that version, I changed the query option parser result what is related to the collection constant. I'd like to update the ASP.NET Core OData 10.x to depend on the ODL 9.0.0 first, then back to this PR. 
> 
> FYI, @WanjohiSammy 

 @xuhg Makes sense. Will you also do the same on the `main` branch?

xuzhg:
> > @anasik The latest ODL 9.0.0 was RTM on nuget.org. In that version, I changed the query option parser result what is related to the collection constant. I'd like to update the ASP.NET Core OData 10.x to depend on the ODL 9.0.0 first, then back to this PR.
> > FYI, @WanjohiSammy
> 
> @xuhg Makes sense. Will you also do the same on the `main` branch?

The changed query option parser is only on ODL 9.0.0 since it's a breaking change. So, I don't think we will do the same on the 'main'.

anasik:
@xuhg do you want me to help you update the dependency?

xuzhg:
> @xuhg do you want me to help you update the dependency?

I updated the dependency, can you rebase your PR?

anasik:
@xuzhg done

anasik:
@xuzhg are we waiting on something?

xuzhg:
> @xuzhg are we waiting on something?

No, Just lack of resource to review. But, we are good. I will merge it if no other concern. 
BTW, may I know whether you are waiting for a public release? or can you verify using the nightly?

anasik:
> > @xuzhg are we waiting on something?
> 
> No, Just lack of resource to review. But, we are good. I will merge it if no other concern. 
> BTW, may I know whether you are waiting for a public release? or can you verify using the nightly?

No no. I just saw the bug and fixed it. It's not affecting us at production or anything so far.

xuzhg (approved):
:shipit:

xuzhg:
@anasik did you fix it in the main branch? Excuse me, maybe I missed that. If no, do you have a plan to port it back? I do think the main branch has the same issue and be better to have such fix. What do you think?

anasik:
> @anasik did you fix it in the main branch? Excuse me, maybe I missed that. If no, do you have a plan to port it back? I do think the main branch has the same issue and be better to have such fix. What do you think?

Yes, I was waiting for this one to close so I can cherry-pick this on to main.

PR #1607, opened 2026-08-12
Fixes #1581: Nullable enum with 'in' operator throws when collection contains starts with a null value

### LINKED ISSUE #1581: Nullable enum with 'in' operator throws ArgumentException when collection contains both null and enum values

anasik:
Cherry picked from `dev-10.x`
See #1604

### DIFF
--- a/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
+++ b/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs
@@ -1489,13 +1489,16 @@ public abstract partial class QueryBinder
     {
         Type constantType = context.Model.GetClrType(edmTypeReference, context.AssembliesResolver);
 
+        if (edmTypeReference != null && edmTypeReference.IsEnum())
+        {
+            constantType = Nullable.GetUnderlyingType(constantType) ?? constantType;
+        }
+
         if (value != null && edmTypeReference != null && edmTypeReference.IsEnum())
         {
             ODataEnumValue odataEnumValue = (ODataEnumValue)value;
             string strValue = odataEnumValue.Value;
             Contract.Assert(strValue != null);
-
-            constantType = Nullable.GetUnderlyingType(constantType) ?? constantType;
 
             IEdmEnumType enumType = edmTypeReference.AsEnum().EnumDefinition();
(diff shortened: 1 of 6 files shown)


anasik:
@xuzhg @WanjohiSammy @habbes

WanjohiSammy (dismissed):

src/Microsoft.AspNetCore.OData/Query/Expressions/ExpressionBinderHelper.cs
```diff
@@ -129,8 +129,8 @@ public static Expression CreateBinaryExpression(BinaryOperatorKind binaryOperato
             right = CreateTimeBinaryExpression(right, querySettings);
         }
 
-        if ((IsType<DateOnly>(leftUnderlyingType) && IsDate(rightUnderlyingType)) ||
-            (IsDate(leftUnderlyingType) && IsType<DateOnly>(rightUnderlyingType)))
+        if ((IsType<DateOnly>(leftUnderlyingType) && (IsDate(rightUnderlyingType) || IsDateOrOffset(rightUnderlyingType))) ||
```
WanjohiSammy: This  is not part of this cherrypicked commit

anasik: Yeah my bad. I normally don't make more than one PR at a time. Lemme remove this

WanjohiSammy (approved):

